PR #110, opened 2015-01-13
Changed stageSize to an object

ephraimtabackman:
Changed stageSize to an object with two integer properties: "width" and "height".

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/flash_context/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow/flash_context/jsonschema/1-0-0
@@ -17,7 +17,15 @@
 			"type": "string"
 		},
 		"stageSize": {
-			"type": "string"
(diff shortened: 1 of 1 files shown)


schemas/com.snowplowanalytics.snowplow/flash_context/jsonschema/1-0-0
```diff
@@ -17,7 +17,15 @@
...
-			"type": "string"
+			"type": "object",
+			"properties": {
+				"width": {
+					"type": "integer"
+				},
+				"height": {
+					"type": "integer"
+				}
+			}
```
alexanderdean: At this level of nesting we should add in: `"required": ["width", "height"]`

alexanderdean:
See also comment about whether any other fields are required...

alexanderdean:
Looks good, merging...

PR #135, opened 2015-02-23
Add mobile_context 1-0-1 with networkType and networkTechnology

duncan:
A proposed version bump to include both a field for Android flavors (distinct builds for different stores, locales, etc) as well as network types (3G, LTE, WiFi).

I put this into a distinct version file following the established pattern. I hope that’s OK. 

The diff is:

```
18a19,21
> 		"buildFlavor": {
> 		    "type": "string"
> 		},
27a31,33
> 		"networkType": {
> 		    "type": "string"
> 		},
```

Does this make sense? Is it something you’d be interested in?

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-1
+++ b/schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-1
@@ -28,6 +28,9 @@
 		"networkType": {
 				"enum": [ "mobile", "wifi"]
 		},
+		"networkTechnology": {
+				"type": "string"
+    },
 		"openIdfa": {
 			"type": "string"
 		},
--- a/schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-1
+++ b/schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-1
@@ -1,0 +1,49 @@
+{
+	"$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+	"description": "Schema for mobile contexts",
+	"self": {
+		"vendor": "com.snowplowanalytics.snowplow",
+		"name": "mobile_context",
+		"format": "jsonschema",
+		"version": "1-0-0"
+	},
+
+	"type": "object",
+	"properties": {
+		"osType": {
+			"type": "string"
+		},
+		"osVersion": {
+			"type": "string"
+		},
+		"osFlavor": {
+		    "type": "string"
+		},
+		"deviceManufacturer": {
+			"type": "string"
+		},
+		"deviceModel": {
+			"type": "string"
+		},
+		"carrier": {
+			"type": "string"
+		},
+		"networkType": {
+		    "type": "string"
+		},
+		"openIdfa": {
+			"type": "string"
+		},
+		"appleIdfa": {
+			"type": "string"
+		},
+		"appleIdfv": {
+			"type": "string"
+		},
+		"androidIdfa": {
+			"type": "string"
+		}
+	},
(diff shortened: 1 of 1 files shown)


alexanderdean:
Hey @duncan! This looks cool. Feedback:

* I think it would be `osBuildFlavor` or `osFlavor`, not `buildFlavor`. Or maybe `androidFlavor` if this is a particularly Android-esque term? What do you think?
* The version line `"version": "1-0-0"` needs updating too

duncan:
`androidFlavor` it is. And the version is fixed up. But, in responding to this, I'm pondering if the `androidFlavor` should go here. Flavors are associated with the software, not the device or OS. See: http://tools.android.com/tech-docs/new-build-system/user-guide#TOC-Product-flavors 
 
Maybe it's better put with the user agent along with the software version number? Or maybe we need to be thinking of a property in a different context?

alexanderdean:
Hmm... Interesting. It looks on that page like there are two sibling data points: Build Type + Product Flavor. Are they easily retrieved by the Android Tracker, or would a user have to supply them?

It could be interesting to have a new context for describing the host app, with something like:

* Source control commit hash of build
* Version number
* Build type

... etc

duncan:
Yep. The flavor is automatically added in the BuildConfig. Here's a sample of the standard settings from the android tracker:

```
  public static final boolean DEBUG = false;
  public static final String APPLICATION_ID = "com.snowplowanalytics.snowplow.tracker";
  public static final String BUILD_TYPE = "release";
  public static final String FLAVOR = "";
  public static final int VERSION_CODE = 1;
  public static final String VERSION_NAME = "1.0";
```

Also of interest is the BUILD_TYPE. We have a similar mechanism in our current APIs that tells us this in a request header, but having a context with all this info—and similar info on OS X, iOS, etc, would be pretty good.

The only question on the Android side is how easy it is to get to the Build Config automatically as it's put into a package. But the package name is based on package in the `AndroidManifest.xml`, so I'm pretty sure we can.

If that's the case, I'd drop back to simply suggesting `networkType` on this context in favor of adding an "Application" context of some sort.

alexanderdean:
Hi @duncan - okay agreed, so let's drop this back to networkType and https://github.com/snowplow/iglu-central/issues/136. On networkType - should this be a string, or an enum - how knowable are the values?

duncan:
Thinking out loud (and reading through the information provided by the various APIs), there are two different things network information gives us. The primary is the distinction between cellular and wifi. With cellular, the customer is typically paying for data usage. With wifi not. 

The second thing network information gives us is speed. Well, sorta. It used to be that wifi could be considered to be faster than cellular, but that assumption is pretty much shot these days. My LTE connection bandwidth regularly exceeds the 50Mbps that my home networking connection gives. But, distinguishing between edge, 3g, and LTE does give us an indication of expected performance. 

Network type is easy. It's an enum of either `mobile` or `wifi`. 

The mobile radio technology is a bigger set and is likely to keep growing. If we capture this, it should be a string. So, maybe it's two properties:

```
"networkType": {
   "enum": [ "mobile", "wifi"]
},
"networkTechnology": {
   "type": "string" 
},
```

alexanderdean:
These look good to me!

duncan:
Cool! I think this one is a wrap... I'll look at implementing in the ObjC client first as that's where I'm going to spend the next part of my day.

alexanderdean:
The related Snowplow assets are here: https://github.com/snowplow/snowplow/tree/feature/blocked-schemas

See the milestone description for why these are in "blocked schemas".

To use them at Wunderlist: upgrade your Redshift table as per, then deploy the new JSON Paths file into your S3 Bucket representing your `:jsonpath_assets:` path. It will override the globally hosted one.

duncan:
Cool. Thanks for the info on how the blocked schemas worked.

PR #139, opened 2015-02-26
Adding "none" for network type

duncan:
While implementing this in ObjC, I realized that there are times when events will get generated when there's no network connection. Ergo, we need a none. 

I'm not quite sure how to attach a new PR to a closed one that's part of a Milestone—so this one seems have bundled up the full set of changes.

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-1
+++ b/schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-1
@@ -26,7 +26,7 @@
 			"type": "string"
 		},
 		"networkType": {
-				"enum": [ "mobile", "wifi"]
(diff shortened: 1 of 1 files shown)


alexanderdean:
Ah - good point. I'll add in null as well in case there are any mobile devices where we're not allowed to check the connection.

alexanderdean:
Actually I'll leave null out as we are using absence-from-required-fields to indicate null-support in this schema...

duncan:
Ah. That'll work.

alexanderdean:
Hey @duncan - I think maybe we got crossed wires on this one. Did you want to keep "none" as a flag value meaning that there is currently no network connection? (Versus null meaning "we don't know".) It's fine if you do.

duncan:
Maybe an offline would be better?

alexanderdean:
Agree - "offline" sounds good...

Issue #136, opened 2015-02-25
Add com.snowplowanalytics.snowplow/application_context/jsonschema/1-0-0


alexanderdean:
As per discussion in https://github.com/snowplow/iglu-central/pull/135

duncan:
Based on the things we collect, a first draft would be something like:

```
{
  "application": "Wunderlist",
  "version": "2.3.4",
  "build": "s9f2k2d",
  "androidFlavor" : "china",
  "buildType": "release"
}
```

Notes: I'm suggesting `build` instead of our own `githash` as git is too specific. We use git hashes, but others would use a linear build id like `432` or even like Mac OS X’s `14C109`.

For a 1.0.0 version, the above would capture what we need and I'd love to get something like it into our next release which will be the first with our Snowplow collectors.

alexanderdean:
It's taking shape! What about:

```json
{
  "name": "Wunderlist",
  "version": "2.3.4",
  "buildId": "s9f2k2d",
  "buildFlavor" : "china",
  "buildType": "release"
}
```

`buildFlavor` would make the context a little less mobile-specific...

alexanderdean:
Or event `buildStage: release` as type is a bit vague?

duncan:
I'm good with `buildFlavor`. Especially since on iOS, we could use this for "App Store", "Ad Hoc" etc based on provisioning profile. 

As to the `buildType/Stage`, I wonder if a boolean is more appropriate. Something like a `"developmentBuild": "true"` or a `"releaseBuild": "false"`... Hrmmmm...

alexanderdean:
Hmm, I quite like "buildIsRelease": bool. Release is perhaps less ambiguous than buildIsDebug, buildIsDevelopment...

alexanderdean:
Random thought - there could be something interesting (more for server-side apps) in capturing the log level (debug / warn / info kind of thing) - that's a pretty common unit of currency...

duncan:
`buildIsRelease` sounds good. Or, `buildIsProduction`? They're pretty equivalent—just thinking through how Rails has the concept of environment and how that'd map to this.

I can see that a `logLevel` property could be useful. It could probably be picked up pretty easily from Log4J or whatever configuration. But, I passed the idea by a few of our server peeps and they were lukewarm on the idea. So, for now, I think I'm pretty neutral on that one.

alexanderdean:
[...]

PR #140, opened 2015-02-27
Initial draft of application context

duncan:
As discussed in #136, an initial draft of an application context with `name`, `version`, `build`, `androidFlavor`, and `buildIsRelease` properties.

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/application_context/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow/application_context/jsonschema/1-0-0
@@ -1,0 +1,31 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for context related to a host application",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow",
+    "name": "application_context",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+
+  "type": "object",
+  "properties": {
+    "name": {
+      "type": "string"
+    },
+    "version": {
+      "type": "string"
+    },
+    "build": {
+      "type": "string"
+    },
+    "androidFlavor": {
+      "type": "string"
+    },
+    "buildIsRelease": {
+      "type": "boolean"
+    }
+  },
+  "required": ["name", "version"],
+  "additionalProperties": false
+}
--- a/schemas/com.snowplowanalytics.snowplow/application_context/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow/application_context/jsonschema/1-0-0
@@ -19,7 +19,7 @@
     "build": {
       "type": "string"
     },
-    "androidFlavor": {
+    "buildFlavor": {
       "type": "string"
     },
     "buildIsRelease": {
--- a/schemas/com.snowplowanalytics.snowplow/application_context/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow/application_context/jsonschema/1-0-0
@@ -19,7 +19,7 @@
     "build": {
       "type": "string"
     },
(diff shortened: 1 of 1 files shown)


schemas/com.snowplowanalytics.snowplow/application_context/jsonschema/1-0-0
```diff
@@ -0,0 +1,31 @@
...
+    "name": {
+      "type": "string"
+    },
+    "version": {
+      "type": "string"
+    },
+    "build": {
+      "type": "string"
+    },
+    "androidFlavor": {
```
alexanderdean: This was going to be buildFlavor no?

duncan: 😳 yah. it was. Fixing.

duncan:
I guess the other thing one could call it that might be slightly more generic would be `buildVariant`... ?

alexanderdean:
Oh yah, I like `buildVariant`.

alexanderdean:
Coming back to this cold 6 months later, "build" and "buildVariant" feel a bit unintuitive.

alexanderdean:
Hey @jbeemster - the original genesis for this context was Android apps. Would it be possible to automatically generate this context for an Android app, or would it need to be manually set by the app author? What about for Objective-C?

jbeemster:
It would need to be setup as a Tracker option but all that information should be available for both Android and Objective-C.  I am thinking it might be worthwhile exposing a Context object of sorts which handles all of the various contexts that we have popping up..

alexanderdean:
Sounds good - can you create tickets for both mobile trackers then @jbeemster ?

alexanderdean:
From a Java/Scala/Clojure perspective, the obvious field missing is organization aka vendor.

alexanderdean:
Consider grouping the build fields into a nested sub-object

TiernanKennedy:
Spec looks good to me @alexanderdean.

With iOS we always have a build number, so requiring that would be fine. 

I'm not sure why buildVariant is needed, given that we have the mobile_context?

alexanderdean:
Thanks @TiernanKennedy. I think `buildVariant` is an Android thing?

TiernanKennedy:
Hmm, should there be OS specific stuff in here? Part of me thinks there should just be the addition of `build` and `version` in the mobile context. Is there a reason to keep them separate? @alexanderdean

alexanderdean:
The mobile context is intended to describe the mobile device which is hosting the app; the application context is intended to describe the app which is running the tracker - so application build and version details belong in the application context. We could potentially rename `buildVariant` to something less Android-flavoured...

alexanderdean:
Related: http://stackoverflow.com/questions/30223676/adding-a-build-variant-and-running-it-on-ios-8

TiernanKennedy:
Ah I see what you mean now. Yeah then `buildVariant` does make sense.

:+1:

alexanderdean:
We're missing the application license as well...

alexanderdean:
Should we add sourceRepository, sourceTag and sourceCommit too?

bogaert:
> We're missing the application license as well...
> Should we add sourceRepository, sourceTag and sourceCommit too?

I don't think we need either for a first release of the schema.

snowplowcla:
@duncan has signed the Software Grant and Corporate Contributor License Agreement

alexanderdean:
I'm not sure what I meant by application license.

> >  Should we add sourceRepository, sourceTag and sourceCommit too?
>
> I don't think we need either for a first release of the schema.

Because of *Redshift blocked schemas*, the first release will be the only release for the foreseeable future.

chuwy:
Pushing back to R97. @keanerobinson please let us know what do you think of it when you're back.

chuwy:
Closing in favor of https://github.com/snowplow/iglu-central/pull/847

PR #150, opened 2015-03-13
Added stream_write_failed schema (closes #149)

### LINKED ISSUE #149: Schema for a com.snowplowanalytics.monitoring/stream_write_failed event
Fred! While you are thinking about back-off and retry for Scala Kinesis Enrich, could you draft up an event to model a put failure?

Have a think of what fields would be useful to record in this event (this will be emitted by Snowplow)...

fblundun:


alexanderdean:
Merged at the command line

PR #185, opened 2015-06-24
Kinesis app schemas

fblundun:


schemas/com.snowplowanalytics.monitoring/sink_write_failed/jsonschema/1-0-0
```diff
@@ -0,0 +1,37 @@
...
+		"failureCount": {
+			"type": "integer"
+		},
+
+		"initialFailureTime": {
+			"type": "number"
+		},
+
+		"lastRetryPeriod": {
+			"type": "number"
```
alexanderdean: number or integer for these three numeric fields? The Redshift schema says int, which is int4, so would need to be constrained to these values: -2147483648 to +2147483647

[1] http://docs.aws.amazon.com/redshift/latest/dg/r_Numeric_types201.html

For example numeric constraints see: https://github.com/snowplow/iglu-central/blob/master/schemas/org.w3/PerformanceTiming/jsonschema/1-0-0#L13

fblundun: Good point!

I think failureCount can be an integer since we are unlikely to have 2147483647 failures.
But initialFailureTime is in ms since the Unix epoch, i.e. about 1 435 234 971 319, which is bigger than the Redshift int4 maximum.

So it sounds like I need to change the Redshift type to bigint like it is for performance timing.

The types can all be integer rather than number and I can add the -9223372036854775808 to 9223372036854775807 constraint to the schema.

alexanderdean: Thanks Fred - can you liaise with Josh to get the schemas/Redshift updated?

fblundun: I will do!

PR #197, opened 2015-07-08
Snowplow cli schemas

fblundun:


alexanderdean:
Looks good! Feedback:

* All timestamps should be `"format": "date-time"`, not a custom pattern

PR #200, opened 2015-07-28
Added com.amazon.aws.cloudfront/wd_access_log/jsonschema/1-0-4 (close…

fblundun:
…s #199)

### DIFF
--- a/schemas/com.amazon.aws.cloudfront/wd_access_log/jsonschema/1-0-0
+++ b/schemas/com.amazon.aws.cloudfront/wd_access_log/jsonschema/1-0-0
@@ -38,7 +38,8 @@
 			"maxLength": 8192
 		},
 		"scStatus": {
-			"type": ["string", "null"]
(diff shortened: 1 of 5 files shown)


alexanderdean:
Hey @fblundun ! This looks good - but I think that if we are adding maxLength constraints on an existing field like scStatus, we have to backport ("patch") the prior schemas to have the same constraints, or else we don't strictly have an additive change... Do you agree?

fblundun:
Good point! Should I add the maxLengths to the previous versions, or just make this version 2-0-0?

alexanderdean:
Me on HipChat in case anybody is interested:

```
I think a good test of these sorts of questions is to imagine we were starting from scratch today
so imagine we are sitting down to draft 1-0-0
knowing what we know about schema-guru
we would add those constraints from day 1
therefore, I think we should patch the existing schemas
```

alexanderdean:
Thanks Fred, looks good

PR #215, opened 2015-09-26
Add schemas for cookie extractor enrichment

kazjote:
Implements https://github.com/snowplow/snowplow/issues/998


schemas/com.snowplowanalytics.snowplow/cookie_extractor_context/jsonschema/1-0-0
```diff
@@ -0,0 +1,25 @@
...
+		"version": "1-0-0"
+	},
+
+	"type": "object",
+	"properties": {
+		"cookies": {
+			"type": "object",
+			"properties": {
+				"name": "string",
+				"value": "string"
```
fblundun: This should be

```json
"name": {
    "type": "string
},
"value": {
    "type": "string
}
```

schemas/com.snowplowanalytics.snowplow/cookie_extractor_context/jsonschema/1-0-0
```diff
@@ -0,0 +1,25 @@
...
+			},
+			"required": ["name", "value"]
+		}
+	},
+	"required": ["cookies"],
+	"additionalProperties": false
+}
```
fblundun: I think there's one level too many here. Rather than

```json
{
  "cookies": {
    "name": "x",
    "value": "y"
  }
}
```

Shouldn't it just be

```json
{
  "name": "x",
  "value": "y"
}
```

?

jbeemster: Looking at the test suite from the companion PR there is no "cookie" wrapper.

alexanderdean: Is there a reason why we say browser cookies have Snowplow as their vendor?

Why not org.ietf/http_cookie or org.ietf.rfc-6265/http_cookie ?

A

kazjote:
@jbeemster, @fblundun thanks for the review - cookie_extractor_context was messed up. I fixed it.

@alexanderdean I am not sure I fully understand what vendor means. I understood it as the entity that maintain and decide on the schema. In this understanding vendor would be Snowplow / Snowplow community. This understanding follows the description that is here: http://snowplowanalytics.com/blog/2014/05/15/introducing-self-describing-jsons/
```
vendor tells us who created this JSON Schema
```

alexanderdean:
Hey @kazjote - sorry for the confusion: the vendor is the organization that defined the original entity, not the entity which transcribed the entity to JSON Schema. So if Snowplow creates schema for MailChimp, we put them under com.mailchimp, not com.snowplowanalytics. Think of the vendor as the namespace for the 

In this case, the entity is an http cookie, which isn't an entity invented by Snowplow, but rather an entity standardized by IETF in RFC-6265.

Another way of thinking of it is: if we define this as org.ietf.rfc-6265/http_cookie, then this schema (and companion Redshift table) can be used by:

1. This enrichment
2. A different implementation of this enrichment one day
3. Someone adding cookies as contexts in the JavaScript Tracker

TL;DR - the vendor and name should describe the entity itself, not the current mechanism producing that entity.

kazjote:
Thanks for the explanation. I renamed context schema and update enricher pull request https://github.com/snowplow/snowplow/pull/2072

alexanderdean:
Looks great - thanks @kazjote !

alexanderdean:
@fblundun - from your review of the CookieExtractorEnrichment - should we be tolerating null for either of these fields?

alexanderdean:
https://hc.apache.org/httpcomponents-core-ga/httpcore/apidocs/org/apache/http/HeaderElement.html

> Any amount of white space is allowed between any part of the header, element or param and is ignored. A missing value in any element or param will be stored as the empty String; if the "=" is also missing null will be stored instead.

alexanderdean:
@fblundun, @kazjote - I've confirmed to need to allow null for cookie value (however not for cookie name):

https://github.com/snowplow/snowplow/blob/RetentionGrid-feature/cookie_extractor/3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/CookieExtractorEnrichmentSpec.scala#L59

alexanderdean:
These are in the release branch now, thanks so much, closing...

PR #221, opened 2015-10-07
Add com.amazon.aws.ec2/instance_identity_document/jsonschema/1-0-0

chuwy:


schemas/com.amazon.aws.ec2/instance_identity_document/jsonschema/1-0-0
```diff
@@ -0,0 +1,75 @@
...
+    "instanceType" : {
+      "type" : "string"
+    },
+    "version" : {
+      "type" : "string"
+    },
+    "pendingTime" : {
+      "type" : "string",
+      "format" : "date-time",
+      "minLength" : 20,
```
alexanderdean: Do we need minLength and maxLength on date-times?

chuwy: Not sure. ISO 8601 defines different patterns for date time, which can have different lenghts. So AWS uses strings with 20 chars.

alexanderdean: It feels like when we know it's a date-time, the length of the string containing the date-time isn't relevant. Does that make sense?

chuwy: Yes, that makes sense. In Redshift it will TIMESTAMP anyway. I'll fix it.

alexanderdean:
@chuwy - how confident are we about the fixed lengths of instanceId, ramdiskId, imageId, kernelId?

chuwy:
Hard question. Let's say 99%.
I searched this through the web and and all examples I saw has form like xxx-yyyyyyyy (ramdisk, image, kernel) and i-yyyyyyyy (instance) where yyyyyyyy is 8 byte hex string. 
I also saw couple examples on SO where others developers relied on this form.
But what I haven't found is specification that explicitly states that it couldn't be in other form. That's for 1%. I think the only to be sure is to ask on amazon forms. I'll do it.

alexanderdean:
Thanks @chuwy - 99% is good enough for me!

chuwy:
https://forums.aws.amazon.com/thread.jspa?threadID=217387
Also I for same 99% sure that billingProducts has form of array with bp-xxxxxxxx and ramdiskId is in most cases will be null. But I thought it not relevant.

PR #236, opened 2015-10-14
Ecommerce schemas

fblundun:


alexanderdean:
The biggest problem I can see here is that GA seems to stringly type any properties of type "Number" or "Currency".

This is something that doesn't work well in our Iglu / Schema Guru / Snowplow tooling - we expect a number to be formatted as a number.

Unlike GA, we can't maintain a master list of properties of type "Number" or "Currency" and attempt to cast them in our ETL; we need these fields to be correctly typed from the beginning...

schemas/com.google.analytics.ecommerce/item/jsonschema/1-0-0
```diff
@@ -0,0 +1,38 @@
...
+			"type": "string",
+			"maxLength": 500
+		},
+		"category": {
+			"type": "string",
+			"maxLength": 500
+		},
+		"currency": {
+			"type": "string",
+			"maxLength": 16
```
alexanderdean: Per https://support.google.com/analytics/answer/6205902#supported-currencies and "The local currency must be specified in the ISO 4217 standard" [1] I think we can safely make all currency fields minLength 3, maxLength 3...

[1] https://developers.google.com/analytics/devguides/collection/analyticsjs/ecommerce#multicurrency

fblundun: Sounds sensible!

fblundun:
According to [this](https://developers.google.com/analytics/devguides/collection/analyticsjs/enhanced-ecommerce#action-types), productFieldOption.position and impressionFieldObject.position are both numbers like `2` but promoFieldObject.position is a string like `"banner_slot_1"`. That's why the type of the "position" field isn't consistent across these schemas.

alexanderdean:
@fblundun - that makes sense in its own way. It sounds like for products/impressions, the position is basically the position in a catalogue/layout view, whereas in a promo the position refers to the ad unit showing the promo...

PR #251, opened 2015-11-11
Webhooks/sendgrid

ninjabear:


alexanderdean:
This looks good to me! Note that when we do the CHANGELOG and release, we should be referring to `com.sendgrid/processed/jsonschema/1-0-0` not `com.sendgrid/processed`

ninjabear:
@alexanderdean hopefully that's what you were saying? I also properly nested all the files under a jsonschema directory which I hadn't noticed is what the others do

alexanderdean:
That's the one - thanks @ninjabear !

alexanderdean:
"additionalProperties" should be true unless we are guaranteeing that no additional properties could be added by SendGrid (all per HipChat discussion)

ninjabear:
@alexanderdean do you think this should include sub trees? [Example](https://gist.github.com/ninjabear/c7b1ab3525d1bb60e9c7#file-deferred_schema_with_optionals-json-L35)

I've set them all to true for now, as I think the same problems with setting it to false probably apply to sub items

alexanderdean:
Sorry for delay in responding - *yes* I think we should allow additionalProperties in sub-trees.

alexanderdean:
Closing this now we have a draft branch (releases/r64).

PR #254, opened 2015-11-16
Webhooks/urbanairship

ninjabear:

PR #277, opened 2015-12-16
Webhooks/urbanairship

ninjabear:
Hey @alexanderdean we created and closed #254 for this but in testing I regenerated the jsonschemas (to take into account the sample provided by UA). I think we can strip out the old commits for this and replace them with these.


alexanderdean:
Hey @ninjabear - okay sure thing. I had a branch, releases/r46, I was about to press the button on. Would you mind creating a new branch, release/46-final, and cherry-picking over what we can keep and then splicing in these? Many thanks!

ninjabear:
Sure will do!

ninjabear:
See #278, will close this PR

PR #284, opened 2016-01-26
Added com.snowplowanalytics.snowplow.enrichments/api_lookup_enrichmen…

### LINKED ISSUE #276: Schema for a com.snowplowanalytics.snowplow.enrichments/api_lookup_enrichment_config

alexanderdean:
…t_config/jsonschema/1-0-0 (close #276)

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow.enrichments/api_request_enrichment_config/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow.enrichments/api_request_enrichment_config/jsonschema/1-0-0
@@ -1,0 +1,165 @@
+{
+	"$schema": "http://json-schema.org/schema#",
+	"description": "Schema for API Request Enrichment configuration",
+	"self": {
+		"vendor": "com.snowplowanalytics.snowplow.enrichments",
+		"name": "api_request_enrichment_config",
+		"format": "jsonschema",
+		"version": "1-0-0"
+	},
+
+	"type": "object",
+	"properties": {
+		"vendor": {
+			"type": "string"
+		},
+		"name": {
+			"type": "string"
+		},
+		"enabled": {
+			"type": "boolean"
+		},
+
+		"parameters": {
+			"type": "object",
+			"properties": {
+				"inputs": {
+					"type": "array",
+					"items": {
+						"type": "object",
+						"properties": {
+							"key": {
+								"type": "string",
+								"pattern": "^[A-Za-z0-9_-]+$"
+							},
+							"pojo": {
+								"type": "object",
+								"properties": {
+									"field": {
+										"type": "string"
+									}
+								},
+								"additionalProperties": false
+							},
+							"json": {
+								"type": "object",
+								"properties": {
+									"field": {
+										"type": "string",
+										"enum": ["unstruct_event", "contexts", "derived_contexts"]
+									},
+								  "schemaCriterion": {
+									  "type": "string",
+									  "pattern": "^iglu:[a-zA-Z0-9-_.]+/[a-zA-Z0-9-_]+/[a-zA-Z0-9-_]+/[0-9]+-([0-9]+|\\*)-([0-9]+|\\*)$"
+								  },
+								  "jsonPath": {
+									  "type": "string"
+								  }
+								},
+								"additionalProperties": false
+							}
+						},
+						"additionalProperties": false,
+						"minProperties": 2,
+						"maxProperties": 2,
+						"required": ["key"]
+					}
+				},
+				"api": {
+					"type": "object",
+					"minProperties": 1,
+					"maxProperties": 1,
+					"properties": {
+						"http": {
+							"type": "object",
+							"properties": {
+								"method": {
+									"type": "string",
+									"enum": ["GET", "POST", "PUT"]
+								},
+								"uri": {
+									"type": "string"
+								},
+								"timeout": {
+									"type": "integer",
+									"minimum": 1,
+									"maximum": 60000
+								},
+								"authentication": {
+									"type": "object",
+									"properties": {
+										"httpBasic": {
+											"type": "object",
+											"properties": {
+												"username": {
+													"type": "string"
+												},
+												"password": {
+													"type": "string"
+												}
+											},
+											"required": ["username", "password"],
+											"additionalProperties": false
+										}
+									},
+									"additionalProperties": false
+								}
+							},
+							"required": ["method", "uri", "timeout", "authentication"],
+							"additionalProperties": false
+						}
+					},
+					"additionalProperties": false
+				},
+				"outputs": {
+					"type": "array",
+					"minItems": 1,
+					"items": {
+						"type": "object",
+						"properties": {
+							"schema": {
+								"type": "string",
+								"pattern": "^iglu:[a-zA-Z0-9-_.]+/[a-zA-Z0-9-_]+/[a-zA-Z0-9-_]+/[0-9]+-[0-9]+-[0-9]+$"
+							},
+							"json": {
+								"type": "object",
+								"properties": {
+									"jsonPath": {
+										"type": "string"
+									}
+								},
+								"required": ["jsonPath"],
+								"additionalProperties": false
+							}
+						},
+						"required": ["schema"],
+						"minProperties": 2,
+						"maxProperties": 2,
+						"additionalProperties": false
+					}
+				},
+				"cache": {
+					"type": "object",
+					"properties": {
+						"size": {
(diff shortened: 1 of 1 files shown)


schemas/com.snowplowanalytics.snowplow.enrichments/api_lookup_enrichment_config/jsonschema/1-0-0
```diff
@@ -0,0 +1,166 @@
...
+								},
+								"required": ["jsonPath"],
+								"additionalProperties": false
+							}
+						},
+						"required": ["schema"],
+						"minProperties": 2,
+						"maxProperties": 2,
+						"additionalProperties": false
+					}
```
alexanderdean: Can we assert that the outputs array must be non-empty?

chuwy: I thought about that, but consider API with only PUTs requests. Kind of logging for enrichment process.

chuwy: Uhm. No, I'm wrong. Current implementation just won't perform requests without outputs.

alexanderdean: Interesting thought - okay let's leave as is for now. We can always bump to a 2-0-0 if we decide to restrict in the future

alexanderdean: Okay, let's enforce the limit. We can always loosen it with a 1-0-1 in the future.

chuwy: Ok.

chuwy: Updated.

schemas/com.snowplowanalytics.snowplow.enrichments/api_request_enrichment_config/jsonschema/1-0-0
```diff
@@ -0,0 +1,167 @@
...
+					"type": "object",
+					"minProperties": 1,
+					"maxProperties": 1,
+					"properties": {
+						"http": {
+							"type": "object",
+							"properties": {
+								"method": {
+									"type": "string",
+									"enum": ["GET", "POST", "PUT", "DELETE"]
```
alexanderdean: Do we really support all of these methods @chuwy ?

chuwy: @alexanderdean [yes](https://github.com/snowplow/snowplow/blob/feature/api-lookup/3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/HttpClient.scala#L64-L84), and actually [all](http://doc.akka.io/api/akka-stream-and-http-experimental/2.0/index.html#akka.http.scaladsl.model.HttpMethods$) other valid HTTP methods will work.

alexanderdean: Right - I guess the follow-on question is whether we want to support all of those. `GET` of course, `POST` and `PUT` are probably fine - for the case of a badly constructed (i.e. not very RESTful) API. `DELETE` I think we definitely don't want to support - in the case where this enrichment is accidentally configured with API credentials that allow all write operations, a `DELETE` here could let a bad actor who only has access to the Snowplow configuration file to delete data in a source system.

chuwy: Some thoughts on it: 1) hypothetically any method can result in bad behavior, even OPTION; 2) it's really hard to make a typo here, writing DELETE instead of GET; 3) it's very rare thing for APIs to support DELETE method properly 4) it's just a matter of feature fullness, if some one write API which support it, why we should forbid restrict it.
So if we're going to remove it, I think good reason would be because DELETE is useless, and not because it is harmful.

alexanderdean: 1) Sure - but if a `GET` deletes data, that's the responsibility of the API author; they can't say we've done anything wrong on our side 2) I'm referring to a bad actor - e.g. a disgruntled analyst who has the ability to update Snowplow enrichments 3) sure - but if somebody has implemented `DELETE` correctly, then it will delete data 4) see 2)

The wider point is: HTTP methods which exist to delete data shouldn't be supported in an enrichment which is meant to be read-only.

alexanderdean:
Pushing back to next release...

alexanderdean:
Hey @chuwy - if you can remove `DELETE` from the supported methods, I think we can merge this and get R49 out...

chuwy:
Sure!

chuwy:
And, yes not deleting data in enrichment process makes sense, as it goes against the core idea.

chuwy:
DELETE deleted.

schemas/com.snowplowanalytics.snowplow.enrichments/api_request_enrichment_config/jsonschema/1-0-0
```diff
@@ -0,0 +1,167 @@
...
+									"maximum": 60000
+								},
+								"authentication": {
+									"type": "object",
+									"properties": {
+										"httpBasic": {
+											"type": "object",
+											"properties": {
+												"username": {
+													"type": ["string", "null"],
```
chuwy: @alexanderdean just to clarify. Remind me what we decided about optional username/password? Basic auth is basically header with `base64(concatenate(username, ':', password))` so in our case `null` is equal to empty string. As it will be `concatenate(username, ':', '')`. So does it make sense to use `null` at all?

schemas/com.snowplowanalytics.snowplow.enrichments/api_request_enrichment_config/jsonschema/1-0-0
```diff
@@ -0,0 +1,167 @@
...
+									}
+								},
+								"additionalProperties": false
+							},
+							"json": {
+								"type": "object",
+								"properties": {
+									"field": {
+										"type": "string",
+										"enum": ["unstruct_event", "contexts", "derived_contexts"]
```
chuwy: I just realized `contexts` is pretty confusing here. May be we should change that to `custom_contexts`?

alexanderdean: I agree it's a bit confusing, but I think it's `contexts` because that's the column name. I think keeping it as the column name is safest for now...

chuwy: Makes sense.

schemas/com.snowplowanalytics.snowplow.enrichments/api_request_enrichment_config/jsonschema/1-0-0
alexanderdean: What does the relevant RFC say about the lack of a username or password? Does it refer to a null, or an empty string, or a missing value?

chuwy: I [can't see](https://tools.ietf.org/html/rfc2617) anything about lack of username or password. I suppose by-design both fields are required. Anyway, if I implement this with null, akka-http require both to be set, so I'm using empty string anyway.

alexanderdean:
```
To receive authorization, the client sends the userid and password,
   separated by a single colon (":") character, within a base64 [7]
   encoded string in the credentials.

      basic-credentials = base64-user-pass
      base64-user-pass  = <base64 [4] encoding of user-pass,
      user-pass   = userid ":" password
      userid      = *<TEXT excluding ":">
      password    = *TEXT

   Userids might be case sensitive.
```

So I see `*TEXT` - therefore "" is more valid I think than null. Let's remove null.

alexanderdean:
Merged into release branch now, closing

PR #304, opened 2016-02-23
Added com.snowplowanalytics.iglu/resolver-config/jsonschema/1-0-1 with apikey

chuwy:
@alexanderdean could you take a look? I decided to not add `format: uuid` because it can be misleading. JSON Schema validate it with regex including capitals, while for iglu scala repository capitals will be illegal.


alexanderdean:
Looks good to me! @fblundun can you take a look too...

fblundun:
This looks good to me.

alexanderdean:
Merged into release branch now, closing

PR #313, opened 2016-03-11
Allow for users to pass additional metadata with their Mandrill events

yalisassoon:
We've seen from one of our clients that it's possible to create your own additional metadata fields in Mandirll. When this is done that additional data is included as extra properties in the metadata object. Previously this object had `additionalProperties` set to `False`. This PR patches that so that they're set to true - preventing the events from failing validation for users who have setup their own metadata fields.


alexanderdean:
Scheduling into next release

PR #319, opened 2016-04-11
Patched com.snowplowanalytics.snowplow.enrichments/api_request_enrichment_config/jsonschema/1-0-0

### LINKED ISSUE #316: Patch com.snowplowanalytics.snowplow.enrichments/api_request_enrichment_config/jsonschema/1-0-0 to add JSONPath validation
Unfortunately, we cannot provide any strict JSONPath validation and I decided to not short-circuit enrichment on parsing invalid JSONPath as it won't give user any helpful error message.

Hence, I think it would be good to add at least `"pattern": "^\\$.*$"` (I think it is always valid for our use case), so users won't at least confuse JSONPath with `jq` syntax or similar tools.

chuwy:
...to add JSONPath validation (closes #316)


alexanderdean:
Merged into final release PR, closing

PR #324, opened 2016-04-19
Contexts for TrialReach events

swanandgore:
I need to add a couple of contexts to augment TrialReach's Snowplow events.


alexanderdean:
Closing per the support thread, thanks Swanand.

Issue #54, opened 2014-10-18
Schema for a com.zendesk/ticket_closed event


alexanderdean:

alexanderdean:
We are taking a different approach, closing.

PR #340, opened 2016-05-10
HdfsRepository supports #54

liningalex:


alexanderdean:
This is widening a type so should be a new schema (same contents) with version 1-1-0 - @chuwy do you agree?

chuwy:
@alexanderdean yes, I agree - all prior data will be valid against new schema, some new data can be invalid against old schema.
@liningalex we're using SchemaVer to versioning our configuration (you can find out more [here](http://snowplowanalytics.com/blog/2014/05/13/introducing-schemaver-for-semantic-versioning-of-schemas/) if interested). It means by introducing some changes to configuration you will need to add new schema with bumped version, leaving current versions as is. So, could you please repush your Schema (content is perfect) as new `1-1-0` version (in filename and `$.self.version` in content). Sorry for not letting you know before.

liningalex:
Done:-) please review. thanks.

liningalex:
Hi Anton, The change was done a week ago, could you review again :-).
thanks. Lining

chuwy:
Hello @liningalex. Sorry for delay. This should be a single schema file named `1-1-0` and based on our current `1-0-1` (with your addition of course). And also don't forget please to bump version in file itself (by `$.self.version` path), now it is incorrect `1-0-0`. Thanks!

liningalex:
Hi Anton, I pushed the changes, could you review again :-) Thanks.

On Mon, May 16, 2016 at 9:11 PM, Anton Parkhomenko <notifications@github.com

chuwy:
Hi @liningalex. 
We're using `$.self` object to describe *current* Schema (resolver-config). 
URI in `$.$schema` (`http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#`) used to describe all Self-describing Schemas ([link](http://tools.ietf.org/html/draft-zyp-json-schema-04#section-6)) and it should remain unchanged.
So, in your PR only version by path **`$.self`** should be bumped. Version in `$.$schema` need to always remain `1-0-0`.
Also, could you please use tabs instead of spaces to indent JSON to be consistent with rest of Schemas. 
Thank you very much.

liningalex:
HI Anton, could you please review again. many thanks for many reviews :-)

On Wed, May 25, 2016 at 1:09 AM, Anton Parkhomenko <notifications@github.com

chuwy:
Hey @liningalex! That is exactly what we need! I'll add few more minor notes to https://github.com/snowplow/iglu-scala-client/pull/55 to prepare it for merge. Thanks again!

alexanderdean:
@liningalex has signed our CLA.

alexanderdean:
Happy to keep open...

PR #346, opened 2016-05-18
Patch com.snowplowanalytics.iglu/jsonschema/resolver-config to forbid invalid priority and cacheSize

chuwy:
This PR patches both resolver-config Schemas to restrict priority and cacheSize to integer and also sets cacheSize minimum to 0.


alexanderdean:
Cherry-picked, closing

PR #354, opened 2016-06-17
Patch com.amazon.aws.lambda/s3_notification_event/jsonschema/1-0-0

ninjabear:
Remove upper limit on 'size' field as this is in bytes


alexanderdean:
Pushing back as wasn't in R60

alexanderdean:
This was cherry-picked in.

PR #361, opened 2016-06-25
Added Android application schemas

### LINKED ISSUE #360: Add com.snowplowanalytics.snowplow/application_error/jsonschema/1-0-0

alexanderdean:
…-0 (closes #360)

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/application_background/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow/application_background/jsonschema/1-0-0
@@ -1,0 +1,20 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for an application background event",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow",
+    "name": "application_background",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+
+  "type": "object",
+  "properties": {
+    "backgroundIndex": {
+      "type": "integer",
+      "minimum": 0,
+      "maximum": 2147483647
+    }
+  },
+  "additionalProperties": false
+}
--- a/schemas/com.snowplowanalytics.snowplow/application_error/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow/application_error/jsonschema/1-0-0
@@ -1,0 +1,77 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for an application error",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow",
+    "name": "application_error",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+
+  "type": "object",
+  "properties": {
+    "programmingLanguage": {
+      "enum": [ 
+        "JAVA",
+        "SCALA",
+        "KOTLIN",
+        "GROOVY",
+        "RUBY",
+        "GOLANG",
+        "JAVASCRIPT",
+        "PHP",
+        "PYTHON",
+        "OBJECTIVEC",
+        "SWIFT",
+        "C",
+        "CPLUSPLUS",
+        "CSHARP",
+        "ACTIONSCRIPT",
+        "LUA",
+        "RUST",
+        "HASKELL",
+        "CLOJURE",
+        "ERLANG",
+        "ELIXIR",
+        "CRYSTAL",
+        "PONY",
+        "NIM"
+      ]
+    },
+    "message": {
+      "type": "string",
+      "maxLength": 2048
+    },
+    "threadName": {
+      "type": ["string", "null"],
+      "maxLength": 1024
+    },
+    "threadId": {
+      "type": ["integer", "null"],
+      "minimum": 0,
+      "maximum": 2147483647
+    },
+    "stackTrace": {
+      "type": ["string", "null"],
+      "maxLength": 8192
+    },
+    "lineNumber": {
+      "type": ["integer", "null"],
+      "minimum": 0,
+      "maximum": 2147483647
+    },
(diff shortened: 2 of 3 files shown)


alexanderdean:
This looks good - but I think it would be worth doing a timeboxed spike on https://github.com/snowplow/snowplow-javascript-tracker/issues/16 before settling on this schema. In particular, I'm interested in whether we are going to have separate app_crash vs general error/exception, whether we are going to need language/platform specific error schemas etc.

We don't have to finalize all these decisions before this release, but at least if we've looked at Android and JS (which are pretty different), we'll have a more informed view from the start...

jbeemster:
I have had a quick look through some of the other error trackers and there is scope to flesh out the fields a touch, however I feel like things such as in the GA Tracker with PageURL that error occurred on can be handled by appending this schema as a context to a `page_view` event - rather than building many of these schemas we keep this one quite compact and modular.

From Airbrake there are a few interesting fields that we could add (https://airbrake.io/docs/#create-notice-v3) but it feels like most of the `error.*` fields are just extracted data points from the actual Stack Trace.  Which we could either do in the Tracker itself or as part of a custom enrichment where we pull apart the StackTrace to find interesting information in it such as the line the exception came from etc.

That being said it would be a bit of a nicer data-set if we had the stack trace decompiled into important components and would make using the data a lot easier.

alexanderdean:
Hey @jbeemster - makes sense. With an exception in the JS Tracker, we would do it the other way round: we would fire the exception event but still populate the `page_url` and `page_title` fields (which will move out into a page context at some point anyway).

I agree that the Airbrake error fields look a bit like premature enrichment. I guess the only advantage of that logic living in trackers is if you need proglang-specific handling of each error type - so the Snowplow Ruby Tracker knows how to parse a Ruby stacktrace, while a Snowplow Android Tracker knows how to parse an Android stacktrace.

I think it means as a minimum if we don't do in-tracker extraction, we need to add an enum to the event of the proglang that generated the stacktrace. We shouldn't have to parse the tracker version to try to figure that out.

Airbrake and GA seem to use the language of errors and exceptions - isn't application *crash* overly specific as a name?

jbeemster:
Agreed on the proglang specifics - will add that as a field!

The name is a bit too specific, it also assumes that they have no handling of the exception to attempt recovery.  We could change the name to be `uncaught_exception` to match exactly what it is actually grabbing?

alexanderdean:
Shall we just call it `application_error`? Some (most?) proglangs don't have exception catching semantics.

jbeemster:
That works for me!  For something cross platform we probably don't want anything to specific...

alexanderdean:
Cool!

jbeemster:
Hey @alexanderdean have added the other two schemas in here as well to simplify things.  The Redshift ddls and jsonpaths have been committed over here: https://github.com/snowplow/snowplow/tree/feature/application-schemas

I have uploaded these to SnowplowMini and asserted that events from the latest tracker do indeed make it through.  Anything more to be done on this?

alexanderdean:
Hey @jbeemster - these look good. A couple of things for the app error to make it a bit more futureproof:

* Make threadId an int not a short
* Put a (largeish) max length on message
* Put a (very large) max length on stack trace

jbeemster:
Hey @alexanderdean how about:

* 1kb (1024 max-length) for message
* 8kb (8192 max-length) for stack-trace
* 256b for thread-name

alexanderdean:
Sounds great! In the Android Tracker we should truncate those values to those lengths to prevent app errors from failing validation due to string lengths...

jbeemster:
Shall do!

schemas/com.snowplowanalytics.snowplow/application_background/jsonschema/1-0-0
```diff
@@ -0,0 +1,20 @@
...
+    "vendor": "com.snowplowanalytics.snowplow",
+    "name": "application_background",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+
+  "type": "object",
+  "properties": {
+    "index": {
+      "type": "integer",
```
alexanderdean: @jbeemster - if I search for "Android index" or "Android application index" nothing comes up. What are these indexes?

jbeemster: The index is just the amount of times it has gone foreground or background, nothing to do with Android specifically.  Maybe count would be a better name?

alexanderdean: Is it the number of times it has gone foreground or background for each event, or is it the number of times it has gone foreground for the foreground event, and the number of times it has gone background for the background event?

jbeemster: Every time it goes foreground or background we iterate a number and plug it in there.  So each foreground and background event simply iterates by 1 per event.

jbeemster: From your original comment: https://github.com/snowplow/iglu-central/issues/362#issuecomment-228703645

alexanderdean: Right - maybe `lifecycleEventIndex`? Let's discuss tomorrow.

jbeemster: To my mind it going to the background/foreground is a lifecycle event in itself so naming it lifecycleEventIndex feels a touch redundant.

Sounds good!

alexanderdean: Right - we need to make clear that this `index` isn't a `fragmentIndex` or a `backgroundIndex`...

jbeemster: Maybe just `eventIndex` then.  It states more that it an index to do with the wrapping event rather than anything else to do with fragments/activities etc.

alexanderdean:
Cherry-picked over, closing

PR #386, opened 2016-07-28
Fix/lint

chuwy:
Contains all fixes we found using `igluctl`, CI with `igluctl` and CHANGELOG (I assume this is r60).

I edited CI script to using `igluctl` in the same manner we used fge validator - running it for each schema one by one, but if we'll change it to process whole repository - it will run much faster (2sec vs ~2min).


chuwy:
@alexanderdean 

2. No reasons! Except absence of fancy colored output.
3. Only to deploy it locally and run SCE, SHE tests with it in resolver config. That is how I found geolocation context bug, but we need to at least check we have tests for all touched enrichments (this PR mostly about enrichments Schemas).

alexanderdean:
Hey @chuwy - awesome, let's go with the 2secs option.

Would you be able to *deploy it locally and run SCE, SHE tests with it in resolver config*? I'm not so worried about enrichment test coverage - from inspection I can see that the changes in the PR are consistent with each other: if the test suites pass, then it means there is nothing systemically wrong and we can say that R60 is safe to release.

chuwy:
@alexanderdean great! Will do it!

chuwy:
@alexanderdean I rewrote CI script to perform faster.

chuwy:
30s vs ~2min.

PR #403, opened 2016-09-21
Add elementContent to JSON path and schema

miike:
Element content is missing from the JSON path and schema for link_click_1. This property was added in the 1-0-1 [JSON schema](https://github.com/snowplow/iglu-central/blob/edb500bd34eabdd41684e16d1698f0bae6baa8b7/schemas/com.snowplowanalytics.snowplow/link_click/jsonschema/1-0-1).

I've set the length at 255 but happy to change to whatever would be more appropriate.


miike:
Related to https://github.com/snowplow/snowplow/issues/1149, I'll close this as it sounds like this will break loads for users using the existing 1-0-0 schema.

alexanderdean:
Yep, I'm going to write a Discourse post to clarify all this soon...

PR #419, opened 2016-10-21
Factotum webhooks

ninjabear:


ninjabear:
I still need to add in the DDL/jsonpaths for these, but are the schemas what you were expecting @alexanderdean ?

ninjabear:
I have added the DDL/JSON Paths for this now too

PR #421, opened 2016-10-25
Patch Factotum job_update and task_update (closes #420)

### LINKED ISSUE #420: Patch com.snowplowanalytics.factotum/job_update/jsonschema/1-0-0
Both of these schemas need the following changes:

1. The max size of a Factfile must be >5000 bytes (in Redshift)
2. The max size of the `stderr` and `stdout` fields in `taskStates` must be no larger than 10,000 (Factotum will trim to this size).
3. The max size of the `task_states` field in redshift must be >5000 (preferably the maximum available size)

ninjabear:
Set stderr and stdout maxLength to 10_000
Set Factfile max size in Redshift to "max" (both schemas/tables)
Set "taskStates" max size in Redshift to "max" (both schemas/tables)


ninjabear:
Hey @alexanderdean ! Is this what you were expecting? I think because of the 65kb limit of `VARCHAR` fields in a Redshift row Factotum events may still eventually have some problems getting into Redshift. So this isn't a perfect solution, but it is a solution that'll work for all jobs I can think of right now.

alexanderdean:
Thanks @ninjabear - this is good - it's a manual implementation of https://github.com/snowplow/iglu/issues/220.

Only thing - please split this into 6 commits for consistency. Sorry to be a stickler about this.

PR #495, opened 2017-02-15
Optimizely X summary schema

miike:
Adds the Optimizely X summary schema. Companion PR to https://github.com/snowplow/snowplow-javascript-tracker/pull/552


snowplowcla:
@miike has signed the Software Grant and Corporate Contributor License Agreement

BenFradet:
@alexanderdean getting this on your radar since this is needed for the js tracker release.

alexanderdean:
Assuming this is fine but just getting reviewers to say so...

alexanderdean:
Hmm. I wonder if we should rename this to:

    com.optimizely.optimizelyx/summary/jsonschema/1-0-0 

This feels cleaner (separate namespace for the OptimizelyX product)...

Thoughts?

miike:
I think this comes down to

>Does the product for an event belong in the vendor field or the event field?

I don't think there's necessarily a correct answer but I'd agree and lean towards including it in the vendor field rather than the event name - even if this makes it something closer to a namespace rather than strictly a vendor.

alexanderdean:
Makes sense @miike - I will move the schema into `com.optimizely.optimizelyx`. You are right, we use this field more as a namespace than as a strict vendor (see e.g. all the `com.snowplowanalytics.` folders in Iglu Central)...

miike:
I've modified this PR to move the event name from `optimizelyx_summary` to `summary` and event vendor from `com.optimizely` to `com.optimizely.optimizelyx`. I've also updated the companion PR to reflect this path change https://github.com/snowplow/snowplow-javascript-tracker/pull/567

BenFradet:
@miike integrated your changes in snowplow/snowplow-javascript-tracker#562

alexanderdean:
Thanks @miike , thanks @BenFradet ! This Iglu Central release is going out today :ship:

alexanderdean:
Split into three commits and cherry-picked into #544

PR #596, opened 2017-06-07
Release 84

dilyand:


schemas/com.google.analytics.measurement-protocol/app_tracking/jsonschema/1-0-0
```diff
@@ -0,0 +1,31 @@
...
+    },
+    "applicationVersion": {
+      "type": ["string", "null"],
+      "maxLength": 100
+    },
+    "applicationInstallerId": {
+      "type": ["string", "null"],
+      "maxLength": 150
+    }
+  },
```
BenFradet: since this seems to be modeling a `screenview` hit type, I'm wondering why there is no:

```json
"hitType": {
  "enum": [
    "sceenview"
  ]
}
```

like in the other schemas representing hit types (e.g. `item` or `page_view`)

alexanderdean:
I'd go further than @BenFradet's comment - I don't think the current approach of:

```
+  "properties": {
+    "hitType": {
+      "enum": [
+        "pageview"
+      ]
+    },
```

in each entity is correct. Rather, there should be a single `hitType` in the hit entity holding an enum of all possible values. This is the only thing that makes sense from an analytics perspective.

schemas/com.google.analytics.measurement-protocol/screen_view/jsonschema/1-0-0
```diff
@@ -0,0 +1,25 @@
...
+      ]
+    },
+    "screenName": {
+      "type": "string",
+      "maxLength": 2048
+    }
+  },
+  "required": ["hitType", "screenName"],
+  "additionalProperties": false
+}
```
BenFradet: Unclear as to how this relates to the app_tracking schema which seems to contain all data regarding a `screenview` hit type

alexanderdean:
@dilyand - please can we make a couple of changes:

* Rename `traffic_sources` to `traffic_source`, given it's a singular entity (the source of this traffic)
* Split out `hit` into `hit` plus `general` entities, per the spreadsheet (it's not our job to start merging the MP buckets into one)

Please also add:

* `com.google.analytics/undocumented`
* `com.google.analytics/private`

alexanderdean:
Also can you rebase the Patch commits into the original Add commits. It doesn't make sense to patch something that hasn't been yet been released, in a single release.

alexanderdean:
Also can we rename the `linkId` entity to just `link` (otherwise it's `linkId.id` which is weird)

alexanderdean:
@BenFradet can provide you @dilyand with "safe" data types for `com.google.analytics/undocumented` and `com.google.analytics/private`

BenFradet:
|field|type|example value|schema|
|:-:|:-:|:-:|:-:|
|a|string|967390869|undocumented|
|jid|string|74167946|undocumented|
|gjid|string|457285730|undocumented|
|_v|string|j55|private|
|_s|int|1|private|
|_u|string|SCCAAAADI~|private|
|_gid|string|1660085573.1496766859|private|
|_r|int|1|private|

There could be others though.

alexanderdean:
Also rename `app_tracking` to `app` for consistency with `event`

alexanderdean:
Thanks @BenFradet - @dilyand for safety with the internal/advertising fields, I would go string:256 for the strings and int8 (bigint) for the numbers.

dilyand:
@alexanderdean Just as a matter of process, I plan to:

- go through each schema and patch with discussed changes
- have one commit per patch
- when all patches are done, I'll squash the commits so we only have one commit per schema ("Add com.google.analytics.measurement-protocol/schema_name/jsonschema/1-0-0")

Is that okay or are they any other guidelines I should be aware of?

alexanderdean:
It's up to you - I guess your approach makes it easier to revert some of the changes if we have second thoughts...

schemas/com.google.analytics.measurement-protocol/product_custom_metric/jsonschema/1-0-0
```diff
@@ -0,0 +1,28 @@
...
+      "minimum": 1,
+      "maximum": 200
+    },
+    "metricIndex": {
+      "type": "integer",
+      "minimum": 1,
+      "maximum": 200
+    },
+    "value": {
+      "type": ["number", "null"]
```
BenFradet: shouldn't this be an integer according to: https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#pr_cm_ ?

schemas/com.google.analytics.measurement-protocol/hit/jsonschema/1-0-0
```diff
@@ -0,0 +1,30 @@
...
+  "self": {
+    "vendor": "com.google.analytics.measurement-protocol",
+    "name": "hit",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+
+  "type": "object",
+  "properties": {
+    "hitType": {
```
alexanderdean: Ouch, agree

dilyand:
All fixed. Thanks @BenFradet , @alexanderdean !

kymc:
Is this still alive/planned? It's a really great idea...

alexanderdean:
Yes, it will happen! Hopefully soon, we're just having some scheduling congestion getting it back in progress

PR #627, opened 2017-08-12
Adding 'subaccount' property for all Mandrill events

acgray:
The subaccount property is only currently included in the message_sent schema but appears in all Mandrill-sent events.  This currently causes these events to fail enrichment.


snowplowcla:
@acgray has signed the Software Grant and Corporate Contributor License Agreement

acgray:
Replaced by #628

PR #628, opened 2017-08-14
Fix/mandrill subaccount

acgray:
The subaccount property is only currently included in the message_sent schema but appears in all Mandrill-sent events. This currently causes these events to fail enrichment.


snowplowcla:
@acgray has signed the Software Grant and Corporate Contributor License Agreement

BenFradet:
This needs to be in the same release than #745

BenFradet:
incorporated into #753, thanks!

alexanderdean:
Many thanks @acgray !

PR #647, opened 2017-10-06
Ssh tunnel

chuwy:
Adds following property to PostgreSQL and Redshift:

```json
        "tunnel": {
            "type": ["object", "null"],
            "properties": {
                "host": {
                    "type": "string"
                },
                "port": {
                    "type": "integer",
                    "minimum": 1,
                    "maximum": 65535
                },
                "user": {
                    "type": "string"
                },
                "passphrase": {
                    "type": "string"
                },
                "sshKeyParameter": {
                    "type": "string"
                },
                "localPort": {
                    "type": "integer",
                    "minimum": 1,
                    "maximum": 65535
                },
                "destinationHost": {
                    "type": "string"
                },
                "destinationPort": {
                    "type": "integer",
                    "minimum": 1,
                    "maximum": 65535
                }
            },
            "required": ["host", "port", "user", "localPort", "destinationHost", "destinationPort"]
        }
```

Configuration will look like following:

```json
{
    "schema": "iglu:com.snowplowanalytics.snowplow.storage/redshift_config/jsonschema/2-0-0",
    "data": {
        "name": "AWS Redshift enriched events storage",
        "host": "localhost",
        "database": "snowplow",
        "port": 15151,
        "sslMode": "DISABLE",
        "username": "loader",
        "password": "Supersecret1",
        "roleArn": "arn:doesnmatter",
        "schema": "atomic",
        "maxError": 1,
        "compRows": 20000,
        "purpose": "ENRICHED_EVENTS",
        "tunnel": {
            "host": "bastion.acme.com",
            "port": 22,
            "user": "snowplow-loader",
            "localPort": 15151,
            "sshKeyParameter": "snowplow.rdbloader.redshift.key",
            "hostDestination": "10.0.0.11",
            "portDestination": 5439
        }
    }
}
```

@jbeemster proposed to omit `localPort` and choose arbitrary, but I think this would more explicit and also afraid EMR/Hadoop/etc might take this port eventually.

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow.storage/postgresql_config/jsonschema/1-1-0
+++ b/schemas/com.snowplowanalytics.snowplow.storage/postgresql_config/jsonschema/1-1-0
@@ -1,0 +1,122 @@
+{
+    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "description": "Snowplow PostgreSQL storage configuration",
+    "self": {
+        "vendor": "com.snowplowanalytics.snowplow.storage",
+        "name": "postgresql_config",
+        "format": "jsonschema",
+        "version": "1-1-0"
+    },
+    "type": "object",
+    "properties": {
+        "name": {
+            "type": "string"
+        },
+        "host": {
+            "type": "string"
+        },
+        "database": {
+            "type": "string"
+        },
+        "port": {
+            "type": "integer",
+            "minimum": 1,
+            "maximum": 65535
+        },
+        "sslMode": {
+            "type": "string",
+            "enum": ["DISABLE", "REQUIRE", "VERIFY_CA", "VERIFY_FULL"]
+        },
+        "schema": {
+            "type": "string"
+        },
+        "username": {
+            "type": "string"
+        },
+        "password": {
+            "type": ["string", "object"],
+            "properties": {
+                "ec2ParameterStore": {
+                    "type": "object",
+                    "properties": {
+                        "parameterName": {
+                            "type": "string"
+                        }
+                    },
+                    "required": ["parameterName"]
+                }
+            },
+            "required": ["ec2ParameterStore"]
+        },
+        "sshTunnel": {
+            "type": ["object", "null"],
+            "properties": {
+                "bastion": {
+                    "type": "object",
+                    "properties": {
+                        "host": {
+                            "type": "string"
+                        },
+                        "port": {
+                            "type": "integer",
+                            "minimum": 1,
+                            "maximum": 65535
+                        },
+                        "user": {
+                            "type": "string"
+                        },
+                        "passphrase": {
+                            "type": "string"
+                        },
+                        "key": {
+                            "type": "object",
+                            "properties": {
+                                "ec2ParameterStore": {
+                                    "type": "object",
+                                    "properties": {
+                                        "parameterName": {
+                                            "type": "string"
+                                        }
+                                    },
+                                    "required": ["parameterName"]
+                                }
+                            },
+                            "required": ["ec2ParameterStore"]
+                        }
+                    },
+                    "required": ["host", "port", "user"]
+                },
+                "destination": {
+                    "type": "object",
+                    "properties": {
+                        "host": {
+                            "type": "string"
+                        },
+                        "port": {
+                            "type": "integer",
+                            "minimum": 1,
+                            "maximum": 65535
+                        }
+                    },
+                    "required": ["host", "port"]
+                },
+                "localPort": {
+                    "type": "integer",
+                    "minimum": 1,
+                    "maximum": 65535
+                }
+            },
+            "required": ["bastion", "destination", "localPort"]
+        },
+        "id": {
+            "type": "string",
+            "format": "uuid"
+        },
+        "purpose": {
+            "type": "string",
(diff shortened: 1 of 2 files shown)


alexanderdean:
> @jbeemster proposed to omit localPort and choose arbitrary

Isn't `data.tunnel.localPort` just the same as `data.port`? Is there a situation where they could be different?

alexanderdean:
@chuwy can you clarify what these parameters signify:

* `sshKeyParameter`
* `passphrase` - which seems to be missing from your example (but present in schema)

Is there a reason why you use `hostDestination` rather than `destinationHost` or better `destination.host`?

chuwy:
> Isn't data.tunnel.localPort just the same as data.port

They're same. I cannot come up with example where they're different right now, but my point to have both is that root config and `tunnel` are separate entities - former represents everything necessary (and nothing more) to create JDBC connection, later represents everything (and nothing more) to establish SSH tunnel. If we're going to merge them into single property - then I also cannot imagine a situation where `data.host` could be non-`localhost`. So, we either do everything explicit or simple.

`sshKeyParameter` is EC2 Systems Manager parameter name which stores SSH key.

`passphrase` is user's password. AFAIK it is possible to require both password and SSH key to connect.

> Is there a reason why you use hostDestination rather than destinationHost or better destination.host?

No, I'm happy to change.

chuwy:
My another "simple" option was:

```json
{
    "schema": "iglu:com.snowplowanalytics.snowplow.storage/redshift_config/jsonschema/2-0-0",
    "data": {
        "name": "AWS Redshift enriched events storage",
        "host": "10.0.0.11",
        "database": "snowplow",
        "port": 5439,
        "sslMode": "DISABLE",
        "username": "loader",
        "password": "Supersecret1",
        "roleArn": "arn:doesnmatter",
        "schema": "atomic",
        "maxError": 1,
        "compRows": 20000,
        "purpose": "ENRICHED_EVENTS",
        "tunnel": {
            "host": "bastion.acme.com",
            "port": 22,
            "sshKeyParameter": "snowplow.rdbloader.redshift.key",
            "user": "snowplow-loader"
        }
    }
}
```

alexanderdean:
Hey @chuwy - I prefer the explicit version then.

I think for consistency though we need more like:

```json
            "local": {
              "host": "bastion.acme.com",
              "port": 22
            }
            "user": "snowplow-loader",
            "localPort": 15151,
            "sshKeyParameter": "snowplow.rdbloader.redshift.key",
            "destination": {
              "host": "10.0.0.11",
              "port": 5439
            }
```

Also I think `"sshKeyParameter"` is both too vague (as a parameter name) and too specific (in terms of assuming that this tunnel will always use EC2 Systems Manager. What about something like:

```json
            "key": {
               "ec2ParameterStore": {
                 "parameterName": "snowplow.rdbloader.redshift.key"
               }
             }
```

jbeemster:
Hey @alexanderdean @chuwy - there is one other attribute missing from the parameter store which is the ARN of the KMS key used to encrypt the value.  It is likely that we might not want to use the default key that comes with the sub-account.

chuwy:
Hey @jbeemster, no I was quite surprised as well, but request doesn't need (and cannot) to contain KMS key: `new GetParameterRequest().withName(name).withWithDecryption(true)`. KMS key associated with each `SecurityString` is publicly available in Parameter Store.

chuwy:
@alexanderdean ok, I like it more. Though I think `bastion` would be more descriptive than `local` and `user` and `passphrase` should also belong to `bastion`. 

Also PR will include changes required for https://github.com/snowplow/snowplow-rdb-loader/issues/19, so password could be either string or:

```json
            "password": {
              "ec2ParameterStore": {
                "parameterName": "snowplow.rdbloader.redshift.password"
              }
            }
```

chuwy:
Configs in current version could look like following:

```json
{
    "schema": "iglu:com.snowplowanalytics.snowplow.storage/redshift_config/jsonschema/2-1-0",
    "data": {
        "name": "AWS Redshift enriched events storage",
        "host": "localhost",
        "database": "snowplow",
        "port": 15151,
        "sslMode": "DISABLE",
        "username": "loader",
         "password": {
             "ec2ParameterStore": {
                 "parameterName": "snowplow.rdbloader.redshift.key"
             }
         },
        "roleArn": "arn:doesnmatter",
        "schema": "atomic",
        "maxError": 1,
        "compRows": 20000,
        "purpose": "ENRICHED_EVENTS",
        "tunnel": {
            "bastion": {
                "host": "bastion.acme.com",
                "port": 22,
                "user": "snowplow-loader",
                "key": {
                     "ec2ParameterStore": {
                         "parameterName": "snowplow.rdbloader.redshift.key"
                     }
                }
            },
            "destination": {
                "host": "10.0.0.11",
                "port": 5439
            },
            "localPort": 15151,
        }
    }
}
```

alexanderdean:
I like it, would be good to get feedback from @BenFradet and @jbeemster too.

alexanderdean:
Yes, I think I prefer sshTunnel too...

chuwy:
Updated to `sshTunnel`.

alexanderdean:
Nudge @jbeemster

jbeemster:
@alexanderdean @chuwy - no other feedback from me!

alexanderdean:
Cherry-picked into https://github.com/snowplow/iglu-central/pull/648, closing

PR #366, opened 2016-07-04
Add json schemas for mailgun webhook

ronnyml:

PR #664, opened 2017-10-31
R82 - supporting snowplow r97 knossos

knservis:


schemas/com.mailgun/message_bounced/jsonschema/1-0-0
```diff
@@ -0,0 +1,62 @@
...
+		"X-Mailgun-Sid": {
+			"type": "string",
+			"pattern": "^([A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{4}|[A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)$"
+		},
+		"my_var_1": {
+			"type": "string"
+		},
+		"my-var-2": {
+			"type": "string"
+		}
```
BenFradet: what do those refer to?

schemas/com.mailgun/message_clicked/jsonschema/1-0-0
```diff
@@ -0,0 +1,75 @@
...
+		},
+		"device-type": {
+			"type": "string"
+		},
+		"my_var_1": {
+			"type": "string"
+		},
+		"my-var-2": {
+			"type": "string"
+		}
```
BenFradet: same here

alexanderdean:
This git commit looks incorrect:

    Add com.mailgun/message_opened/jsonschema/1-0-0 (closes #366) 

#366 is just a general PR, it's not the corresponding ticket name...

schemas/com.mailgun/message_clicked/jsonschema/1-0-0
knservis: Ok I found out where these come from. When you create a webhook with mailgun you are given the option to test it and see the result in the postbin. Unfortunately testing it creates some additional variables (my_var_1, and  my-var-2). Furthermore it does not contain some fields (h and message-id), and some of the documented optional fields (campaign-id, campaign-name, mailing-list, tag). I am going through those schemas again now.

alexanderdean: Hey @knservis - thanks for digging into this. So just to confirm, those my_var_1, and my-var-2 are Mailgun-authored fields (albeit test path only) - no operator involvement?

knservis: @alexanderdean that is correct for the test, although these extra variables could be anything (e.g. var-your-42-with-salt) as far as I can tell and that depends on the operator. It seems that the best way to handle those in the iglu schema would be with: 
```
 "additionalProperties": true
```

knservis: The same should also happen for the "attachment-x" field which in fact translates to "attachment-1", "attachment-2" et c.

alexanderdean: Thanks. What does "that depends on the operator" mean? What's an operator?

knservis: @alexanderdean yes

knservis: @alexanderdean *operator = integrator

jsonpaths/com.olark/transcript_1.json
```diff
@@ -0,0 +1,42 @@
...
+        "$.hierarchy.rootId",
+        "$.hierarchy.rootTstamp",
+        "$.hierarchy.refRoot",
+        "$.hierarchy.refTree",
+        "$.hierarchy.refParent",
+        "$.data.groups",
+        "$.data.id",
+        "$.data.items",
+        "$.data.kind",
+        "$.data.operators..*.emailAddress",
```
chuwy: I'm very surprised this was generated by Igluctl and don't think this is going to work.

schemas/com.mailgun/message_bounced/jsonschema/1-0-0
```diff
@@ -0,0 +1,97 @@
...
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "attachment-count": {
+      "description": "Number of attachments in attachment-x (where x is a number) fields.",
+      "type": "integer",
+      "maximum": 32767
+    },
+    "body-plain": {
```
chuwy: I'd still go with removing it from payload. If I understand it right - this will repeat multiple times (newsletter) and has no real value. I can be wrong.

schemas/com.mailgun/message_bounced/jsonschema/1-0-0
```diff
@@ -0,0 +1,97 @@
...
+      "description": "text/plain body of the message.",
+      "type": "string",
+      "maxLength": 524288
+    },
+    "event": {
+      "description": "Event name ('bounced').",
+      "type": "string",
+      "maxLength": 16
```
chuwy: Are we sure about such short length? Can't it be `enum` instead?

schemas/com.mailgun/message_bounced/jsonschema/1-0-0
```diff
@@ -0,0 +1,97 @@
...
+    },
+    "timestamp": {
+      "description": "Number of seconds passed since January 1, 1970 (see securing webhooks).",
+      "type": "integer",
+      "maximum": 2147483647
+    },
+    "token": {
+      "description": "Randomly generated string with length 50 (see securing webhooks).",
+      "type": "string",
+      "maxLength": 64
```
chuwy: If this has known length of 50, then we can use  `"minLength": 50`/`"maxLength": 50`. This will result in fixed-length `CHAR(50)`.

schemas/com.mailgun/message_bounced/jsonschema/1-0-0
```diff
@@ -0,0 +1,97 @@
...
+    },
+    "signature": {
+      "description": "String with hexadecimal digits generate by HMAC algorithm (see securing webhooks).",
+      "type": "string",
+      "pattern": "^([A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{4}|[A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)$",
```
chuwy: This is quite unreliable `base64` regexp. Description doesn't look like Base64.

schemas/com.olark/transcript/jsonschema/1-0-0
```diff
@@ -0,0 +1,202 @@
...
+      "maxLength": 128
+    },
+    "kind": {
+      "type": "string",
+      "maxLength": 64
+    },
+    "operators": {
+      "type": "object",
+      "properties": {
+        ".*": {
```
chuwy: Maybe you wanted to use `patternProperties` here? I don't think this is valid schema.

schemas/com.mailgun/message_clicked/jsonschema/1-0-0
```diff
@@ -0,0 +1,122 @@
...
+  "description": "Schema for a Mailgun message clicked event",
+  "self": {
+    "vendor": "com.mailgun",
+    "name": "message_clicked",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "body-plain": {
```
chuwy: Same about `body-plain`. Long, possibly unreadable (HTML), repeatable.

schemas/com.mailgun/message_clicked/jsonschema/1-0-0
```diff
@@ -0,0 +1,122 @@
...
+      "type": "string",
+      "maxLength": 64
+    },
+    "country": {
+      "description": "Two-letter country code (as specified by ISO3166) the event came from or ‘unknown’ if it couldn’t be determined.",
+      "type": "string",
+      "maxLength": 2
+    },
+    "device-type": {
+      "description": "Device type the link was clicked on. Can be ‘desktop’, ‘mobile’, ‘tablet’, ‘other’ or ‘unknown’.",
```
chuwy: It looks like complete list. We can use `enum` then.

schemas/com.mailgun/message_clicked/jsonschema/1-0-0
```diff
@@ -0,0 +1,122 @@
...
+      "format": "uri",
+      "maxLength": 8192
+    },
+    "user-agent": {
+      "description": "User agent string of the client triggered the event.",
+      "type": "string", 
+      "maxLength": 1024
+    }
+  },
+  "additionalProperties": true
```
chuwy: Nothing `required`?

schemas/com.mailgun/message_clicked/jsonschema/1-0-0
alexanderdean: Thanks, in that case I think we have to remove the my-vars and attachment-xs, and set additionalProperties: true...

jsonpaths/com.olark/transcript_1.json
knservis: What is igluctl supposed to create in this case (so that I can open an issue there)?

schemas/com.mailgun/message_bounced/jsonschema/1-0-0
knservis: I thought the same. As you say this will be almost the same in all messages and may cause other problems (e.g. kinessis choking on it).

schemas/com.mailgun/message_clicked/jsonschema/1-0-0
knservis: It may be ok to use an enum in this particular case as the data is unlikely to add new values soon (such as 'phablet') but I would like to understand all the uses of the data first. As far as I can tell this will make no difference to Redshift.

schemas/com.mailgun/message_bounced/jsonschema/1-0-0
knservis: Agreed. I am not sure why this was included.

schemas/com.mailgun/message_clicked/jsonschema/1-0-0
knservis: It is not clear to me what is required from the documentation.

schemas/com.olark/transcript/jsonschema/1-0-0
knservis: Yes, thank you.

schemas/com.mailgun/message_bounced/jsonschema/1-0-0
knservis: I thought about enum but:
a. This is a constant (i.e. it will always be 'bounced' in this case, so we may as well omit it. 
b. Enums have to be updated to add a value. 

The length is certainly sufficient.

schemas/com.mailgun/message_bounced/jsonschema/1-0-0
knservis: To be clear HMAC is often base64 encoded (and in this case it is), but I am not sure we need to check that here, rather than a proper validation (documented under securing webhooks). The question is do we need to check that here?

chuwy: Most likely because Schema Guru tries to recognize `base64`, but it inevitably results in false positives, especially for low-volume data.

schemas/com.mailgun/message_bounced/jsonschema/1-0-0
chuwy: Yeah, I agree. `enum`s are good, but need be used only when we're 100% sure.

alexanderdean: Yes, enums are fairly risky when we don't control the original data source (i.e. all webhooks)...

jsonpaths/com.olark/transcript_1.json
chuwy: I think it should fail with something like `Unsupported characters` error.

schemas/com.unbounce/event_context/jsonschema/1-0-0
```diff
@@ -0,0 +1,43 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for Unbounce event context (https://documentation.unbounce.com/hc/en-us/articles/203510044-Using-a-Webhook#content6)",
+  "self": {
+    "vendor": "com.unbounce",
+    "name": "event_context",
```
alexanderdean: Why is this called an "event_context"? Is this Unbounce jargon? In Snowplow, an event and a context are different things...

CHANGELOG
```diff
@@ -1,3 +1,18 @@
+Release 81 (2017-10-31)
+-----------------------
+Add com.mailgun/message_opened/jsonschema/1-0-0 (#665)
```
alexanderdean: Can we move this entry down to next to the other mailgun ones

schemas/com.mailgun/message_bounced/jsonschema/1-0-0
```diff
@@ -0,0 +1,92 @@
...
+    },
+    "Message-Id": {
+      "description": "The Id of the message that bounced.",
+      "type": "string",
+      "maxLength": 1024
+    },
+    "timestamp": {
+      "description": "Number of seconds passed since January 1, 1970 (see securing webhooks).",
+      "type": "string",
+      "maxLength": 16
```
alexanderdean: I believe each of the Mailgun schemas has this `timestamp` field. In the Scala adapter, we should be proactively changing it into a ISO8601 datetime (and updating the schema to match). Then it will appear in Redshift as a first-class `TIMESTAMP` column.

schemas/com.unbounce/event_context/jsonschema/1-0-0
knservis: The original committer had two schemas for unbounce, an "event_context" and and "event_payload", but I do not know the reason he needed two (the data.json field can contain anything). I believe we can rename it to something more suitable (e.g. "form_post_event"). Any suggestions?

schemas/com.olark/transcript/jsonschema/1-0-0
```diff
@@ -0,0 +1,209 @@
...
+          "name": {
+            "type": "string",
+            "maxLength": 128
+          },
+          "id": {
+            "type": "string",
+            "maxLength": 128
+          }
+        },
+        "additionalProperties": false
```
chuwy: Just in case - since we're inside the `array`, these `maxLength` and `additionalProperties` add almost no value - they cannot be mapped to Redshift types.

schemas/com.olark/transcript/jsonschema/1-0-0
```diff
@@ -0,0 +1,209 @@
...
+        },
+        "browser": {
+          "type": "string",
+          "maxLength": 128
+        },
+        "fullName": {
+          "type": "string",
+          "maxLength": 256
+        },
+        "emailAddress": {
```
chuwy:
```json
{
  "format": "email",
  "maxLength": 256
}
```

https://github.com/snowplow/iglu/pull/258

schemas/com.olark/transcript/jsonschema/1-0-0
```diff
@@ -0,0 +1,209 @@
...
+      "patternProperties": {
+        ".*": {
+          "type": "object",
+          "properties": {
+            "username": {
+              "type": "string", 
+              "maxLength": 128
+            },
+            "emailAddress": {
+              "type": "string",
```
chuwy: `"format": "email"`

schemas/com.unbounce/event_context/jsonschema/1-0-0
```diff
@@ -0,0 +1,43 @@
...
+    },
+    "data.json": {
+      "type": "object",
+      "additionalProperties": true
+    },
+    "data.xml": {
+      "type": "string",
+      "description": "The submitted form data in XML format.",
+      "maxLength": 65535
+    }
```
chuwy: As far as I remember it just duplicates `data.json`?

schemas/com.mailgun/message_bounced/jsonschema/1-0-0
```diff
@@ -0,0 +1,92 @@
...
+    "name": "message_bounced",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "attachment-count": {
+      "description": "Number of attachments in attachment-x (where x is a number) fields.",
+      "type": "string",
+      "maxLength": 8
```
chuwy: So, can't we transform it to `integer` inside adapter?

schemas/com.mailgun/message_clicked/jsonschema/1-0-0
```diff
@@ -0,0 +1,116 @@
...
+    },
+    "device-type": {
+      "description": "Device type the link was clicked on. Can be ‘desktop’, ‘mobile’, ‘tablet’, ‘other’ or ‘unknown’.",
+      "type": "string",
+      "maxLength": 64
+    },
+    "domain": {
+      "description": "Domain that sent the original message.",
+      "type": "string",
+      "maxLength": 256
```
chuwy: `"format": "hostname"?

schemas/com.mailgun/message_clicked/jsonschema/1-0-0
```diff
@@ -0,0 +1,116 @@
...
+      "description": "The address of mailing list the original message was sent to.",
+      "type": "string"
+    },
+    "message-id": {
+      "description": "The Id of the message that bounced.",
+      "type": "string",
+      "maxLength": 1024
+    },
+    "recipient": {
+      "description": "Recipient who clicked.",
```
chuwy: Is it id or email?

schemas/com.mailgun/message_clicked/jsonschema/1-0-0
```diff
@@ -0,0 +1,116 @@
...
+      "type": "string",
+      "maxLength": 128
+    },
+    "tag": {
+      "description": "Message tag, if it was tagged. See Tagging.",
+      "type": "string"
+    },
+    "timestamp": {
+      "description": "Number of seconds passed since January 1, 1970 (see securing webhooks).",
+      "type": "string",
```
chuwy: Again, this look like something we can translate into more descriptive type.

schemas/com.unbounce/event_context/jsonschema/1-0-0
knservis: Correct.

schemas/com.mailgun/message_bounced/jsonschema/1-0-0
knservis: @chuwy I think everything comes to the adaptor as a string (This used to be an integer and the job would die before getting to the adaptor)

schemas/com.unbounce/event_context/jsonschema/1-0-0
alexanderdean: > I do not know the reason he needed two

I have a theory, which is that the original idea was to turn the data.json field into the event (any event), and then attach the rest of the fields to this event as an "event context". This would explain the weird name.

This is really powerful - it would mean that you can configure your form against a JSON Schema that you configure yourself.

@knservis - was there any code in the adapter that supports this theory?

schemas/com.mailgun/message_bounced/jsonschema/1-0-0
alexanderdean: > I think everything comes to the adaptor as a string

Right - but how it *leaves* the adaptor is at your discretion.

schemas/com.mailgun/message_clicked/jsonschema/1-0-0
alexanderdean: See my comment above on timestamp conversion.

schemas/com.olark/transcript/jsonschema/1-0-0
alexanderdean: But @chuwy they may be useful for some other storage target we support in the future?

schemas/com.mailgun/message_bounced/jsonschema/1-0-0
knservis: @chuwy Thank you for you explanation yesterday, it helped a lot. Given my new understanding I would say that we should drop this field altogether. Agree?

chuwy: Hey @knservis do you mean `attachment-count`? If understand its purpose right - it has an analytical value, but obviously just not as a string, but as an integer.

knservis: @chuwy This is the number of attachments on the message that was sent to the user. Methinks it is probably not useful for the same reason body-plain is not useful (i.e. you know what you sent out). I suppose it doesn't matter too much as it doesn't take up much space.

schemas/com.mailgun/message_clicked/jsonschema/1-0-0
knservis: @chuwy I have changed all those to email

schemas/com.mailgun/message_bounced/jsonschema/1-0-0
chuwy: Yep, I agree even if it's not useful - it's not a big deal as well (unlike `body-plain`). Let's leave.

schemas/com.olark/transcript/jsonschema/1-0-0
knservis: @Chuwy I changed the format. About the field length I realise now I read the wrong RFC (RFC 3696), but I wonder if in the case of the ESP error in mailgun it can actually be 320?

alexanderdean:
Pending @chuwy review.

schemas/com.olark/offline_message/jsonschema/1-0-0
```diff
@@ -0,0 +1,122 @@
...
+    "items": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "kind": {
+            "type": "string",
+            "maxLength": 64
+          },
+          "timestamp": {
```
alexanderdean: Do we know what format this timestamp is in?

schemas/com.olark/transcript/jsonschema/1-0-0
```diff
@@ -0,0 +1,210 @@
...
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "body": {
+            "type": "string"
+          },
+          "timestamp": {
+            "type": "string",
+            "maxLength": 64
```
alexanderdean: Do we know what format this timestamp is in? It has a maxLength.

knservis: @alexanderdean It appears to be in float unfortunately e.g. "timestamp":"1307116657.1" or "timestamp":"1307116661.25"

alexanderdean: Can we translate to ISO8601?

knservis: @alexanderdean That is already done (https://github.com/snowplow/iglu-central/pull/664/commits/7d0a8972e654644d8c3bf6731f1dcaf51ae3e5f1#diff-29edb5b9239c6b0b35b397d86d5d0cd1R27)

schemas/com.olark/offline_message/jsonschema/1-0-0
knservis: @alexanderdean That is also already done (but for some reason not marked as "outdated" in the PR

alexanderdean: Thanks

knservis:
@chuwy ok NP

alexanderdean:
Thanks @knservis. The main thing that I just noticed and I am unsure about are the Mailgun properties, like `X-Mailgun-Sid`. Obviously this is a legal JSON property name, but it's going to tough to work with in any other environment (Redshift, Avro etc).

Question to @chuwy @BenFradet - do we "have form" (prior art) for cleaning up property names in the adapters, or are we fairly strict about passing them through unchanged?

BenFradet:
The [MP adapter](https://github.com/snowplow/snowplow/pull/3250/files) is all about renaming

knservis:
@alexanderdean  I am dropping X-Mailgun-Sid and Message-Id from all mailgun schemas as they are not documented here and I not useful (https://documentation.mailgun.com/en/latest/api-webhooks.html). They do appear in a postbin when you test the hook and that is the reason the original committer included them.

schemas/com.olark/transcript/jsonschema/1-0-0
alexanderdean: Nudge @knservis

alexanderdean:
Hey @knservis - this is looking good:

* Please rebase now that R81 is out
* You missed an email: format from @chuwy 
* Shall we just standardize all the email fields at 512 so that they can all be pseudonymized in due course?
* Are there any other fields which might constitute PII/sensitive data?

knservis:
@alexanderdean 
1. Done
2. In fact after talking to @chuwy it looks like all these format checkers will send all the data to the "bad" bucket once PII is on. The way I see it we can do one of three things:
2.a. Remove all format checking for all PII fields. That may be a bot problematic as the user can specify any field. 
2.b. Remove the PII field all together when PII enrichment runs and replace it with a new field called "${fieldName}" +"PII". This still means that we have to revisit all the schemas to ensure that they accept additional fields, but I am not sure that they will end up in RS. 
2.c. Ensure that everything that does validation (enrichment and shredding) knows about which fields have been configured as PII. This adds a dependency on PII enrichment (the config part) to shredder, but it may be the best solution. I would go with that. This suggests an new ticket as par of GDPR-1 to include support for PII fields in validation. 
3. sha-256 in hex is 64 characters long and sha-512 is 128 characters long. I think we should standardise at 256 to support both pseudonymized and plain emails. (Of course the problem remains that the user can psudonymize any field using any strategy..)
4. ^^

alexanderdean:
All approved. @knservis please bump the release date to today. Exciting!

knservis:
@alexanderdean Done. We can have the PII field discussion here https://github.com/snowplow/snowplow/issues/3528

PR #688, opened 2017-12-29
Add Twilio SMS queued schema

miike:
#685

PR #691, opened 2018-01-04
Add Twilio sent, delivered and queued SMS events

miike:
Replaces #688 #689 #690


snowplowcla:
@miike has signed the Software Grant and Corporate Contributor License Agreement. Thanks so much

alexanderdean:
Thanks @miike! I think we have a Twilio account... Do you have some configuration instructions so we can QA these schemas?

miike:
Yes - will add this in tomorrow. What's the best way to share documentation relating to webhooks like this? Should there be a high level `docs` directory with Markdown READMEs?

alexanderdean:
Great stuff - ultimately the webhook setup guide will just go into the wiki like so:

https://github.com/snowplow/snowplow/wiki/Setting-up-a-Webhook

It's probably easiest just to add the setup guide to this thread first and we can port over later...

miike:
I've just had another look at the messaging service webhooks and it looks like I'll need to write a minimal adapter for this so that events don't all use the same schema.

PR #697, opened 2018-01-11
Add com.snowplowanalytics.snowplow.enrichments/pii_enrichment_config/jsonschema/1-0-0

knservis:


schemas/com.snowplowanalytics.snowplow.enrichments/pii_pseudonymizer_config/1-0-0
```diff
@@ -0,0 +1,109 @@
...
+                "type": "object",
+                "properties": {
+                  "hashFunction": {
+                    "enum": [
+                      "MD2",
+                      "MD5",
+                      "SHA-1",
+                      "SHA-256",
+                      "SHA-384",
+                      "SHA-512"
```
alexanderdean: Are we supporting all of these hash functions for release?

alexanderdean:
It's looking pretty good @knservis ! I'm adding @chuwy for a second opinion as he's written a lot of enrichments and their configs.

schemas/com.snowplowanalytics.snowplow.enrichments/pii_pseudonymizer_config/1-0-0
knservis: We can restrict them, but by default we support all implemented MessageDigest algorithms.

schemas/com.snowplowanalytics.snowplow.enrichments/pii_pseudonymizer_config/1-0-0
```diff
@@ -0,0 +1,109 @@
...
+                  "json"
+                ]
+              }
+            ],
+            "additionalProperties": false
+          }
+        },
+        "strategy": {
+          "type": "object",
+          "properties": {
```
chuwy: Aren't keys inside `strategy` required? If not - it's better to make then `type: [object, null]`. To make their absence explicit.

schemas/com.snowplowanalytics.snowplow.enrichments/pii_pseudonymizer_config/1-0-0
```diff
@@ -0,0 +1,109 @@
...
+            "type": "object",
+            "properties": {
+              "pseudonymize": {
+                "type": "object",
+                "properties": {
+                  "hashFunction": {
+                    "enum": [
```
chuwy: Even for `enum` it's better to specify `type: string`.

schemas/com.snowplowanalytics.snowplow.enrichments/pii_pseudonymizer_config/1-0-0
```diff
@@ -0,0 +1,109 @@
+{
```
chuwy: Path is wrong. It's missing `jsonschema` in `pii_pseudonymizer_config/1-0-0`.

schemas/com.snowplowanalytics.snowplow.enrichments/pii_pseudonymizer_config/1-0-0
```diff
@@ -0,0 +1,109 @@
...
+          }
+        },
+        "required": [
+          "pii",
+          "strategy"
+        ],
+        "additionalProperties": false
+      }
+    },
+    "required": [
```
chuwy: These `required` and `additionalParameters` are misplaced, which makes its invalid schema. They shouldn't be on same level with datum properties.

knservis: Good point. Is there a jsonschema lint?

schemas/com.snowplowanalytics.snowplow.enrichments/pii_pseudonymizer_config/1-0-0
knservis: interesting. Why is that?

schemas/com.snowplowanalytics.snowplow.enrichments/pii_pseudonymizer_config/1-0-0
knservis: Just a bug. I have moved them.

schemas/com.snowplowanalytics.snowplow.enrichments/pii_pseudonymizer_config/1-0-0
chuwy: Because it is an additional guarantee that it won't be anything apart from string.

schemas/com.snowplowanalytics.snowplow.enrichments/pii_pseudonymizer_config/1-0-0
knservis: Oh I see. Not on that line. In the actual path. Ok I see it.

schemas/com.snowplowanalytics.snowplow.enrichments/pii_pseudonymizer_config/1-0-0
chuwy: ...if you'll want to transform it into other format (scala case class) that does not "understand" enums, chances are that `type: string` will help here.

knservis:
@chuwy The build succeeded here: https://travis-ci.org/snowplow/iglu-central/builds/328030931 Maybe we could extend igluctl to check the path too, or do we expect other stuff in the tree?

chuwy:
Hm, good idea. Igluctl actually checks paths, but only to check that `SchemaMap` (vendor+name+format+version) conforms path. I'll create a ticket to explore.

alexanderdean:
Hey @knservis - please update the spec in the snowplow/snowplow issue to mention all the supported hashing algo's. Then this schema LGTM.

alexanderdean:
I will get this into the upcoming Iglu Central release...

schemas/com.snowplowanalytics.snowplow.enrichments/pii_pseudonymizer_config/jsonschema/1-0-0
```diff
@@ -0,0 +1,114 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for PII pseudonymization enrichment",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow.enrichments",
+    "name": "pii_pseudonymizer_config",
```
alexanderdean: Can we change the name (and thus filenames and table name etc) to `pii_enrichment_config`. pseudonymizer is just one behavior of the enrichment.

knservis:
@alexanderdean added enum and rename

alexanderdean:
Thanks! The former is correct (not all the whitelisted fields actually needed widening).

alexanderdean:
Cherry-picked into https://github.com/snowplow/iglu-central/pull/596, closing

PR #701, opened 2018-01-15
Add com.snowplowanalytics.snowplow/pii_transformation/jsonschema/1-0-0

knservis:
- [ ] test this schema when this is ready https://github.com/snowplow/snowplow/issues/3580

@chuwy I hate the repetition in this schema, but you have told me that we cannot add references, is that right? Is there something you can suggest to avoid it?

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/pii_transformation/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow/pii_transformation/jsonschema/1-0-0
@@ -1,0 +1,162 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for a PII enrichment original values",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow",
+    "name": "pii_transformation",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "pii": {
+      "type": "object",
+      "properties": {
+        "pojo": {
+          "description": "A scalar field replacement from the EnrichedEvent POJO (e.g. user_id)",
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
+              "fieldName": {
+                "enum": [
+                  "user_id",
+                  "user_ipaddress",
+                  "user_fingerprint",
+                  "domain_userid",
+                  "network_userid",
+                  "ip_organization",
+                  "ip_domain",
+                  "tr_orderid",
+                  "ti_orderid",
+                  "mkt_term",
+                  "mkt_content",
+                  "se_category",
+                  "se_action",
+                  "se_label",
+                  "se_property",
+                  "mkt_clickid",
+                  "refr_domain_userid",
+                  "domain_sessionid"
+                ],
+                "description": "The name of the field"
+              },
+              "originalValue": {
+                "type": "string",
+                "description": "The original value (before the PII transform), of the field"
+              },
+              "modifiedValue": {
+                "type": "string",
+                "maxLength": 128,
+                "description": "The value that this field was replaced with (e.g. the hash of the original)"
+              }
+            },
+            "additionalProperties": false,
+            "required": [
+              "fieldName",
+              "originalValue",
+              "modifiedValue"
+            ]
+          }
+        },
+        "json": {
+          "description": "A JSON field replacement from the EnrichedEvent POJO (e.g. contexts). Each entry in this array is a match of the jsonPath for that field",
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
+              "fieldMapping": {
+                "description": "A single replacement (there can be multiple) caused by specifying that JsonPath",
+                "type": "object",
+                "properties": {
+                  "fieldName": {
+                    "enum": [
+                      "contexts",
+                      "derived_contexts",
+                      "unstruct_event"
+                    ],
+                    "description": "The name of the field"
+                  },
+                  "jsonPath": {
+                    "type": "string",
+                    "description": "The JsonPath that was specified"
+                  },
+                  "originalValue": {
+                    "type": "string",
+                    "description": "The original value of the field"
+                  },
+                  "modifiedValue": {
+                    "type": "string",
+                    "maxLength": 128,
+                    "description": "The modified value (e.g. hash) of the original value"
+                  },
+                  "schema": {
+                    "type": "string",
+                    "description": "The iglu schema corresponding to this field and value",
+                    "pattern": "^iglu:([a-zA-Z0-9-_.]+)/([a-zA-Z0-9-_]+)/([a-zA-Z0-9-_]+)/([1-9][0-9]*|\\*)-((?:0|[1-9][0-9]*)|\\*)-((?:0|[1-9][0-9]*)|\\*)$"
+                  }
+                },
+                "required": [
+                  "fieldName",
+                  "jsonPath",
+                  "originalValue",
+                  "modifiedValue"
+                ],
+                "additionalProperties": false
+              }
+            }
+          }
+        }
+      },
+      "anyOf": [
+        {
+          "required": [
+            "pojo"
+          ]
+        },
+        {
+          "required": [
+            "json"
+          ]
+        }
+      ],
+      "additionalProperties": false
+    },
+    "strategy": {
+      "description": "The strategy used to modify the original value",
+      "type": "object",
+      "properties": {
+        "pseudonymize": {
+          "description": "The pseudonymize strategy for modifying the original value",
+          "type": "object",
+          "properties": {
+            "hashFunction": {
+              "enum": [
+                "MD2",
+                "MD5",
+                "SHA-1",
+                "SHA-256",
+                "SHA-384",
+                "SHA-512"
+              ],
+              "description": "The function used to pseudonymize"
+            }
+          },
+          "required": [
+            "hashFunction"
+          ],
+          "additionalProperties": false
+        }
+      },
+      "required": [
+        "pseudonymize"
+      ],
+      "additionalProperties": false
+    }
+  },
+  "required": [
+    "pii",
+    "strategy"
+  ],
+  "additionalProperties": false
+}


schemas/com.snowplowanalytics.snowplow/pii_enrichment/jsonschema/1-0-0
```diff
@@ -0,0 +1,654 @@
...
+  "properties": {
+    "pii": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "pojo": {
+            "type": "object",
+            "properties": {
+              "user_id": {
```
chuwy: Am I right that we're using snake_case here to match `EnrichedEvent`'s original format? If this is the only reason than I'm not sure there's a value in that:

1. We recently decided to standardize on camelCase
2. We often transform key's cases between their original format and schema-based
3. I believe `EnrichedEvent` should and will use camelCase someday

However, even with that - this is not a strong opinion, I generally have nothing against snake_case.

schemas/com.snowplowanalytics.snowplow/pii_enrichment/jsonschema/1-0-0
```diff
@@ -0,0 +1,654 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for a PII enrichment original values",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow",
+    "name": "pii_enrichment",
```
chuwy: Is there any better names for the schema? I wouldn't understand its purpose without `description` above.

schemas/com.snowplowanalytics.snowplow/pii_enrichment/jsonschema/1-0-0
```diff
@@ -0,0 +1,654 @@
...
+                }
+              },
+              "user_fingerprint": {
+                "type": "object",
+                "properties": {
+                  "originalValue": {
+                    "type": "string"
+                  },
+                  "modifiedValue": {
+                    "type": "string"
```
chuwy: I believe all `modifiedValue` should have `maxLength: 128` (SHA-512's length - longest hash of these if I'm not mistaken)

schemas/com.snowplowanalytics.snowplow/pii_enrichment/jsonschema/1-0-0
```diff
@@ -0,0 +1,654 @@
...
+                      }
+                    }
+                  }
+                }
+              },
+              "user_fingerprint": {
+                "type": "object",
+                "properties": {
+                  "originalValue": {
+                    "type": "string"
```
chuwy: `originalValue`s can have `maxLength`s based on their constraints from `atomic-def`.

alexanderdean:
I don't think this is quite the right design:

1. The strategy for the PII treatment is set at a global level, so why does this event store it against every field? That's a guess as to where the enrichment might go next
2. Treating each POJO field as a sub-object is needlessly verbose, it's brittle (because we will have to change this event if we add another supported field), and it will generate *more* work in the downstream service because we will have to look for each option individually. Much better to just have an array of `pojoField[]`

knservis:
@alexanderdean 
1. Currently the code supports a strategy per field (the configuration doesn't). That will make it forward compatible but we may not want to include that now. I am just worried about the versioning (I would like to be able to express a dependency on on a specific configuration schema version). Is that possible?
2. Agreed. I don't much like it either. If we push the field name into a generic pojoField property, do we constrain it (i.e. use an enum)?

schemas/com.snowplowanalytics.snowplow/pii_enrichment/jsonschema/1-0-0
knservis: We are going with generic pojoField.

schemas/com.snowplowanalytics.snowplow/pii_enrichment/jsonschema/1-0-0
knservis: What do you suggest?

schemas/com.snowplowanalytics.snowplow/pii_enrichment/jsonschema/1-0-0
knservis: Moving to generic pojoField

schemas/com.snowplowanalytics.snowplow/pii_enrichment/jsonschema/1-0-0
chuwy: `pii_original` maybe? Or `pii_values`

knservis: pii_original_values sounds better to me but I think @alexanderdean was thinking that in the future this may be used for more things (e.g. unique customer identifier).

alexanderdean:
> Currently the code supports a strategy per field (the configuration doesn't). That will make it forward compatible but we may not want to include that now. I am just worried about the versioning (I would like to be able to express a dependency on on a specific configuration schema version). Is that possible?

Basically - as and when we need to support per-field strategies, we can do a 2-0-0 of the schema, and that can express that at that time. But keen we don't try and solve for problems we don't have/understand yet.

> If we push the field name into a generic pojoField property, do we constrain it (i.e. use an enum)?

Sure, let's constrain it for now.

schemas/com.snowplowanalytics.snowplow/pii_enrichment/jsonschema/1-0-0
alexanderdean: @knservis's idea here was `pii_identification` - what do you think @chuwy ? We are looking for something that captures the action that led to the event, the moment in time - not just the description of the bundle of data we are sending (especially as that bundle could change). Other options `pii_transformation`, `pii_...`?

chuwy: I think `pii_transformation` is definitely step forward. `pii_capture`? `pii_snapshot` is also about description of data, but sounds good.

knservis: I think pii_capture is the best of those (pii_cleartext_mapping is another idea), let's not drag this too long. @alexanderdean ?

alexanderdean: Let's go with `pii_transformation` for now.

knservis:
@alexanderdean @chuwy I have changed  pii/json[]/field to an enum and removed pii/json[]/strategy as we only specify a single strategy for the whole event. Please have a look

alexanderdean:
Have you linted this? I can see at least one bug.

knservis:
@alexanderdean I did run lint and so did the CI.

alexanderdean:
Hey @knservis - ah, it's not an error in the schema itself, but it will break when you try to validate a schema against it. It's `pojo` versus `pojoField`...

While I am here - why is `strategy` *inside* the items within the array of `pii`? As discussed, there is only one strategy for all PII in the event?

knservis:
Hi @alexanderdean thank you for that. I fixed both issues.

knservis:
@chuwy I did not think about it. Good idea. Actually @alexanderdean we may want to add another field in pii/json[]/fieldMapping called jsonPathMatched. The JsonPath value can be something like ``` "$.email.*" ``` and the JsonPathMatched will be ```/emailAdddress1```. Does that make sense?

knservis:
@chuwy and @alexanderdean I have updated this with description and moved the function down. Please have another look.

alexanderdean:
Looks good to me! I really like us adding the description columns. It's us raising the standard of our schemas.

knservis:
@alexanderdean Maybe we should not merge this until it can be tested. I will add a todo.

knservis:
@alexanderdean @chuwy I don't think it makes sense to have pii be an array since both pojo and json are arrays. pii should be an object.

alexanderdean:
I agree the current structure is too complicated. Why don't you paste in to this thread an example as-is *instance*, and then show the proposed to-be *instance*. Then we can iterate at the level of instances rather than ping-ponging schema changes.

knservis:
Here is the proposed schema:
```
{
  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
  "description": "Schema for a PII enrichment original values",
  "self": {
    "vendor": "com.snowplowanalytics.snowplow",
    "name": "pii_transformation",
    "format": "jsonschema",
    "version": "1-0-0"
  },
  "type": "object",
  "properties": {
    "pii": {
      "type": "object",
      "properties": {
        "pojo": {
          "description": "A scalar field replacement from the EnrichedEvent POJO (e.g. user_id)",
          "type": "array",
          "items": {
            "type": "object",
            "properties": {
              "fieldName": {
                "enum": [
                  "user_id",
                  "user_ipaddress",
                  "user_fingerprint",
                  "domain_userid",
                  "network_userid",
                  "ip_organization",
                  "ip_domain",
                  "tr_orderid",
                  "ti_orderid",
                  "mkt_term",
                  "mkt_content",
                  "se_category",
                  "se_action",
                  "se_label",
                  "se_property",
                  "mkt_clickid",
                  "refr_domain_userid",
                  "domain_sessionid"
                ],
                "description": "The name of the field"
              },
              "originalValue": {
                "type": "string",
                "description": "The original value (before the PII transoform), of the field"
              },
              "modifiedValue": {
                "type": "string",
                "maxLength": 128,
                "description": "The value that this field was replaced with (e.g. the hash o fteh original)"
              }
            },
            "additionalProperties": false,
            "required": [
              "fieldName",
              "originalValue",
              "modifiedValue"
            ]
          }
        },
        "json": {
          "description": "A JSON field replacement from the EnrichedEvent POJO (e.g. contexts). Each entry in this array is a match of the jsonPath for that field",
          "type": "array",
          "items": {
            "type": "object",
            "properties": {
              "fieldMapping": {
                "description": "A single replacement (there can be multiple) caused by specifying that JsonPath",
                "type": "object",
                "properties": {
                  "field": {
                    "enum": [
                      "contexts",
                      "derived_contexts",
                      "unstruct_event"
                    ],
                    "description": "The name fo the field"
                  },
                  "jsonPath": {
                    "type": "string",
                    "description": "The JsonPath that was specified"
                  },
                  "jsonPathMatched": {
                    "type": "string",
                    "description": "The JsonPath that was matched for this value"
                  },
                  "originalValue": {
                    "type": "string",
                    "description": "The original value of the field"
                  },
                  "modifiedValue": {
                    "type": "string",
                    "maxLength": 128,
                    "description": "The modified value (e.g. hash of the original) value"
                  }
                },
                "required": [
                  "field",
                  "jsonPath",
                  "originalValue",
                  "modifiedValue"
                ],
                "additionalProperties": false
              }
            }
          }
        }
      },
      "anyOf": [
        {
          "required": [
            "pojo"
          ]
        },
        {
          "required": [
            "json"
          ]
        }
      ],
      "additionalProperties": false
    },
    "strategy": {
      "description": "The strategy used to modify the original value",
      "type": "object",
      "properties": {
        "pseudonymize": {
          "description": "The pseudonymize strategy for modifying the original value",
          "type": "object",
          "properties": {
            "hashFunction": {
              "enum": [
                "MD2",
                "MD5",
                "SHA-1",
                "SHA-256",
                "SHA-384",
                "SHA-512"
              ],
              "description": "The function used to pseudonymize"
            }
          },
          "required": [
            "hashFunction"
          ],
          "additionalProperties": false
        }
      },
      "required": [
        "pseudonymize"
      ],
      "additionalProperties": false
    }
  },
  "required": [
    "pii",
    "strategy"
  ],
  "additionalProperties": false
}
```

this is an instance of the event under that schema:

```
{
  "pii": {
    "pojo": [
      {
        "fieldName": "user_id",
        "originalValue": "john@acme.com",
        "modifiedValue": "4b2d8785b49bad23638b17d8db76857a79bf79441241a78a97d88cc64bbf766e"
      },
      {
        "fieldName": "user_ipaddress",
        "originalValue": "70.46.123.145",
        "modifiedValue": "36595ea260a82b7e2d7cf44121892bf31031a9c27077d8c802454464178456c2"
      },
      {
        "fieldName": "user_fingerprint",
        "originalValue": "its_you_again!",
        "modifiedValue": "9f9fc89b7a5428f2646347974404650fc8776f791afc2200efc8a82aa754e7e6"
      }
    ],
    "json": [
      {
        "field": "contexts",
        "originalValue": "jim@acme.com",
        "modifiedValue": "3571b422ecb9ac85cb654b2fce521ae351d4695b0fb788aac75caf724e7881f0",
        "jsonPath": "$.emailAddress"
      },
      {
        "field": "contexts",
        "originalValue": "bob@acme.com",
        "modifiedValue": "405ac8384fa984f787f9486daf34d84d98f20c4d6a12e2cc4ed89be3bcb06ad6",
        "jsonPath": "$.data.emailAddress2"
      },
      {
        "field": "unstruct_event",
        "originalValue": "50.56.129.169",
        "modifiedValue": "b5814ada7bb3abb2ed7f8713433a60ed3b3780f7d98a95c936cc62abb16f316f",
        "jsonPath": "$.ip"
      }
    ]
  },
  "strategy": {
    "pseudonymize": {
      "hashFunction": "SHA-256"
    }
  }
}
```

alexanderdean:
I thought you couldn't get `jsonPath` working? Have I misunderstood?

knservis:
Currently the schema is:
```
{
  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
  "description": "Schema for a PII enrichment original values",
  "self": {
    "vendor": "com.snowplowanalytics.snowplow",
    "name": "pii_transformation",
    "format": "jsonschema",
    "version": "1-0-0"
  },
  "type": "object",
  "properties": {
    "pii": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "pojo": {
            "description": "A scalar field replacement from the EnrichedEvent POJO (e.g. user_id)",
            "type": "array",
            "items": {
              "type": "object",
              "properties": {
                "fieldName": {
                  "enum": [
                    "user_id",
                    "user_ipaddress",
                    "user_fingerprint",
                    "domain_userid",
                    "network_userid",
                    "ip_organization",
                    "ip_domain",
                    "tr_orderid",
                    "ti_orderid",
                    "mkt_term",
                    "mkt_content",
                    "se_category",
                    "se_action",
                    "se_label",
                    "se_property",
                    "mkt_clickid",
                    "refr_domain_userid",
                    "domain_sessionid"
                  ],
                  "description": "The name of the field"
                },
                "originalValue": {
                  "type": "string",
                  "description": "The original value (before the PII transoform), of the field"
                },
                "modifiedValue": {
                  "type": "string",
                  "maxLength": 128,
                  "description": "The value that this field was replaced with (e.g. the hash o fteh original)"
                }
              },
              "additionalProperties": false,
              "required": [
                "fieldName",
                "originalValue",
                "modifiedValue"
              ]
            }
          },
          "json": {
            "description": "A JSON field replacement from the EnrichedEvent POJO (e.g. contexts). Each entry in this array is a match of the jsonPath for that field",
            "type": "array",
            "items": {
              "type": "object",
              "properties": {
                "fieldMapping": {
                  "description": "A single replacement (there can be multiple) caused by specifying that JsonPath",
                  "type": "object",
                  "properties": {
                    "field": {
                      "enum": [
                        "contexts",
                        "derived_contexts",
                        "unstruct_event"
                      ],
                      "description": "The name fo the field"
                    },
                    "jsonPath": {
                      "type": "string",
                      "description": "The JsonPath that was specified"
                    },
                    "jsonPathMatched": {
                      "type": "string",
                      "description": "The JsonPath that was matched for this value"
                    },
                    "originalValue": {
                      "type": "string",
                      "description": "The original value of the field"
                    },
                    "modifiedValue": {
                      "type": "string",
                      "maxLength": 128,
                      "description": "The modified value (e.g. hash of the original) value"
                    }
                  },
                  "required": [
                    "field",
                    "jsonPath",
                    "originalValue",
                    "modifiedValue"
                  ],
                  "additionalProperties": false
                }
              }
            }
          }
        },
        "oneOf": [
          {
            "required": [
              "pojo"
            ]
          },
          {
            "required": [
              "json"
            ]
          }
        ],
        "additionalProperties": false
      }
    },
    "strategy": {
      "description": "The strategy used to modify the original value",
      "type": "object",
      "properties": {
        "pseudonymize": {
          "description": "The pseudonymize strategy for modifying the original value",
          "type": "object",
          "properties": {
            "function": {
              "enum": [
                "MD2",
                "MD5",
                "SHA-1",
                "SHA-256",
                "SHA-384",
                "SHA-512"
              ],
              "description": "The function used to pseudonymize"
            }
          },
          "required": ["function"],
          "additionalProperties": false
        }
      },
      "required": ["pseudonymize"],
      "additionalProperties": false
    }
  },
  "required": [
    "pii",
    "strategy"
  ],
  "additionalProperties": false
}
```
An instance of which is: 
```
{
  "pii": [
    {
      "pojo": [
        {
          "fieldName": "user_id",
          "originalValue": "john@acme.com",
          "modifiedValue": "4b2d8785b49bad23638b17d8db76857a79bf79441241a78a97d88cc64bbf766e"
        },
        {
          "fieldName": "user_ipaddress",
          "originalValue": "70.46.123.145",
          "modifiedValue": "36595ea260a82b7e2d7cf44121892bf31031a9c27077d8c802454464178456c2"
        },
        {
          "fieldName": "user_fingerprint",
          "originalValue": "its_you_again!",
          "modifiedValue": "9f9fc89b7a5428f2646347974404650fc8776f791afc2200efc8a82aa754e7e6"
        }
      ],
      "json": [
        {
          "field": "contexts",
          "originalValue": "jim@acme.com",
          "modifiedValue": "3571b422ecb9ac85cb654b2fce521ae351d4695b0fb788aac75caf724e7881f0",
          "jsonPath": "$.emailAddress"
        },
        {
          "field": "contexts",
          "originalValue": "bob@acme.com",
          "modifiedValue": "405ac8384fa984f787f9486daf34d84d98f20c4d6a12e2cc4ed89be3bcb06ad6",
          "jsonPath": "$.data.emailAddress2"
        },
        {
          "field": "unstruct_event",
          "originalValue": "50.56.129.169",
          "modifiedValue": "b5814ada7bb3abb2ed7f8713433a60ed3b3780f7d98a95c936cc62abb16f316f",
          "jsonPath": "$.ip"
        }
      ]
    }
  ],
  "strategy": {
    "pseudonymize": {
      "function": "SHA-256"
    }
  }
}
```
which to me makes no sense

knservis:
@alexanderdean I could not get jsonPathMatched working

knservis:
Adding the changes to compare (will rebase once done)

alexanderdean:
I think this is much saner. @chuwy ?

chuwy:
So, if I parsed it correctly, proposed change is to remove wrapping array?

```json
{
  "pii": {
    "pojo": [{}, {}],
    "json": [{}, {}]
  }
}
```

instead of previous:

```json
{
  "pii": [
    {
      "pojo": [{}, {}],
      "json": [{}, {}]
    }
  ]
}
```

Yes, definitely much better.

After getting a better picture of how resulting JSON looks like, I got a better idea how we can structure it, but I'll probably submit an issue instead of another refactoring.

alexanderdean:
Not sure I understood - what will be in the ticket @chuwy ? Why not just comment here, given this schema is still WIP?

chuwy:
Idea is to follow somewhat similar to Analytics SDK. We can reduce amount of arrays further. Something like:

```json
{
  "pii": {
    "pojo": {
      "user_id": "4b2d8785b49bad23638b17d8db76857a79bf79441241a78a97d88cc64bbf766e",
      "originalValue": "john@acme.com"
    },
    "json": {
      "com_amce_event_1": {
        "values": [
          {
            "jsonPath": "$.emailAddress",
            "originalValue": "jim@acme.com",
            "modifiedValue": "3571b422ecb9ac85cb654b2fce521ae351d4695b0fb788aac75caf724e7881f0",
          }
        ]
      }
    }
  },
  "strategy": {
    "pseudonymize": {
      "function": "SHA-256"
    }
  }
}
```

JSON schema URIs can be full instead of "normalized" of course. I just like the fact that presence of anonymized fields can be easily queried by key.

alexanderdean:
Ah - no we discounted that approach already, @chuwy. The tl;dr is that the access pattern (think downstream consumer of the event) is much more convenient with arrays and the event is less fragile with arrays too.

alexanderdean:
Hey @knservis - is this a complete entry for JSON:

```
        {
          "field": "contexts",
          "originalValue": "jim@acme.com",
          "modifiedValue": "3571b422ecb9ac85cb654b2fce521ae351d4695b0fb788aac75caf724e7881f0",
          "jsonPath": "$.emailAddress"
        },
```

it's missing the actual JSON Schema it was pulled from, which is crucial information...

alexanderdean:
Adding @BenFradet to review as this is going into Snowplow Pipeline.

schemas/com.snowplowanalytics.snowplow/pii_transformation/jsonschema/1-0-0
```diff
@@ -0,0 +1,162 @@
...
+                "description": "The name of the field"
+              },
+              "originalValue": {
+                "type": "string",
+                "description": "The original value (before the PII transoform), of the field"
+              },
+              "modifiedValue": {
+                "type": "string",
+                "maxLength": 128,
+                "description": "The value that this field was replaced with (e.g. the hash o fteh original)"
```
BenFradet: typo

schemas/com.snowplowanalytics.snowplow/pii_transformation/jsonschema/1-0-0
```diff
@@ -0,0 +1,162 @@
...
+                "description": "A single replacement (there can be multiple) caused by specifying that JsonPath",
+                "type": "object",
+                "properties": {
+                  "fieldName": {
+                    "enum": [
+                      "contexts",
+                      "derived_contexts",
+                      "unstruct_event"
+                    ],
+                    "description": "The name fo the field"
```
BenFradet: typo

schemas/com.snowplowanalytics.snowplow/pii_transformation/jsonschema/1-0-0
```diff
@@ -0,0 +1,162 @@
...
+                    "description": "The JsonPath that was specified"
+                  },
+                  "originalValue": {
+                    "type": "string",
+                    "description": "The original value of the field"
+                  },
+                  "modifiedValue": {
+                    "type": "string",
+                    "maxLength": 128,
+                    "description": "The modified value (e.g. hash of the original) value"
```
BenFradet: typo

schemas/com.snowplowanalytics.snowplow/pii_transformation/jsonschema/1-0-0
```diff
@@ -0,0 +1,162 @@
...
+                  },
+                  "schema": {
+                    "type": "string",
+                    "description": "The iglu schema corresponding to this field and value",
+                    "pattern": "^iglu:([a-zA-Z0-9-_.]+)/([a-zA-Z0-9-_]+)/([a-zA-Z0-9-_]+)/([1-9][0-9]*|\\*)-((?:0|[1-9][0-9]*)|\\*)-((?:0|[1-9][0-9]*)|\\*)$"
+                  }
+                },
+                "required": [
+                  "field",
```
BenFradet: `fieldName`

schemas/com.snowplowanalytics.snowplow/pii_transformation/jsonschema/1-0-0
knservis: Where?

BenFradet: parens should be at the end, no?

alexanderdean:
This is waiting on @BenFradet to approve...

PR #708, opened 2018-01-18
Add com.snowplowanalytics.snowplow.enrichments/pii_enrichment_config/jsonschema/2-0-0

knservis:
- [ ] Depends on testing with https://github.com/snowplow/snowplow/issues/3580

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow.enrichments/pii_enrichment_config/jsonschema/2-0-0
+++ b/schemas/com.snowplowanalytics.snowplow.enrichments/pii_enrichment_config/jsonschema/2-0-0
@@ -1,0 +1,164 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for PII pseudonymization enrichment",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow.enrichments",
+    "name": "pii_enrichment_config",
+    "format": "jsonschema",
+    "version": "2-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "vendor": {
+      "type": "string",
+      "description": "The name of the vendor for this config (the only valid value for scala-common enrich is com.snowplowanalytics.snowplow.enrichments)"
+    },
+    "name": {
+      "type": "string",
+      "description": "The name of the config (the only valid value for scala-common enrich is pii_enrichment_config)"
+    },
+    "enabled": {
+      "type": "boolean",
+      "description": "Whether to enable this enrichment"
+    },
+    "emitEvent": {
+      "type": "boolean",
+      "description": "Whether to emit identification events from this enrichment"
+    },
+    "parameters": {
+      "type": "object",
+      "properties": {
+        "pii": {
+          "description": "List of all the fields for which pdeudonymization will be performed",
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
+              "pojo": {
+                "description": "Scalar field which contains a single string value, on which pseudonymization will be performed in the enire field (e.g. `user-123`)",
+                "type": "object",
+                "properties": {
+                  "field": {
+                    "enum": [
+                      "user_id",
+                      "user_ipaddress",
+                      "user_fingerprint",
+                      "domain_userid",
+                      "network_userid",
+                      "ip_organization",
+                      "ip_domain",
+                      "tr_orderid",
+                      "ti_orderid",
+                      "mkt_term",
+                      "mkt_content",
+                      "se_category",
+                      "se_action",
+                      "se_label",
+                      "se_property",
+                      "mkt_clickid",
+                      "refr_domain_userid",
+                      "domain_sessionid"
+                    ]
+                  }
+                },
+                "required": [
+                  "field"
+                ],
+                "additionalProperties": false
+              },
+              "json": {
+                "description": "JSON field which contains a JSON string value, on which pseudonymization will be performed on a specific JSON path",
+                "type": "object",
+                "properties": {
+                  "field": {
+                    "enum": [
+                      "contexts",
+                      "derived_contexts",
+                      "unstruct_event"
+                    ]
+                  },
+                  "schemaCriterion": {
+                    "type": "string",
+                    "pattern": "^iglu:([a-zA-Z0-9-_.]+)/([a-zA-Z0-9-_]+)/([a-zA-Z0-9-_]+)/([1-9][0-9]*|\\*)-((?:0|[1-9][0-9]*)|\\*)-((?:0|[1-9][0-9]*)|\\*)$"
+                  },
+                  "jsonPath": {
+                    "type": "string",
+                    "pattern": "^\\$.*$"
+                  }
+                },
+                "required": [
+                  "field",
+                  "schemaCriterion",
+                  "jsonPath"
+                ],
+                "additionalProperties": false
+              }
+            },
+            "oneOf": [
+              {
+                "required": [
+                  "pojo"
+                ]
+              },
+              {
+                "required": [
+                  "json"
+                ]
+              }
+            ],
+            "additionalProperties": false
+          }
+        },
+        "strategy": {
+          "description": "The pseudonymization strategy which will be applied to all the fields specified in the `pii` section",
+          "type": "object",
+          "properties": {
+            "pseudonymize": {
+              "description": "Pseudonymization strategy that hashes using a specified algorithm",
+              "type": "object",
+              "properties": {
+                "hashFunction": {
+                  "description": "The hash function that will be used by this strategy",
+                  "enum": [
+                    "MD2",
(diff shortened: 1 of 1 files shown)


schemas/com.snowplowanalytics.snowplow.enrichments/pii_enrichment_config/jsonschema/1-1-0
```diff
@@ -0,0 +1,137 @@
...
+    "name": "pii_enrichment_config",
+    "format": "jsonschema",
+    "version": "1-1-0"
+  },
+  "type": "object",
+  "properties": {
+    "enabled": {
+      "type": "boolean"
+    },
+    "emitIdentificationEvent": {
```
alexanderdean: Shouldn't this be `emitPiiTransformationEvent`?

alexanderdean: Or actually I prefer `emitEvent`, to make it less coupled to the event name

knservis: Done

schemas/com.snowplowanalytics.snowplow.enrichments/pii_enrichment_config/jsonschema/1-1-0
```diff
@@ -0,0 +1,137 @@
...
+      "required": [
+        "pii",
+        "strategy"
+      ],
+      "additionalProperties": false
+    }
+  },
+  "required": [
+    "enabled",
+    "emitIdentificationEvent",
```
alexanderdean: This is still pointing to the old property name. Did you lint this schema?

knservis: oops

schemas/com.snowplowanalytics.snowplow.enrichments/pii_enrichment_config/jsonschema/1-1-0
```diff
@@ -0,0 +1,137 @@
...
+      "required": [
+        "pii",
+        "strategy"
+      ],
+      "additionalProperties": false
+    }
+  },
+  "required": [
+    "enabled",
+    "emitEvent",
```
alexanderdean: Technically this requirement here makes the schema a 2-0-0 (breaking change), because a 1-0-0 instance will fail validation against this schema now. @chuwy do you have a preference on what we do - bump the version or make the field not-required?

knservis: This is 1-1-0 if not required, as this is not a fix right?

alexanderdean: No, the middle digit refers to a schema widening (we call it a `REVISION`), not a `MINOR` - SchemaVer is not SemVer. It's my bad for saying it was 1-1-0 - it should be either 2-0-0 or 1-0-1... Let's see what @chuwy says.

chuwy: Am I wrong in assumption that there's exactly 0 instances with this schema attached in the wild. If this assumption is correct then I prefer to patch it actually (`1-0-0` again).

But if above assumption is wrong then in general, for configurations I prefer to **not** avoid MODEL bumps - user has to change `schema` anyway, let's make it clear that there's breaking changes instead of loosening schema.

alexanderdean: > Am I wrong in assumption that there's exactly 0 instances with this schema attached in the wild. If this assumption is correct then I prefer to patch it actually (1-0-0 again).

This assumption will be wrong by the time this version of the schema is released.

> for configurations I prefer to not avoid MODEL bumps - user has to change schema anyway, let's make it clear that there's breaking changes instead of loosening schema.

Sounds good, 2-0-0 it is!

chuwy: > This assumption will be wrong by the time this version of the schema is released.

Sorry for confusion - I meant `1-0-0`. PII enrichment doesn't exist de-facto, so there's just nothing can be incompatible yet.

alexanderdean: Not sure I follow. By the time this version is released, 1-0-0 will be out in the wild.

chuwy: I think I'm missing some detail as well... anyway I don't have a strong preference, let it be `2-0-0`.

knservis:
@alexanderdean Given the discussion on the approach about changing to direct event mutation on PII, the /parameters/pii/json/field shoudl probably be an enum with a single allowed value "contexts", right?

alexanderdean:
Enum with `contexts`, `derived_contexts` and `unstruct_event`

knservis:
@alexanderdean Done

knservis:
@alexanderdean @chuwy I have added salt and some descriptions. Please have another look

alexanderdean:
Thanks @knservis. This makes sense to me but I looked at it very briefly. @chuwy?

schemas/com.snowplowanalytics.snowplow.enrichments/pii_enrichment_config/jsonschema/2-0-0
```diff
@@ -0,0 +1,195 @@
...
+                    "SHA-256",
+                    "SHA-384",
+                    "SHA-512"
+                  ]
+                },
+                "salt": {
+                  "description": "Describes the method of obtaining a salt that will be added to the field durring hashing",
+                  "type": "object",
+                  "properties": {
+                    "ec2ParameterStore": {
```
chuwy: Can we change it to exact [same format](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow.storage/redshift_config/jsonschema/2-1-0#L36-L50) as it is in storage configuration? I mean `[string, object]` insetad of `oneOf: [plainValue, ec2ParameterStore]` I probably like your approach little bit more, but just in sake of consistency.

alexanderdean: Yes that's probably a good idea.

knservis: @chuwy Can you please check the last commit, to see if that is what you meant?

chuwy: No, sorry, not exactly. `plainValue` is redundant. You just specify `"salt": "foobar"` when you need plain value, not `"salt": { "plainValue": "foobar" }

knservis: Done

knservis:
@chuwy @BenFradet Thanks for the review. Merging

alexanderdean:
Merging to what?

knservis:
@alexanderdean That comment was deleted :-) (It was meant for here https://github.com/snowplow/snowplow/pull/3633#issuecomment-372349697)

schemas/com.snowplowanalytics.snowplow.enrichments/pii_enrichment_config/jsonschema/2-0-0
knservis: Done. I have to say, I don't think this is an improvement over the original

chuwy: > I don't think this is an improvement over the original

I agree with that. But that was just in sake of consistency.

alexanderdean:
@BenFradet can you review this as it affects a Snowplow Pipeline enrichment config.

alexanderdean:
Nudge @BenFradet @chuwy

alexanderdean:
Hey @knservis - when you get the green ticks on this one, please put together a new PR with all the commits for R90. Then you can close the old PRs. Then assign to me. Thanks!

knservis:
@alexanderdean Sure. Should we not wait until we have tested with the ssnpl rc to merge? The changes are substantial and I would like to make sure we haven't missed anything.

knservis:
@chuwy I have added that. We should probably add it to all enrichments, as none of them have it.

knservis:
@alexanderdean I am waiting for the snowplow/snowplow rc to smoke test and finalise things here (to prevent the possibility that something has been missed and the need for a patch).

PR #722, opened 2018-02-07
Patch com.snowplowanalytics.snowplow.enrichments/pii_enrichment_confi…

### LINKED ISSUE #721: Patch com.snowplowanalytics.snowplow.enrichments/pii_enrichment_config/jsonschema/1-0-0
com.snowplowanalytics.snowplow.enrichments/pii_enrichment_config/jsonschema/1-0-0

added in https://github.com/snowplow/iglu-central/issues/692 requires the addition of something like:
```
		"vendor": {
			"type": "string",
                        "default": "com.snowplowanalytics.snowplow.enrichments"
		},
		"name": {
			"type": "string",
                        "default": "pii_enrichment_config"
		},
```
In the required fields to work as expected. I am not certain whether a new version or replacing the existing version is the right thing @alexanderdean 
The current version will not work at all as far as I can tell as streaming enrich  (and presumably spark enrich) require the vendor and name fields but these are not allowed by the schema.

To be honest I don't understand the need for a name and vendor field in the instance of the config, since these are already contained in the "self" section of the config schema where the instance is pointing to.

knservis:
…g/jsonschema/1-0-0 (closes #721)

this adds vendor and name to the config instances (in addition to the config schema) as that is a requirement for all enrichments on stream enrich.

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow.enrichments/pii_enrichment_config/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow.enrichments/pii_enrichment_config/jsonschema/1-0-0
@@ -9,6 +9,12 @@
   },
   "type": "object",
   "properties": {
+    "vendor": {
+      "type": "string"
+    },
+    "name": {
+      "type": "string"
+    },
     "enabled": {
       "type": "boolean"
     },
@@ -126,6 +132,8 @@
     }
   },
   "required": [
+    "vendor",
+    "name",
     "enabled",
     "parameters"
   ],


knservis:
@chuwy I tried using a config without specifying name and vendor and it worked.. No problems with changing it, but what do you think will not work correctly if we don't?

alexanderdean:
We're very behind on this release and we're trying to do too much here. Remove the defaults and make them required.

alexanderdean:
Cherry-picked into [release branch](https://github.com/snowplow/iglu-central/pull/728), closing

PR #724, opened 2018-02-07
Add com.apple.apns/notification_event/jsonschema/1-0-0

mhadam:
This schema is for tracking opened push notifications on mobile devices. I've selected the fields that I think are most important, though other data available from a push notification is detailed [here](https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/PayloadKeyReference.html).


mhadam:
The schema is complete now.
Only question: I'm not sure if the `allOf` & `oneOf` keywords part I added will do what I want. I'm looking to always require the platform to be identified, but then optionally require either an iOS or Android notification be specified.

I couldn't find anything in documentation that requires that `title` and `body` must be defined in the iOS `alert`, but it would be strange to notify without those so I required them.

For record-keeping here's the notification spec references I used:
[iOS remote notification payload](https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CreatingtheNotificationPayload.html#//apple_ref/doc/uid/TP40008194-CH10-SW1)
[iOS remote notification payload - aps keys](https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/PayloadKeyReference.html)
[Android](https://developers.google.com/cloud-messaging/concept-options#notifications_and_data_messages)

schemas/com.snowplowanalytics.snowplow/push_notification_open/jsonschema/1-0-0
```diff
@@ -0,0 +1,129 @@
...
+                    "type": "object",
+                    "properties": {},
+                    "additionalProperties": { "type": ["string", "number"] }
+                }
+            },
+            "required": ["token", "priority", "notification"],
+            "additionalProperties": false
+        }
+    },
+    "allOf": [{"required": ["platform"]},
```
chuwy: The indentation looks little bit broken. I'd make it:

```json
{
  "allOf": [
    { "required": ["platform"] },
    { "oneOf": [{ "required": ["apns"] }, { "required": ["fcm"] }] } 
  ]
}

**Comment by chuwy** [comment]
Hey @mhadam,

Yes, I think yes it is going to work, but this should also work without `anyOf` (I think). You can check all possible instances here: https://www.jsonschemavalidator.net

File: jsonpaths/com.snowplowanalytics.snowplow/push_notification_open_1.json [hunk 2]
```diff
@@ -0,0 +1,34 @@
...
+        "$.hierarchy.rootId",
+        "$.hierarchy.rootTstamp",
+        "$.hierarchy.refRoot",
+        "$.hierarchy.refTree",
+        "$.hierarchy.refParent",
+        "$.data.apns.apn.alert.body",
+        "$.data.apns.apn.alert.title",
+        "$.data.fcm.notification.body",
+        "$.data.fcm.notification.title",
+        "$.data.apns.apn.alert.action-loc-key",
```
alexanderdean: Go with camelCase, so `actionLocKey` - this applies to all the properties.

alexanderdean:
@alexanderdean to re-review

mhadam:
I don't recall if we had discussed/decided this before, but I dropped the `apns` part since I updated the schema to take any information available when a notification is received. The notification content for both remote and local notifications is the same (but local ones aren't sent from APNS).

alexanderdean:
Thanks @mhadam. Can you define the difference between a local and remote notification? I'm hazy on the difference. Also, please go ahead and get this PR down to 3 files not 5.

mhadam:
Remote notifications are sent through the cloud service APNS, the phone basically polls for new notifications from the server. Whereas local ones are created on the device itself, and they're triggered by time, date or location.

alexanderdean:
Thanks @mhadam - your updated approach makes sense to me.

@chuwy please review this and get this added into your imminent Iglu Central release.

Thanks both!

mhadam:
I made a small change: removed the eventType field, added trigger and date. Date is for storing when it gets delivered to the user (which differs from when it's opened - the timestamp of the event). Trigger describes what caused it's delivery (local trigger, or push), which effectively replaces the eventType.

alexanderdean:
Thanks @mhadam . Do you want to rename `date` to `deliveryDate` to remove ambiguity?

alexanderdean:
And do you want to make `trigger` an enum?

mhadam:
For the trigger enum, I'm thinking something along the lines of:

- timeInterval
- calendar
- location
- push

alexanderdean:
Sounds good. We normally would use TIME_INTERVAL as an enum value.

chuwy:
Closing in favor of https://github.com/snowplow/iglu-central/pull/743 (assuming greenlight was given for this).

PR #726, opened 2018-02-08
Extend push.bash to handle JSON Paths files (close #678)

### LINKED ISSUE #678: Infrastructure: extend push.bash to handle JSON Paths files
### Background

To support loading these records into Redshift, the JSON Paths files [found in this repo](https://github.com/snowplow/iglu-central/tree/master/jsonpaths) must be published (as publically accessible files) to Amazon S3.

Due to a limitation in Redshift's [`COPY FROM JSON` command](http://docs.aws.amazon.com/redshift/latest/dg/copy-usage_notes-copy-from-json.html), these JSON Paths files must be made available in an S3 bucket in the same region as the Redshift being loaded.

### Earlier solution

Historically, the JSON Paths files lived in the [snowplow/snowplow repo](), and there was a script in that repo designed to perform this sync process (but not the original deployment of the JSON Paths files). However, this script was removed when we changed how Snowplow artifacts are published to S3.

You can see the last version of the script here:

https://github.com/snowplow/snowplow/blob/3177968e579f2f7925e6a2b1c00dc2566046df1c/vagrant/push/sync-hosted-assets.bash

### Proposed solution

I propose that we extend `push.bash` to:

* Push (all) the JSON Paths files to the `s3://snowplow-hosted-assets` bucket
* Synchronize the JSON Paths files in `s3://snowplow-hosted-assets` to all the other regional buckets

oguzhanunlu:
Following checklist at #678 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
-language: java
-
-jdk:
- - oraclejdk7
-
-script:
-  ./ci/lint.bash
+script: "./ci/lint.bash"
+before_deploy: pip install --user awscli
+deploy:
+- provider: script
+  skip_cleanup: true
+  script: "./ci/push.bash"
+  on:
+    tags: true
+    condition: '"$(./ci/is_release_tag.bash $TRAVIS_TAG)" != "" && $? == 0'
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,3 +7,7 @@ deploy:
   on:
     tags: true
     condition: '"$(./ci/is_release_tag.bash $TRAVIS_TAG)" != "" && $? == 0'
+env:
+  global:
+  - secure: qcvBVPk1h8FHt7/HMgBd1JAYLjJn2055vMQI3mx3pMAf8Y70nY2kDeEyIgd7bJDg4ZpaNiCs3SDNxENLZ+RO2dRS5i1Od0P0zYLA6E9BUzaF1WkJ/HCEIg9xCblvB1TdPFz2TNmUQh6Lcbd5qBqFuWgnHT+GLJppRtwUSS6QojU=
+  - secure: iLvRKQE5WOOcc8Ivq6P80sCZACF6fJcMt8P8wW5eXHnT0ERZyve8G1wNyNmfChStm0xwvUJFFE9FfDSjhtTgja5HZUb4fcyFKQhni+qDkHx2kYsRYN7yTC1jyk9blMNpoBzyNxsCWf+uXWp+Hii/zncCOqpycMtmTi6KCYkD7hI=
--- a/ci/is_release_tag.bash
+++ b/ci/is_release_tag.bash
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+set -e
+
+tag=$1
+
+# Deployment is performed when $TRAVIS_TAG points to head of master.
+# Below is a workaround to check this condition, since Travis doesn't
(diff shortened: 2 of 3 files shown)


chuwy:
Don't forget that it should work only when tagged https://github.com/snowplow/iglu-central/releases

chuwy:
Actually, this should have even more protection. Please add one more defensive clause that checks if tagged commit exists in master: https://github.com/snowplow/snowplow/issues/3540

chuwy:
This is good to merge, @alexanderdean.

alexanderdean:
Thanks @oguzhanunlu and @chuwy. How does this ticket work? How's it wired into the release process?

chuwy:
Hey @alexanderdean, sorry this was premature. It doesn't work exactly as we planned yet - right now it works when ANY of `branch=master`, `isTag` conditions met, which is terribly unsafe: anyone can publish tag and override assets.

My goal is to make it check that BOTH conditions are met. So process is following:

* Deploy tries to start when you pushed a tag
* Script checks if tagged commit belongs to master (only you can merge to master and push)
* If yes - it proceeds to sync

For you process will be following:

* You merge to master and push (important to push the branch before)
* Attach tag
* `git push --tags`

alexanderdean:
Sounds good, keep me posted!

.travis.yml
```diff
@@ -1,7 +1,10 @@
...
-  ./ci/lint.bash
+script: "./ci/lint.bash"
+before_deploy: pip install --user awscli
+deploy:
+- provider: script
+  skip_cleanup: true
+  script: "./ci/push.bash"
+  on:
+    tags: true
+    branch: master
```
chuwy: I still don't think we need `branch: master` here, because actual deploy will happen only on tags.

ci/is_release_tag.bash
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -e
+
+tag=$1
+
+git clone https://github.com/snowplow/iglu-central.git tmp/iglu-central
```
chuwy: Can we add some detailed comments here explaining that this is a workaround for not fetched git ref.

ci/is_release_tag.bash
```diff
@@ -0,0 +1,20 @@
...
+cd tmp/iglu-central/
+head_tags=$(git tag -l --points-at refs/heads/master)
+cd ../../
+rm -rf tmp
+
+for t in $head_tags
+do
+    if [ "$t" == "$tag" ]; then
+        exit 0
```
chuwy:
```bash
echo "Tag $tag points at master branch. Proceed to deploy..."
```

ci/is_release_tag.bash
```diff
@@ -0,0 +1,20 @@
...
+    else
+        exit 1
```
chuwy:
```bash
echo "Tag $tag does not point at master branch. You need to push master before the tag. Aborting deploy..."
```

.travis.yml
oguzhanunlu: Definitely.

oguzhanunlu:
Let me address your feedback and come back @chuwy . Thanks.

oguzhanunlu:
How does it look @chuwy ?

chuwy:
LGTM! Assigning to @alexanderdean.

alexanderdean:
Cherry-picked into release branch, closing

PR #730, opened 2018-02-15
patch redash

alexanderdean:
id - this field has been removed
object_id - can be string OR int
created_at - does not conform to valid date-time


jbeemster:
@alexanderdean @chuwy sorry for missing your last comment - yes happy for this to be released!  Have also checked and it validates against the latest redash still (v5.0.2).

Do you want me to re-generate the SQL and JsonPaths?  Considering this schema is totally broken without this patch not sure there is any harm in force updating those assets as well.

jbeemster:
Force regenerated everything @chuwy !

chuwy:
Closing in favor of https://github.com/snowplow/iglu-central/pull/860. Thanks.

PR #731, opened 2018-02-19
RDB Loader R29 Schemas

chuwy:


schemas/com.snowplowanalytics.snowplow.storage/redshift_config/jsonschema/3-0-0
```diff
@@ -0,0 +1,250 @@
...
+                "sslRootCert": {
+                    "description": "The full path of a .pem or .crt file containing the trust root Certificate Authority",
+                    "type": ["string", "null"],
+                    "maxLength": 1024
+                },
+                "tcpKeepAlive": {
+                    "description": "A value that determines whether TCP keepalives are enabled",
+                    "type": "boolean"
+                },
+                "TCPKeepAliveMinutes": {
```
alexanderdean: Why is the capitalisation inconsistent with `tcpKeepAlive`

alexanderdean:
Thanks @chuwy - could you highlight the changes from the 2-0-0 just to save us some time?

schemas/com.snowplowanalytics.snowplow.storage/redshift_config/jsonschema/3-0-0
chuwy: All JDBC options are written as they're on driver's documentation.

chuwy:
Changes:

1. New `jdbc` object, configurint all JDBC driver options except a) deprecated; b) [authentication-related](https://docs.aws.amazon.com/redshift/latest/mgmt/options-for-providing-iam-credentials.html#jdbc-options-for-providing-iam-credentials) (later one is under the question - happy to add it)
2. No properties can be omitted now: everything is either present or `null` besides of `jdbc`, where all properties can be omitted (we usually omit `id` and `sshTunnel.passphrase`, now they should be added as `null`s at least)
3. `maxError` bumped to 100.000, which is valid value. Previous schemas' upper bound is 10.000
4. Biggest change - remove of `sslMode`. Now to disable SSL - you need to set `jdbc.ssl: false`. It is always enabled by default. `verify-ca` and `verify-full` (notice lowercase) should be set in dedicated `jdbc.sslMode` property.
5. Added multiple string `maxLength` and numeric `maximum` whereas documentation about bounds is available
6. Added descriptions almost for all fields. Find it quite useful especially for encrypted password and SSH tunnel

chuwy:
Just pushed an update, making `id` property required. This is necessary because `id` is what makes double-target loading possible in processing manifest world, otherwise there's no proper way to distinguish two targets.

If user previously loaded data into single target without `id` then decided to add new target and marked both targets as "id1" and "id2" - RDB Loader is now in situation where it is not possible to determine what item was loaded into what target - none items are loaded to "id1" neither to "id2". To avoid confusing situations like this, it is better to force users to stick with ids.

alexanderdean:
Makes total sense. I wanted the `id` for another uniqueness requirement we have (uniquely identifying storage targets in Snowplow Insights), but this is equally important.

chuwy:
Superseded by https://github.com/snowplow/iglu-central/pull/731

PR #741, opened 2018-03-11
Add com.marketo/event/jsonschema/1-0-0

misterpig:
JSON schema for marketo event with all possible fields. Test payloads sent through unpopulated fields as empty strings not nulls. Two fields with format "date". Many fields couldn't be populated so the string length is long (255 and 512) for the vast majority of fields. Marketo only has the one webhook so 3 custom fields are included ("name", "description, "step") to further differentiate the event.

### DIFF
--- a/schemas/com.marketo/event/jsonschema/1-0-0
+++ b/schemas/com.marketo/event/jsonschema/1-0-0
@@ -46,7 +46,7 @@
       "properties": {
         "account_owner_email_address": {
           "type": "string",
-          "maxLength": 255
+          "format": "email"
         },
         "account_owner_first_name": {
           "type": "string",
@@ -175,7 +175,7 @@
         },
         "email_address": {
           "type": "string",
-          "maxLength": 255
+          "format": "email"
           },
         "email_invalid_cause": {
           "type": "string",
@@ -267,7 +267,7 @@
         },
         "lead_owner_email_address": {
           "type": "string",
-          "maxLength": 255
+          "format": "email"
         },
         "lead_owner_first_name": {
           "type": "string",
@@ -567,7 +567,7 @@
         },
         "email": {
           "type": "string",
-          "maxLength": 255
+          "format": "email"
         }
       }
     },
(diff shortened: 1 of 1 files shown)


schemas/com.marketo/event/jsonschema/1-0-0
```diff
@@ -0,0 +1,601 @@
...
+        },
+        "do_not_call_reason": {
+          "type": "string",
+          "maxLength": 512
+        },
+        "do_not_call": {
+          "type": "boolean"
+        },
+        "email_address": {
+          "type": "string",
```
chuwy: There's `email` JSON Schema format.

alexanderdean:
Is there some associated documentation / a PR on how to use this?

BenFradet:
@alexanderdean do you mean snowplow/snowplow#3677 ?

alexanderdean:
Ace! Haven't got to that one in my email yet 🐌

snowplowcla:
@misterpig has signed the Software Grant and Corporate Contributor License Agreement

miike:
@alexanderdean We'll be putting together some documentation on how to get this setup in Marketo this week. The webhook integration is a little odd in that there's only one event that's ever sent (with a customisable payload) but that event can be triggered in multiple steps within a Marketo flow.

This schema represents a superset of these fields with the intention that a customer would likely customise this schema and host in their local Iglu repository rather than relying on this superschema in Iglu Central.

miike:
>I just wanted to remind that you can enforce this schema in adaptor. In other words, this schema should validate not input of adaptor, but output. So if you think this is necessary (I think empty strings instead of null isn't a very good idea) we can make this schema much more stricter.

@chuwy - What would be your advice on this one? Some fields may be sent through correctly as empty strings (which should be stored accordingly) whereas there may also be instances where an empty string should be NULL but is set to the webhook as an empty string. Unfortunately the Marketo [documentation](http://developers.marketo.com/rest-api/lead-database/fields/list-of-standard-fields/) doesn't provide information about which fields are nullable.

chuwy:
@miike if empty strings are valid values and we have no way to distinguish empty-string-as-nulls and valid-empty-strings - I think @misterpig's approach is the correct one.

alexanderdean:
@chuwy what is missing for your green tick?

alexanderdean:
Scheduling!

PR #743, opened 2018-03-16
Release/r88

chuwy:


alexanderdean:
Thanks @chuwy - please add Mike's PR into this one too...

alexanderdean:
Please obviously update the Iglu Central milestones in GH too so they match up with this PR.

chuwy:
Hey @alexanderdean, this is ready now.

alexanderdean:
@chuwy can you take a look through @mhadam's notification event schema? Just looking for anything unusual. I don't expect any domain knowledge from you on the event itself.

chuwy:
Hey @mhadam! https://github.com/snowplow/iglu-central/pull/743/commits/2548d905c1843b99abbbcddc8a02c0a33d471c99 looks good to me, however I'm wondering if it is possible to add some `maxLength` to strings? Problem is that they're still bounded in Redshift, but implicitly by 4096, which is not optimal value in most cases. Also few `description`s will also be useful.

chuwy:
Thanks for update @mhadam. @alexanderdean this PR is ready.

PR #753, opened 2018-04-06
Release 89

BenFradet:


BenFradet:
@alexanderdean this is good to go :+1:

chuwy:
Hey, @alexanderdean, we considered the fact about breaking JSONPaths (sorry, had to make conversation public). The reason why I approved this is because as far as I understand the problem - there's simply nothing to break: all data for `com.mandril` goes to `enriched/bad` because of this omnipresent `subaccount` property, so users either will have to migrate their tables or continue without Mandrill data.

alexanderdean:
Thanks @chuwy - that makes sense. @yalisassoon is that your understanding of the current Mandrill situation - that the webhook is not working for customers in its current form?

BenFradet:
Removed ddl and jsonpaths

CHANGELOG
```diff
@@ -1,3 +1,14 @@
+Release 89 (2018-04-06)
```
alexanderdean: Date needs changing to today

schemas/com.snowplowanalytics.snowplow/ip_lookups/jsonschema/2-0-0
```diff
@@ -0,0 +1,90 @@
...
+		"name": "ip_lookups",
+		"format": "jsonschema",
+		"version": "2-0-0"
+	},
+
+	"type": "object",
+	"properties": {
+		"vendor": {
+			"type": "string",
+            "maxLength": 256
```
alexanderdean: Whitespace is all messed up in this file (see 6 more instances below).

BenFradet: mix-indent ftw...

BenFradet:
@alexanderdean fixed :+1:

PR #754, opened 2018-04-06
Added files for adjust install event

keanerobinson:


alexanderdean:
Thanks @keanerobinson! @chuwy can you please review?

alexanderdean:
What is the hold up here @chuwy ?

alexanderdean:
Good point - @keanerobinson are you okay to look into that?

keanerobinson:
hey guys:

@chuwy - thanks for reviewing my work!

@alexanderdean - sure thing I've since made some [further changes to the schema](https://github.com/snowplow/iglu-central/commit/3878bd91a943ab3a2355791f69233a6ca2426c8c). Now all of the standardised ID fields have stricter `maxLength` entries.

I've checked the changes I've made against the values I'm seeing in Lifecake's Adjust Install table.

Anything else you need from me - please let me know!

keanerobinson:
thanks @chuwy - I've since applied the following changes:

- enum's wherever possible
- lower maxlength across several fields
- changed time fields to number types with a max of 9999999999 (they're always 10 digits)

alexanderdean:
Thanks both. Anything left from your side @chuwy ?

chuwy:
Yes. Any reasons why time fields are `number` rather than `integer`?

keanerobinson:
@chuwy good point - changed the timestamps to `int` types.

keanerobinson:
awesome stuff guys - I'll do a final test with the client I've got this up and running with now (re-upload this version of the schema).

alexanderdean:
Thanks both! We will get this released.

keanerobinson:
@alexanderdean awesome - let me know if I can help in any way!

keanerobinson:
hey guys - tested the latest version of the schema after our discussion and turns out all of the boolean + time fields we set are being sent as strings. 

I've since changed the field types back to string and imposed `maxLength` conditions. The version that I've just committed is now working in SP Mini.

bogaert:
Quite a few of the properties are enums. Is that preferred behavior for a webhook integration? Looking at the values, it seems likely at least some of those values will by changed by Adjust at some point down the line.

alexanderdean:
Hmm, that is a good point @bogaert...

bogaert:
Cool. Unless someone disagrees (@chuwy), let's update the schema to make it slightly more permissive @keanerobinson.

alexanderdean:
Take the longest ENUM value in each field and add 50-75% is a pretty safe rule.

schemas/com.adjust/install/jsonschema/1-0-0
```diff
@@ -0,0 +1,471 @@
...
+          "type": ["string", "null"],
+          "maxLength": 1
+    },
+    "is_organic": {
+          "description": "1 if organic traffic, 0 if non-organic e.g 0",
+          "type": ["string", "null"],
+          "maxLength": 1
+    },
+    "rejection_reason": {
+          "description": "Reason for a rejeted attribution e.g anonymous_traffic, distribution_outlier, too_many_engagements, engagement_injection, incorrect_signature",
```
bogaert: @keanerobinson There's a typo here (rejeted).

keanerobinson:
I've replaced the enums with `maxLength`'s and corrected the type.

Thanks for the help guys!

alexanderdean:
Woot! Will release R90 over the long w/e.

knservis:
issue https://github.com/snowplow/iglu-central/issues/733

PR #772, opened 2018-04-26
R91

chuwy:
Absence of these schemas on IC doesn't really block anything yet because we have them in embedded registry, but would be great to review their vendor (`com.snowpwlowanalytics.manifest`) and names.

### DIFF
--- a/schemas/com.snowplowanalytics.manifest/locked_payload/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.manifest/locked_payload/jsonschema/1-0-0
@@ -1,0 +1,21 @@
+{
+	"$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+	"description": "Schema for payload added by processing manifest to records denoting failure due race condition",
+	"self": {
+		"vendor": "com.snowplowanalytics.manifest",
+		"name": "locked_payload",
+		"format": "jsonschema",
+		"version": "1-0-0"
+	},
+
+	"type": "object",
+	"properties": {
+		"recordId": {
+			"description": "ID of a record which caused a race condition",
+			"type": "string",
+			"format": "uuid"
+		}
+	},
+	"required": ["recordId"],
+	"additionalProperties": false
+}
--- a/schemas/com.snowplowanalytics.snowplow.storage.rdbshredder/processed_payload/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow.storage.rdbshredder/processed_payload/jsonschema/1-0-0
@@ -1,0 +1,23 @@
+{
+	"$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+	"description": "Schema for processing manifest payload added by RDB Shredder after job has finished",
+	"self": {
+		"vendor": "com.snowplowanalytics.snowplow.storage.rdbshredder",
+		"name": "processed_payload",
+		"format": "jsonschema",
+		"version": "1-0-0"
+	},
+
+	"type": "object",
+	"properties": {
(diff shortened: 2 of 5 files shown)


schemas/com.snowplowanalytics.snowplow.storage.rdbshredder/processed_payload/jsonschema/1-0-0
```diff
@@ -0,0 +1,23 @@
...
+		"vendor": "com.snowplowanalytics.snowplow.storage.rdbshredder",
+		"name": "processed_payload",
+		"format": "jsonschema",
+		"version": "1-0-0"
+	},
+
+	"type": "object",
+	"properties": {
+		"shreddedTypes": {
+			"description": "List of shredded types incountered within single batch",
```
alexanderdean: "List of shredded types encountered within a single batch"

schemas/com.snowplowanalytics.manifest/locked_payload/jsonschema/1-0-0
```diff
@@ -0,0 +1,21 @@
...
+		"vendor": "com.snowplowanalytics.manifest",
+		"name": "locked_payload",
+		"format": "jsonschema",
+		"version": "1-0-0"
+	},
+
+	"type": "object",
+	"properties": {
+		"recordId": {
+			"description": "Id of a record caused race condition",
```
alexanderdean: "ID of a record which caused a race condition"

alexanderdean:
Adding @BenFradet for a second review.

@chuwy this will be R91 (nothing else going into R91). I will release R90 soon...

schemas/com.snowplowanalytics.snowplow/application_error/jsonschema/1-0-2
```diff
@@ -0,0 +1,90 @@
...
+        "ACTIONSCRIPT",
+        "LUA",
+        "RUST",
+        "HASKELL",
+        "CLOJURE",
+        "ERLANG",
+        "ELIXIR",
+        "CRYSTAL",
+        "PONY",
+        "NIM"
```
BenFradet: :smile:

schemas/com.snowplowanalytics.snowplow/application_error/jsonschema/1-0-2
```diff
@@ -0,0 +1,90 @@
...
+      "minimum": 0,
+      "maximum": 2147483647
+    },
+    "stackTrace": {
+      "type": ["string", "null"],
+      "maxLength": 8192
+    },
+    "causeStackTrace": {
+      "type": ["string", "null"],
+      "maxLength": 8192
```
BenFradet: are we sure about the length of stack traces?

chuwy: Actually, not at all, I don't even think there's an upper bound. But I guess good news is that usually most useful information is on top of stacktrace and it is intended to be human-readable, so we don't care that much if it was truncated.

BenFradet: yes that's what I meant, will I have to truncate my stack traces to integrate with this schema and do I want to be doing that?

chuwy: Yes, we'll have to truncate it. Though these events are usually sent via middleware, such as trackers or processing manifest and it is up to middleware to construct valid payload. We already truncate it in Scala tracker and I think must do it in all other trackers.

chuwy: ...and obviously in processing manifest as well.

alexanderdean:
@chuwy could you integrate all other PRs/issues from R91 and prepare for release?

PR #776, opened 2018-05-07
Add com.snowplowanalytics.iglu/resolver-config/jsonschema/1-0-3

oguzhanunlu:


chuwy:
I don't think this needs to be in R91. It is basically useless until we released a client supporting it.

oguzhanunlu:
I added in R91 just because issue was included in that.

chuwy:
Ok, makes sense! Could you please create associated ticket in Iglu Client.

PR #784, opened 2018-06-07
Factotum Schema Patch

jbeemster:
Hey @ninjabear @alexanderdean - so to use a higher string length in factotum we also need to patch these schemas.  I have opted for simply removing the limit on these fields as this gives us a lot more flexibility in terms of playing with much higher limits.  The SQL DLL also does not reference these fields directly as it is an array so nothing needs to be patched at that level.

Thoughts / qualms about this?

### DIFF
--- a/schemas/com.snowplowanalytics.factotum/job_update/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.factotum/job_update/jsonschema/1-0-0
@@ -114,12 +114,10 @@
 						"type": "string"
 					},
 					"stdout": {
-						"type": "string",
-						"maxLength": 10000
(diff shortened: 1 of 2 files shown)


jbeemster:
Cool thanks @ninjabear !   @alexanderdean when do you think we will be able to get this merged in?

alexanderdean:
Green tick from me - @chuwy would you mind pulling this into a new R92 release, getting that PR ready and adding to the Sprint board? Then assign the PR to me. Assume release today or tomorrow.

chuwy:
Closing in favor of https://github.com/snowplow/iglu-central/pull/785

PR #807, opened 2018-07-12
Add events to enrichment settings file (Closes #806)

### LINKED ISSUE #806: Add events to enrichment settings file
First step for: snowplow/snowplow#3826
We would have a filed in the enrichments settings to make that enrichment optional for the that specific event name.

szareiangm:
Added `events` to all of the enrichments.


szareiangm:
So do you mean that I should create a data structure definition that can be `event_type (page_view, struct etc) and unstruct + iglu UR`? And then I would have a list of this data structure?

szareiangm:
I ended up with this:
```javascript
"events" : {
    "type" : [ "array", "null" ],
    "items" : {
        "anyOf": [
            {
                "type": "object",
                "properties": {
                    "eventType": {
                        "type": "string",
                        "enum": ["struct", "unstruct",
                                  "ad_impression", "transaction", "transaction_item", "page_view","page_ping"]
                    }
                },
                "required": ["eventType"],
                "additionalProperties": false
            },
            {
                "type": "object",
                "properties": {
                    "schema": {
                        "type": "string",
                        "pattern": "^iglu:([a-zA-Z0-9-_.]+)/([a-zA-Z0-9-_]+)/([a-zA-Z0-9-_]+)/([1-9][0-9]*|\\*)-((?:0|[1-9][0-9]*)|\\*)-((?:0|[1-9][0-9]*)|\\*)$"
                    }
                },
                "required": ["schema"],
                "additionalProperties": false
            }
        ]
    },
    "description": "Run this enrichment for the event names in the list. Use null to disable"
},
```
Does it follow your points? @chuwy

chuwy:
Hey @szareiangm! Yes, that's exactly right. Except that I'd use `oneOf` instead of `anyOf` as `anyOf` means we can use something like `{"eventType": "ad_impression", "schema": "iglu:com.acme/event/jsonschema/1-0-2"}`, which does not make much sense. 

Otherwise 👍

szareiangm:
What if someone wants to have the enrichment for `page_view` and a custom unstruct event for `user_registration`.  if we use `oneOf` we should have the `iglu:...` for `page_view` too? @chuwy

chuwy:
That would mean that then need to have both objects in config:

```json
[
  {"eventType": "page_view"},
  {"schema": "iglu:com.acme/user_registration/jsonschema/1-0-2"}
]
```

Any particular event cannot be `page_view` and `user_registration` ("classic" event and unstruct event) at the same time.

chuwy:
By the way, I'd go even closer and make my above statement more explicit in config by making unstruct event schema following:

```json
{
    "type": "object",
    "properties": {
        "eventType": {
            "type": "string",
            "enum": ["unstruct"]
        },
        "schema": {
            "type": "string",
            "pattern": "^iglu:([a-zA-Z0-9-_.]+)/([a-zA-Z0-9-_]+)/([a-zA-Z0-9-_]+)/([1-9][0-9]*|\\*)-((?:0|[1-9][0-9]*)|\\*)-((?:0|[1-9][0-9]*)|\\*)$"
        }
    },
    "required": ["eventType", "schema"],
    "additionalProperties": false
}
```

It means that `schema` is applicable only when `eventType` is `unstruct`. On the other hand we would also be able to filter only `unstruct` events by not specifying any `schema`.

szareiangm:
Then, your example is an `array` of `anyOf` either these two objects? (like my proposal)
or two separate `array` of these objects? (this will create two separate configuration entry: `eventType` and `schema`) @chuwy

chuwy:
No, this is still same array that can contain multiple objects which are either "classic" event OR unstruct event.

```haskell
List of [Classic enum | Unstruct schema]
```

or (shortened format):

```scala
List(
  Classic("page_view"),
  Unstruct("iglu:com.acme/user_registration/jsonschema/1-0-2"),
  Classic("ad_impression"),
  Unstruct("iglu:com.acme/foo/jsonschema/1-0-2")
)
```

Oh, and one more point. Unstruct event's schemas must be in schema "crititerion format" (`iglu:com.acme/user_registration/jsonschema/1-*-*`). Otherwise user will have to add each new schema manually, which is very error-prone.

szareiangm:
Is it close to your idea if I add  * for the regex for your last point?
```javascript
        "events" : {
            "type" : [ "array", "null" ],
            "items" : {
                "oneOf": [
                    {
                        "type": "object",
                        "properties": {
                            "eventType": {
                                "type": "string",
                                "enum": ["struct",
                                          "ad_impression", "transaction", "transaction_item", "page_view","page_ping"]
                            }
                        },
                        "required": ["eventType"],
                        "additionalProperties": false
                    },
                    {
                        "type": "object",
                        "properties": {
                            "eventType": {
                                "type": "string",
                                "enum": ["unstruct"]
                            },
                            "schema": {
                                "type": "string",
                                "pattern": "^iglu:([a-zA-Z0-9-_.]+)/([a-zA-Z0-9-_]+)/([a-zA-Z0-9-_]+)/([1-9][0-9]*|\\*)-((?:0|[1-9][0-9]*)|\\*)-((?:0|[1-9][0-9]*)|\\*)$"
                            }
                        },
                        "required": ["eventType", "schema"],
                        "additionalProperties": false
                    }
                ]
            },
            "description": "Run this enrichment for the event names in the list. Use null to disable"
        }
```

chuwy:
Yep, thank you @szareiangm. Regex seems to be [valid](https://github.com/snowplow/iglu/blob/develop/0-common/scala-core/src/main/scala/com.snowplowanalytics.iglu/core/SchemaCriterion.scala#L126-L133).

szareiangm:
Yay!
I saw the `\\*` now. I will change the files to reflect our discussion in a few minutes.

szareiangm:
I updated my PR with the new changes.

szareiangm:
@chuwy I have a question. Converting this schema to a case class needs a serializer (`oneOf`). The only similar code that I could find was in `schemaddl` library.  Do you have any ideas how to do serialize this without writing too much duplicate code or importing a library for just one line?

snowplowcla:
@szareiangm  has signed the Software Grant and Corporate Contributor License Agreement

PR #809, opened 2018-07-12
Add com.amazon.aws.cloudfront/wd_access_log/jsonschema/1-0-6

BenFradet:


schemas/com.amazon.aws.cloudfront/wd_access_log/jsonschema/1-0-6
```diff
@@ -0,0 +1,112 @@
...
+    },
+    "csProtocolVersion": {
+      "type": ["string", "null"],
+      "maxLength": 8
+    },
+    "fleStatus": {
+      "type": ["string", "null"],
+      "maxLength": 64
+    },
+    "fleEncryptedFields": {
```
chuwy: Shouldn't it be integer?  From [docs](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/AccessLogs.html) it looks like it is a natural number

BenFradet: problem is that it can be "-" too

alexanderdean:
We are hoping to get R94 out today - can we get a green tick?

schemas/com.amazon.aws.cloudfront/wd_access_log/jsonschema/1-0-6
chuwy: Okay, makes sense.

alexanderdean:
Merged

PR #810, opened 2018-07-13
Add: com.getvero schemas

misterpig:
Old PR can be found here #742


schemas/com.getvero/bounced/jsonschema/1-0-0
```diff
@@ -12,25 +12,60 @@
...
+			"type": ["string", "null"],
+			"maxLength": 255
+		},
+		"message_id": {
+			"type": ["string", "null"],
 			"maxLength": 255
 		},
 		"type": {
 			"type": "string",
 			"maxLength": 12
```
rzats: Please extend the `maxLength` for all `type` fields just for safety, since e.g. the "user_unsubscribed" type might be renamed to something like "user_successfully_unsubscribed" in the future which would break the schema

schemas/com.getvero/bounced/jsonschema/1-0-0
```diff
@@ -12,25 +12,60 @@
...
+					"maxLength": 255
+				},
+				"triggered_at": {
+					"type": ["string", "null"],
+					"format": "date-time"
+				},
+				"data": {
+					"type": "object",
+					"properties": {
+						"key": {
```
rzats: `data` is most likely meant to be a dictionary-like object that can store arbitrary key/value pairs. From Vero's example event:

`"data":{"key":"value"}`

I'd just set `"additionalProperties": true` here with no required properties. (Also applies to `event.data` fields in other schemas)

chuwy: If you're aware of content of keys or values, you can also use `patternProperties` to restrict them to specific types.

alexanderdean: Agree with all that

alexanderdean:
@rzats while we are waiting on some tweaks from Snowflake, please can you stub out all the relevant *tickets* in R94.

snowplowcla:
@misterpig  has signed the Software Grant and Corporate Contributor License Agreement

alexanderdean:
Merged

PR #821, opened 2018-07-13
R94

rzats:


alexanderdean:
Missing the Prepare for release commit

alexanderdean:
@chuwy @BenFradet could you guys expedite review of this? Many thanks!

chuwy:
Sure, will do.

schemas/com.getvero/bounced/jsonschema/1-0-0
```diff
@@ -12,25 +12,55 @@
 	"properties": {
 		"bounced_at": {
 			"type": "string",				
-			"format": "date-time"
+			"format": "date-time",
+			"maxLength": 10
```
chuwy: What string could include date and time AND be 10 chars long? The only ISO8601-compatible 10 chars string is `YYYY-MM-DD`, but it is clearly only date.

rzats: I think the datetime strings are all UNIX timestamps: https://help.getvero.com/articles/setting-up-veros-webhooks

schemas/com.snowplowanalytics.snowplow.enrichments/iab_spiders_and_robots_enrichment/jsonschema/1-0-0
```diff
@@ -64,25 +64,13 @@
...
                     },
                     "required": ["database", "uri"]
-                },
-                "httpUsername": {
-                    "description": "Optional username for basic HTTP authentication",
-                    "type": "string"
-                },
-                "httpPassword": {
-                    "description": "Optional password for basic HTTP authentication",
-                    "type": "string"
```
BenFradet: why are we removing this completely?

alexanderdean: I spoke to @rzats about it. As I understand it:

* He never implemented the functionality
* He didn't document it

Therefore it makes more sense to remove it entirely given the release urgency.

If that's wrong or you disagree - @rzats or @BenFradet , please shout

schemas/com.getvero/bounced/jsonschema/1-0-0
chuwy: * `date-time` means that  it is ISO8601-compatible **string** containing both date and time
* UNIX timestamp is an **integer** containing both date and time

If we're just stringifying UNIX timestamp in SCE (which doesn't make much sense) then it won't have `date-time` as it won't be  ISO8601-compatible. If we're converting it to proper `date-time` then it is impossible to have 10 characters length. If we're converting it to `YYYY-MM-DD` (which also doesn't make sense) then it should be just `date`. All these options are mutually exclusive.

alexanderdean: Okay, this is a blocker on release @rzats. What is the story here?

alexanderdean: @BenFradet in case @rzats doesn't get back to us in a timely fashion - would you mind looking in the code and telling us which option "must" be correct?

rzats: https://github.com/snowplow/iglu-central/commit/5eec605939970b87781ab150fe0ea9f1ee1bb0a8 max-length has been removed from this field in the snowflake branch, which is presumably the correct option.

BenFradet:
integrated what was changed in #810, namely:

```json
"data": {
  "type": ["object", "null"],
  "additionalProperties": true
}
```

could you have another look please @chuwy ?

BenFradet:
Thanks I'll regenerate and merge :+1:

BenFradet:
@alexanderdean can you merge, it seems Anton and I can't.

PR #825, opened 2018-08-08
Igluctl static deploy command configuration

oguzhanunlu:
This is an early draft for configuration of igluctl's on-the-road `static deploy` command.

Here is an [example config](https://gist.github.com/oguzhanunlu/9f8d943128f9d80bd1137a7d6f0c4a33).

### DIFF
--- a/schemas/com.snowplowanalytics.iglu/igluctl_config/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.iglu/igluctl_config/jsonschema/1-0-0
@@ -1,0 +1,201 @@
+{
+	"$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+	"description": "Schema for Igluctl's configuration",
+	"self": {
+		"vendor": "com.snowplowanalytics.iglu",
+		"name": "igluctl_config",
+		"format": "jsonschema",
+		"version": "1-0-0"
+	},
+	"type": "object",
+	"properties": {
+		"env": {
+			"description": "Target environment of the configuration",
+			"type": "string",
+			"enum": ["test", "mini", "prod"]
+		},
+		"input": {
+			"description": "Path to input schemas",
+			"type": "string",
+			"format": "uri"
+		},
+		"lint": {
+			"description": "Schema for lint command",
+			"type": "object",
+			"properties": {
+				"skipWarnings": {
+					"description": "Don't output messages with log level less than ERROR",
+					"type": "boolean",
+					"default": false
+				},
+				"includedChecks": {
+					"description": "Lint using provided linters",
+					"type": ["array", "null"],
+					"items" : {
+						"type" : "string",
+						"enum": [
+							"rootObject",
+							"unknownFormats",
+							"numericMinMax",
+							"stringLength",
+							"optionalNull",
+							"description"
+						]
+					}
+				}	
+			},
+			"required": ["includedChecks"]
+		},
+		"generate": {
+			"description": "Schema for static generate command",
+			"type": "object",
+			"properties": {
+				"output": {
+					"description": "Directory to put generated data",
+					"type": "string",
+					"default": "."
+				},
+				"dbschema": {
+					"description": "Redshift database schema (e.g. 'atomic')",
+					"type": "string",
+            		"minLength": 1,
+            		"maxLength": 128,
+					"default": "atomic"
+				},
+				"owner": {
+					"description": "Redshift table owner",
+					"type": "string",
+					"maxLength": 128
+				},
+				"db": {
+					"description": "DB to which we need to generate DDL",
+					"type": "string",
+					"minLength": 1,
+					"maxLength": 64,
+					"default": "redshift"
+				},
+				"varcharSize": {
+					"description": "Redshift varchar size",
+					"type": "integer",
+					"minimum": 1,
+					"maximum": 65535,
+					"default": 4096
+				},
+				"withJsonPaths": {
+					"description": "Produce JSON Paths files with DDL or not",
+					"type": "boolean",
+					"default": true
+				},
+				"rawMode": {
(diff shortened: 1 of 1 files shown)


schemas/com.snowplowanalytics.iglu/igluctl_config/jsonschema/1-0-0
```diff
@@ -0,0 +1,201 @@
...
+	"description": "Schema for Igluctl's configuration",
+	"self": {
+		"vendor": "com.snowplowanalytics.iglu",
+		"name": "igluctl_config",
+		"format": "jsonschema",
+		"version": "1-0-0"
+	},
+	"type": "object",
+	"properties": {
+		"env": {
```
chuwy: I'm still hazy on this field. What would be the semantic difference? Does Igluctl behave differently on various envs?

schemas/com.snowplowanalytics.iglu/igluctl_config/jsonschema/1-0-0
```diff
@@ -0,0 +1,201 @@
...
+			"type": "string",
+			"enum": ["test", "mini", "prod"]
+		},
+		"input": {
+			"description": "Path to input schemas",
+			"type": "string",
+			"format": "uri"
+		},
+		"lint": {
+			"description": "Schema for lint command",
```
chuwy: > Schema for lint command

In `description` we're describing a value and its purpose, not the schema.

schemas/com.snowplowanalytics.iglu/igluctl_config/jsonschema/1-0-0
```diff
@@ -0,0 +1,201 @@
...
+			"properties": {
+				"output": {
+					"description": "Directory to put generated data",
+					"type": "string",
+					"default": "."
+				},
+				"dbschema": {
+					"description": "Redshift database schema (e.g. 'atomic')",
+					"type": "string",
+            		"minLength": 1,
```
chuwy: Broken indentation.

schemas/com.snowplowanalytics.iglu/igluctl_config/jsonschema/1-0-0
```diff
@@ -0,0 +1,201 @@
...
+							"stringLength",
+							"optionalNull",
+							"description"
+						]
+					}
+				}	
+			},
+			"required": ["includedChecks"]
+		},
+		"generate": {
```
chuwy: `generate` is missing `required`. In fact, we want to make all fields required, but possibly `null`. This approach is more templating-friendly.

schemas/com.snowplowanalytics.iglu/igluctl_config/jsonschema/1-0-0
```diff
@@ -0,0 +1,201 @@
...
+								"pattern": "^[A-Z0-9-_]+"
+							},
+							"profile": {
+								"description": "AWS profile",
+								"type": "string"
+							},
+							"region": {
+								"description": "AWS region",
+								"type": "string",
+								"enum": [
```
chuwy: I wouldn't do that. AWS SDK is able to upload files to region that appeared after this particular release of AWS SDK. And user will be blocked by this schema if new regions appears.

schemas/com.snowplowanalytics.iglu/igluctl_config/jsonschema/1-0-0
```diff
@@ -0,0 +1,201 @@
...
+							}
+						}
+					}
+				]
+			}
+		}
+	},
+	"required": ["input", "lint", "generate"],
+	"additionalProperties": false
+}
```
chuwy: Newline.

chuwy:
Hey @keanerobinson, @colmsnowplow, could you actually also chime in and check out [the example](https://gist.github.com/oguzhanunlu/9f8d943128f9d80bd1137a7d6f0c4a33). What do you think of it? Would it be easier for you to work with such config?

schemas/com.snowplowanalytics.iglu/igluctl_config/jsonschema/1-0-0
oguzhanunlu: We expect users to have multiple configs and `env` field would be used to identify a config between multiple ones. Igluctl's behavior doesn't change per `env`. 

Should we rely on only file name to identify which config file will be used for which environment (test, mini, prod)?

chuwy: Yes, I'd expect users rely on filename only. We could have something like `description` to say that "This config was used for initial com.acme testing on Mini 0.5.0", but just `test`/`prod` surely not enough.

So we can leave human readable `name` or `description`, but don't think `env` fits the purpose.

schemas/com.snowplowanalytics.iglu/igluctl_config/jsonschema/1-0-0
```diff
@@ -0,0 +1,274 @@
...
+									"null"
+								],
+								"default": "us-west-2"
+							}
+						},
+						"required": [
+							"action",
+							"uploadFormat",
+							"bucketPath",
+							"accessKeyId",
```
chuwy: Most likely we'll never have keys in config. They shouldn't be required.

chuwy: AWS has multiple was to get permissions inside the code:

1. `profile`
2. `DefaultAWSCredentialsProviderChain`

And few more, but those two are already implented in igluctl.

colmsnowplow:
> Hey @keanerobinson, @colmsnowplow, could you actually also chime in and check out the example. What do you think of it? Would it be easier for you to work with such config?

Apologies, I forgot to come back on this. In short yes would be a lot easier. Also I think this'll make a big difference to new users learning the schema process, especially people who are daunted by the command line. Great stuff @oguzhanunlu and @chuwy!

PR #827, opened 2018-08-09
Add com.snowplowanalytics.snowplow.r2f/removal_criteria/jsonschema/1-0-0

knservis:
Number of issues to resolve by the reviewers:

- [x] The list of acceptable pojo fields has simply been copied from PII, but do they all make sense for R2F or could they lead to removing data for multiple people?
- [x] Is r2f an acceptable name?
- [x] Is r2f in an acceptable namespace (com.snowplowanalytocs.snowplow)?

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow.r2f/removal_criteria/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow.r2f/removal_criteria/jsonschema/1-0-0
@@ -1,0 +1,98 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for an individual entry in an R2F (Right to be forgotten) criteria file, each entry specifying an identifier for events that should be deleted.",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow.r2f",
+    "name": "removal_criteria",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "pojo": {
+      "description": "A scalar field from the EnrichedEvent POJO (e.g. user_id)",
+      "type": "object",
+      "properties": {
+        "fieldName": {
+          "enum": [
+            "user_id",
+            "user_ipaddress",
+            "user_fingerprint",
+            "domain_userid",
+            "network_userid",
+            "ip_organization",
+            "ip_domain",
+            "tr_orderid",
+            "ti_orderid",
+            "mkt_term",
+            "mkt_content",
+            "se_category",
+            "se_action",
+            "se_label",
+            "se_property",
+            "mkt_clickid",
+            "refr_domain_userid",
+            "domain_sessionid"
+          ],
+          "description": "The name of the field"
+        },
+        "valueForWhichEventWillBeDeleted": {
+          "type": "string",
+          "description": "The value that, if present in that field, will cause the event to be deleted"
+        }
+      },
+      "additionalProperties": false,
+      "required": [
+        "fieldName",
+        "valueForWhichEventWillBeDeleted"
+      ]
+    },
+    "json": {
+      "description": "A JSON field from the EnrichedEvent POJO (e.g. contexts).",
+      "type": "object",
+      "properties": {
+        "fieldName": {
+          "enum": [
+            "contexts",
+            "derived_contexts",
+            "unstruct_event"
+          ],
+          "description": "The name of the field"
+        },
+        "jsonPath": {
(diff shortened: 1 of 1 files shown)


alexanderdean:
Let's keep the naming discussion on the other thread.

I think the list of acceptable pojo fields is a really interesting question. Interested in @yalisassoon's thoughts here.

jsonpaths/com.snowplowanalytics.snowplow/r2f_data/jsonschema/1-0-0
```diff
@@ -0,0 +1,98 @@
...
+          ],
+          "description": "The name of the field"
+        },
+        "jsonPath": {
+          "type": "string",
+          "description": "The JsonPath that was specified"
+        },
+        "valueForWhichRowWillBeDeleted": {
+          "type": "string",
+          "description": "The value that is present in a row ofr that field, will cause it to be deleted"
```
chuwy: `ofr` -> `row`

jsonpaths/com.snowplowanalytics.snowplow/r2f_data/jsonschema/1-0-0
```diff
@@ -0,0 +1,98 @@
...
+            "se_property",
+            "mkt_clickid",
+            "refr_domain_userid",
+            "domain_sessionid"
+          ],
+          "description": "The name of the field"
+        },
+        "valueForWhichRowWillBeDeleted": {
+          "type": "string",
+          "description": "The value that is present in a row ofr that field, will cause it to be deleted"
```
chuwy: `ofr` -> `for`

jsonpaths/com.snowplowanalytics.snowplow/r2f_data/jsonschema/1-0-0
```diff
@@ -0,0 +1,98 @@
...
+            "se_action",
+            "se_label",
+            "se_property",
+            "mkt_clickid",
+            "refr_domain_userid",
+            "domain_sessionid"
+          ],
+          "description": "The name of the field"
+        },
+        "valueForWhichRowWillBeDeleted": {
```
chuwy: Any chance for better name? `valueToDelete` or just `value` having that we don't call `filedName` a `fieldWhereValueForWhichRowWillBeDeletedShouldBeFound`?

knservis:
@chuwy Thanks for that:

In regards to:

> Also, I'm sorry if it was discussed anywhere already, but shouldn't it contain array of fieldName/value pairs? Running a job just to delete one value sounds wasteful for me.

There will be a list of them. This is a schema on each line of the list.

jsonpaths/com.snowplowanalytics.snowplow/r2f_data/jsonschema/1-0-0
knservis: Please suggest a better name that both:

* Does state intent.
* Does not mislead the user into believing that the value alone, instead of the entire row will be deleted.

alexanderdean:
Actually @knservis @chuwy I think this is a non-issue and we don't need to wait on Yali - for two reasons:

1. We have the dead man's handle which should stop catastrophic event deletion through a poorly configured forget file
2. Reduce to the absurd, all properties will eventually move out of atomic.events into schemas (this is the atomic.events refactor), and therefore worrying about guarding certain fields in the POJO for now, is not time well-spent (as those fields will all ultimately be up-for-grabs in the future)

TL;DR we can proceed with the current list of fields.

knservis:
Thanks @alexanderdean. In which case @alexanderdean @BenFradet and @chuwy is there something else remaining in this PR?

schemas/com.snowplowanalytics.snowplow.r2f/removal_criteria/jsonschema/1-0-0
```diff
@@ -0,0 +1,98 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for R2F (Right to be forgotten) criteria files that contain the identifiers for events that should be deleted.",
```
BenFradet: should we specify that it's for only one criteria, reading it it made me think it was the schema for the file.

alexanderdean: Let's change:

> Schema for R2F (Right to be forgotten) criteria files that contain the identifiers for events that should be deleted.

to:

> Schema for an individual entry in an R2F (Right to be forgotten) criteria file, each entry specifying an identifier for events that should be deleted.

PR #839, opened 2018-09-11
Fixes: com.getvero

misterpig:
Needed to change the property type for `campaign.id` as it'll invalidate Vero payloads (str -> int).
Couple other updates: removed the `type` fields as they're redundant and changed the `campaign.permalink` field property format to `uri`. If these updates should be a separate commit for a later update let me know and I'll revert these two changes.


misterpig:
Hi @chuwy , you're right in that the schema at the moment won't break the pipeline and will accept the Vero payloads. 
Increasing the `permalink` field form VARCHAR(255) to format("uri") is just to ensure that a long url won't invalidate the payload. The `type` field is still present in the payload from Vero which the adapter uses to determine which event we're validating our schemas against. However I understand that including the event `type` in the output is redundant as the adapter removes the `type` field.
These changes are all amendments so I'll consider them for an version update in the future as you've pointed out.
Cheers.

misterpig:
Hey @chuwy, I can see that the file changes to the jsonpath files are strange in that as per your 2. point regarding the additional optional columns in my commit but I haven't added new columns to the schemas, jsonpath and sql files. The jsonpath files in master and my commit only differ in the removal of the `data.type` field.
Any idea what's going on here?

chuwy:
Sorry @misterpig, I'm not following. [Here](https://github.com/snowplow/iglu-central/pull/839/files#diff-f917a13e6c1bfe344113566976b95b56R21) your commit adds several new fields to `bounced` schema that are reflected in JSONPaths and tables. It is an expected behavior if you're patching existing schemas.

misterpig:
Ok this PR is trying to merge an old branch with all the old commits. I'm closing this one and create PR from another new branch.

PR #841, opened 2018-09-14
Add com.snowplowanalytics.mobile/application/jsonschema/1-0-0

keanerobinson:


schemas/com.snowplowanalytics.mobile/application/jsonschema/1-0-0
```diff
@@ -0,0 +1,25 @@
...
+		"version": "1-0-0"
+	},
+	"type": "object",
+	"properties": {
+		"version": {
+			"type": "string",
+			"description": "Version of the application e.g 1.1.0",
+			"maxLength": 64
+		},
+		"build": {
```
chuwy: I'm sorry, I'm not really aware of context (pun!) of this task, but I don't think that any application has a `build`, e.g. I cannot think of any build ids in my CLI nor Server apps, so I assume this is optional. Also description of `build` is not very useful.

schemas/com.snowplowanalytics.mobile/application/jsonschema/1-0-0
```diff
@@ -0,0 +1,25 @@
...
+		"name": "application",
+		"format": "jsonschema",
+		"version": "1-0-0"
+	},
+	"type": "object",
+	"properties": {
+		"version": {
+			"type": "string",
+			"description": "Version of the application e.g 1.1.0",
+			"maxLength": 64
```
chuwy: I think 16 or 32 should be enough for longest types of versioning.

schemas/com.snowplowanalytics.mobile/application/jsonschema/1-0-0
```diff
@@ -0,0 +1,25 @@
+{
+	"$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+	"description": "Schema for an application context",
```
chuwy: Also, not very descriptive. Is it something to uniquely identify an application? What kind of application? Mobile? Desktop? Server?

chuwy:
Also @keanerobinson, could you please edit the commit message to conform https://github.com/snowplow/iglu-central/issues/136

schemas/com.snowplowanalytics.mobile/application/jsonschema/1-0-0
keanerobinson: will amend

schemas/com.snowplowanalytics.mobile/application/jsonschema/1-0-0
keanerobinson: will amend

schemas/com.snowplowanalytics.mobile/application/jsonschema/1-0-0
chuwy: Sorry, I actually just noticed `mobile` in vendor, which both answers some of my questions and slightly confuses me. I think something helping to uniquely identify an application is very useful across all platforms, not just mobile. Beside of current tasks, is there any reasons to restrict it to be mobile-only?

keanerobinson:
hey @alexanderdean the schema is intentionally minimal as the plan is to use it solely for automated build / version tracking for users of our iOS / Android SDK's. Myself and @mhadam didn't find any other properties worth tracking out-of-the-box across iOS / Android. 

Furthermore, in most conversations about this with our mobile customers application build / version are the only fields that have been requested (Lifecake / In Crowd Sports / Hudl / Hush).

schemas/com.snowplowanalytics.mobile/application/jsonschema/1-0-0
keanerobinson: @chuwy apologies I probably should've given more context in my email earlier.

A big focus for the solutions engineering team is the Mobile Foundation (part of the Services OKR's). The whole idea behind Mobile Foundation is to decrease the time to value for our iOS / Android SDK users. The approach as a result is user-led. One decision that we made was to start using a separate vendor for any schema releases as it's a better user experience.

Fully agree that tracking app versions applies not just to Mobile apps but the "Mobile Foundation" release of the application context will be used just for our users of the iOS / Android trackers. @mhadam has already worked on functionality for the iOS/Android trackers to start tracking this automatically which is why we've written a dedicated schema.

alexanderdean:
Makes sense, thanks

schemas/com.snowplowanalytics.mobile/application/jsonschema/1-0-0
```diff
@@ -0,0 +1,25 @@
...
+	"properties": {
+		"version": {
+			"type": "string",
+			"description": "Version number of the application e.g 1.1.0",
+			"maxLength": 32
+		},
+		"build": {
+			"type": "string",
+			"description": "Build name of the application e.g s9f2k2d or 1.1.0 beta",
+			"maxLength": 32
```
alexanderdean: I would make these fields longer. All it takes is for somebody to be using SHA1 hashes or something in their version numbers and we will be in trouble. There's no upside to setting an arbitrarily low limit.

keanerobinson:
Closing this in favour of the [original PR](https://github.com/snowplow/iglu-central/pull/843) which I've branched off and resolved the issues for.

Issue #842, opened 2018-09-17
Patch: com.getvero


misterpig:
Hey @chuwy, regarding the Vero schemas there is a bug that is currently invalidating 5/8 events that is sent from Vero. The PR from last week was committed to an old branch so there were lots of conflicts and I closed it. Since the PR will be for a patch on the schemas is it also possible to include a couple of other changes (removing the `type` field and changing another field's property from VARCHAR to "uri" format). None of these should invalidate any historical data because no data would have been validated up till now.

chuwy:
Hi @misterpig,

For me it sounds like two following statements are mutually exclusive:

> bug that is currently invalidating 5/8 events

> no data would have been validated up till now

If *any* data with old schema slipped through - we cannot patch it with any breaking changes. But lets see how these change look like.

Also, as I mentioned earlier `type` is used in a Scala Common Enrich, so we want to be confident whether we can use or not and what safe fallback do we have (cc @rzats).

Also, our [commit history convention](https://github.com/snowplow/iglu-central/commits/master?after=11f0b087d1bae06b079174b3dbf7cbd0ec2b99f9+34) implies that every schema change/addition has its own commit

PR #844, opened 2018-09-18
Patch: com.getvero

misterpig:
So the bug exists in the `campaign.id` field. Turns out the id field of this is an integer unlike the `user.id` field.
Removing the `type` field wouldn't impact the adapter as the adapter removes the type field anyway; keeping type field in the schema is redundant as it'll contain NULL for every event after enrichment.
Also increased the `campaign.permalink` field to accomodate long URLs.


alexanderdean:
Thanks @misterpig !

snowplowcla:
@misterpig  has signed the Software Grant and Corporate Contributor License Agreement

BenFradet:
@chuwy how should we proceed here? background is in #842

chuwy:
Hey @BenFradet, @misterpig,

I still have the same opinion I left in #842:

* If *any* amount of data (5/8 as far as I can understand) got to DB, we cannot accept this as a patch because we'll need to migrate the table, otherwise it will just break loading process
* We can accept it as `2-0-0` without problems
* Commit history needs to follow ticket:commit 1:1 convention

chuwy:
And also as I mentioned before, it seems we're using `type` property in SCE logic, so I'm not sure we can safely remove it.

BenFradet:
ok thanks, leaving in the 101 milestone for now.

@misterpig if you could do the necessary changes, i.e. move to 2-0-0, we could integrate it into release 113.

PR #847, opened 2018-09-24
R97

chuwy:


alexanderdean:
Is this ready for me to review @chuwy ?

chuwy:
It is now.

chuwy:
@alexanderdean  just updated the release date, in case you will have time to publish it today/tomorrow.

chuwy:
Thanks @alexanderdean, I've addressed your feedback and update the date in CHANGELOG to 2nd of October.

jethron:
Any ETA on this guys?

alexanderdean:
This PR is safe to merge because the RDB Loader uses `TRUNCATECOLUMNS` when loading the shredded tables:

https://github.com/snowplow/snowplow-rdb-loader/blob/master/src/main/scala/com/snowplowanalytics/snowplow/rdbloader/loaders/RedshiftLoadStatements.scala#L212

(without this, the widening of the Id fields in the JSON Schema would immediately cause Redshift loads to break).

PR #857, opened 2018-11-21
Patch com.snowplowanalytics.monitoring.kinesis/storage_write_failed

jbeemster:
Hey @chuwy this seems to be the last "bad" event we have flowing through our internal pipeline so patching this would mean pretty much no data going to bad anymore!

Failures seem to be generated from the Elasticsearch Loader - I have opened a ticket there to explore why this appears to be sending null messages:

https://github.com/snowplow/snowplow-elasticsearch-loader/issues/124

### DIFF
--- a/schemas/com.snowplowanalytics.monitoring.kinesis/storage_write_failed/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.monitoring.kinesis/storage_write_failed/jsonschema/1-0-0
@@ -30,7 +30,7 @@
 			"minimum": 0
 		},
 		"message": {
-			"type": "string"
+			"type": ["string", "null"]
 		}
 	},
 


jbeemster:
Well currently all of that monitoring information is going to the bin and will continue to do so until we:

1. Patch and release a new version of this application
2. Update all of our deployed apps (quite a slow upgrade cadence currently)

For brevity and getting things working now I would really prefer allowing null so that we can get use from the monitoring now (even if some messages are null!).

chuwy:
Okay, makes sense. We always can add next schema without `null` after this fixed in the Elasticsearch Loader. Scheduling.

jbeemster:
Thanks @chuwy !

chuwy:
Closing in favor of https://github.com/snowplow/iglu-central/pull/860. Thanks.

chuwy:
Closing in favor of https://github.com/snowplow/iglu-central/pull/860. Thanks.

PR #860, opened 2018-11-26
R98

chuwy:


chuwy:
This one patches couple of JSON Paths, but I triple-checked that this is safe.

alexanderdean:
Done!

chuwy:
cc @jbeemster, @rzats

PR #877, opened 2018-12-11
R99

chuwy:


yalisassoon:
@alexanderdean - we need this published for a the forthcoming JS tracker release @mhadam has been working on. I understand you're the person who does the merging / releasing? Is that something that can be delegated?

BenFradet:
it'd be great if we could include #879 too :+1:

alexanderdean:
Sorry for the delay - let me know as soon as #897 is finalised.

Very eager to hand over Iglu Central releases once we have a bulletproof CI/CD process in place.

yalisassoon:
What's the status here? (It's gating the next JS tracker release.) Have asked @goodits to look into it

alexanderdean:
Either #897 needs to be finished, or pushed back. Not sure why it was added given the release was already code complete...

PR #879, opened 2018-12-27
Add com.snowplowanalytics.snowplow/recoveries/jsonschema/1-0-0

BenFradet:
I would need a fallback case so that if you want to run a custom not supported out of the box recovery scenario you don't need to update the schema, wdyt would be the best way to go about it?

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/recoveries/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow/recoveries/jsonschema/1-0-0
@@ -11,7 +11,7 @@
   "items": {
     "type": "object",
     "description": "Recovery scenario sum type",
-    "oneOf": [
+    "anyOf": [
       {
         "properties": {
           "name": {
--- a/schemas/com.snowplowanalytics.snowplow/recoveries/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow/recoveries/jsonschema/1-0-0
@@ -10,15 +10,18 @@
   "type": "array",
   "items": {
     "type": "object",
+    "description": "Recovery scenario sum type",
     "oneOf": [
       {
         "properties": {
           "name": {
+            "description": "The PassThrough recovery scenario doesn't modify the payload",
             "enum": [
               "PassThrough"
             ]
           },
           "error": {
+            "description": "Error filter used to discriminate the bad rows that will be affected by this recovery scenario",
             "type": "string"
           }
         },
@@ -31,15 +34,18 @@
       {
         "properties": {
           "name": {
+            "description": "The RemoveFrom{Body,QueryString} will remove part of the payload",
             "enum": [
               "RemoveFromQueryString",
               "RemoveFromBody"
             ]
           },
           "error": {
+            "description": "Error filter used to discriminate the bad rows that will be affected by this recovery scenario",
             "type": "string"
           },
           "toRemove": {
+            "description": "String or regex to remove from the {body, query string}",
             "type": "string"
           }
         },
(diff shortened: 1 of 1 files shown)


schemas/com.snowplowanalytics.snowplow/recoveries/jsonschema/1-0-0
```diff
@@ -0,0 +1,145 @@
...
+    "anyOf": [
+      {
+        "properties": {
+          "name": {
+            "description": "The PassThrough recovery scenario doesn't modify the payload",
+            "enum": [
+              "PassThrough"
+            ]
+          },
+          "error": {
```
chuwy: Should we mention format of `error`s in the description?

schemas/com.snowplowanalytics.snowplow/recoveries/jsonschema/1-0-0
```diff
@@ -0,0 +1,145 @@
...
+        "required": [
+          "name",
+          "error",
+          "toReplace",
+          "replacement"
+        ],
+        "additionalProperties": false
+      },
+      {
+        "properties": {
```
chuwy: I'd still prefer an option with self-describing JSON as a fallback. `additionalProperties` makes this fallback effectively untyped.

BenFradet: The point is not to have to upload a schema to run a custom recovery scenario, I think having a self-desc fallback defeats this purpose, wdyt?

schemas/com.snowplowanalytics.snowplow/recoveries/jsonschema/1-0-0
BenFradet: wdym by format?

schemas/com.snowplowanalytics.snowplow/recoveries/jsonschema/1-0-0
chuwy: At one hand - yes, definitely, but on the other hand if somebody wrote a custom recovery scenario - uploading a schema would be a least of their problems. At the same time schema'ed recovery can motivate one to publish whole recovery scenario.

Also, even if recovery will be written in Scala - JSON schema has stronger validation, so quite a lot invalid configs can be filtered out.

schemas/com.snowplowanalytics.snowplow/recoveries/jsonschema/1-0-0
chuwy: Sorry I though its going to contain complex logic such as `contains("invalid schema") && contains("com.acme")`, but apparently this is going to be simple matcher.

schemas/com.snowplowanalytics.snowplow/recoveries/jsonschema/1-0-0
BenFradet: I'm not quite sure adding steps is going to encourage contributions but anyways the main problem I'm seeing with this approach is that I'm not going to have the same ADT in the `data` array, that seems like unnecessary complexity, wdyt?

alexanderdean:
Can we just move this back to R100 @BenFradet, it is blocking R99 release, cc @yalisassoon

BenFradet:
sure

alexanderdean:
Pushed back, unblocks R99

alexanderdean:
@BenFradet if you want to expedite this I'm sure the other stuff can be moved out of R100

BenFradet:
yes will do, I will just wait for Anton to get back

PR #882, opened 2019-01-09
Bad row schemas

BenFradet:
- [ ] `payload.headers` data model
- [ ] `payload.body` data model, could be an object?
- [ ] `processor.platform` could be an enum?
- [ ] `payload.path` pattern?


schemas/com.snowplowanalytics.badrows/format_violation/jsonschema/1-0-0
```diff
@@ -0,0 +1,87 @@
...
+      "properties": {
+        "artifact": {
+          "type": "string",
+          "description": "Artifact responsible for the creation of this format violation",
+          "maxLength": 512
+        },
+        "version": {
+          "type": "string",
+          "description": "Version of the artifact responsible for the creation of this format violation",
+          "pattern": "^(\\d+\\.\\d+\\.\\d+.*)$",
```
miike: Do we need to add the option here to validate release candidates as well?

schemas/com.snowplowanalytics.badrows/schema_violations/jsonschema/1-0-0
```diff
@@ -0,0 +1,167 @@
...
+                  "igluRepositories": {
+                    "type": "array",
+                    "description": "Iglu repositories tried or where the offending schema was found",
+                    "items": {
+                      "type": "string",
+                      "description": "Name of the Iglu repository",
+                      "maxLength": 128
+                    }
+                  },
+                  "property": {
```
miike: Is this equivalent to pointer?

schemas/com.snowplowanalytics.badrows/format_violation/jsonschema/1-0-0
BenFradet: it's covered by the `.*` at the end :+1:

schemas/com.snowplowanalytics.badrows/schema_violations/jsonschema/1-0-0
BenFradet: no this would be the violated property, e.g. `maxLength` but this schema is really dependent on what is going to end up being in the new iglu client

schemas/com.snowplowanalytics.badrows/enrichment_failures/jsonschema/1-0-0
```diff
@@ -0,0 +1,115 @@
...
+      "properties": {
+        "timestamp": {
+          "type": "string",
+          "description": "Timestamp at which the payload was collected",
+          "format": "date-time"
+        },
+        "headers": {
+          "type": "object",
+          "description": "List of the headers part of this collector payload"
+        },
```
BenFradet: what would be a good data model here? array of key-value pairs?

schemas/com.snowplowanalytics.badrows/format_violation/jsonschema/1-0-0
```diff
@@ -0,0 +1,87 @@
...
+          "type": "string",
+          "description": "Path requested by this collector payload",
+          "maxLength": 256
+        },
+        "queryString": {
+          "type": [ "string", "null" ],
+          "description": "Query string associated with this collector payload"
+        },
+        "body": {
+          "type": [ "string", "null" ],
```
BenFradet: could we have the body as an object?

schemas/com.snowplowanalytics.badrows/format_violation/jsonschema/1-0-0
```diff
@@ -0,0 +1,87 @@
...
+        "version": {
+          "type": "string",
+          "description": "Version of the artifact responsible for the creation of this format violation",
+          "pattern": "^(\\d+\\.\\d+\\.\\d+.*)$",
+          "maxLength": 32
+        },
+        "platform": {
+          "type": "string",
+          "description": "Platform on which the artifact was running when the format violation occurred",
+          "maxLength": 64
```
BenFradet: I think we need to make that an enum, wdyt?

schemas/com.snowplowanalytics.badrows/format_violation/jsonschema/1-0-0
```diff
@@ -0,0 +1,87 @@
...
+          "maxLength": 256
+        }
+      },
+      "required": [ "timestamp", "message" ],
+      "additionalProperties": false
+    },
+    "payload": {
+      "type": "object",
+      "description": "The collector payload that resulted in a format violation",
+      "properties": {
```
BenFradet: should we add more top-level properties, like the ones in https://github.com/snowplow/snowplow/blob/master/2-collectors/thrift-schemas/collector-payload-1/src/main/thrift/collector-payload.thrift ?

schemas/com.snowplowanalytics.badrows/tracker_protocol_violations/jsonschema/1-0-0
```diff
@@ -0,0 +1,104 @@
...
+                "type": "string",
+                "description": "Tracker protocol violation",
+                "maxLength": 128
+              }
+            },
+            "required": [ "key", "value" ],
+            "additionalProperties": false
+          }
+        }
+      },
```
BenFradet: is this a good data model?

schemas/com.snowplowanalytics.badrows/schema_violations/jsonschema/1-0-0
```diff
@@ -0,0 +1,167 @@
...
+              },
+              {
+                "properties": {
+                  "etc": {
+                    "type": "string"
+                  }
+                },
+                "required": [ "etc" ],
+                "additionalProperties": false
+              }
```
BenFradet: this adt will depend on the iglu client work

schemas/com.snowplowanalytics.badrows/enrichment_failures/jsonschema/1-0-0
```diff
@@ -0,0 +1,115 @@
...
+              },
+              {
+                "properties": {
+                  "etc": {
+                    "type": "string"
+                  }
+                },
+                "required": [ "etc" ],
+                "additionalProperties": false
+              }
```
BenFradet: we need to look into the enrichments in order to make this ADT either more extensive or restrictive.

schemas/com.snowplowanalytics.badrows/enrichment_failures/jsonschema/1-0-0
```diff
@@ -0,0 +1,115 @@
...
+            ]
+          }
+        }
+      }
+    },
+    "payload": {
```
BenFradet: As discussed with Anton, there are two options for payloads we incorporate in bad rows happening during schema validation and enrichment:

- HTTP payload, this might help during exploring since the payload schema is consistent for all bad row types (except storage)
- tracker protocol payload + collector metadata, this is better structured and reflect that Snowplow has "understood" the incoming payload, the burden then falls to the recovery process to reconstruct a collector payload from this data

schemas/com.snowplowanalytics.badrows/storage_loader_failure/jsonschema/1-0-0
```diff
@@ -0,0 +1,73 @@
...
+      "additionalProperties": false
+    },
+    "payload": {
+      "type": "object",
+      "description": "The collector payload that resulted in a storage loader failure",
+      "properties": {
+        "enrichedEvent": {
+          "type": "string",
+          "description": "Enriched event associated with this failure"
+        },
```
BenFradet: should we adopt a more extensive data model like https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/elasticsearch_enriched_event/jsonschema/2-0-0

schemas/com.snowplowanalytics.badrows/storage_loader_failure/jsonschema/1-0-0
```diff
@@ -0,0 +1,73 @@
...
+        "id": {
+          "type": "string",
+          "description": "Machine-readable unique identifier of the storage loader that produced this failure",
+          "format": "uuid"
+        }
```
BenFradet: should this rather belong in the failure or in the processor?

yalisassoon:
Might it be possible to surface `event_id` as a first class field in at least some of the schemas (e.g. from in a consistent way? It feels like a useful step to take when debugging bad rows would be to check how many of them have actually been loaded (e.g. because they have been reprocessed). This would be significantly easier if all the bad rows generated from the point in the pipeline where the `event_id` has been parsed out exposed that value in a consistent way.

schemas/com.snowplowanalytics.badrows/schema_violations/jsonschema/1-0-0
```diff
@@ -0,0 +1,167 @@
...
+          "format": "date-time"
+        },
+        "messages": {
+          "type": "array",
+          "description": "List of failure messages associated with the schema violations",
+          "items": {
+            "type": "object",
+            "description": "Individual schema violation",
+            "anyOf": [
+              {
```
yalisassoon: I'm trying to figure out what's going on here. (Again - apologies for the dumb questions.) Can I confirm I've understood it right:

* This is an array with one item for each violation / issue that emerged validating the schemas for a particular event
* Each array can be one of three types, where different data points are available for each type of error:
  - For the first type four data points are captured (`schemaKey`, `instance`, `igluRepositories` and `message`)
  - For the second type we have the most detailed data set with the following fields: `schemaKey`, `instance`, `igluRepositories`, `property`, `value`, `expected`)
  - For the third we only have a single data point, `etc`
* Presumably the first two types correspond to two types of violation, where for the second type the field name and value that caused the violation have been parsed out of the self describing JSON, but for the first type they have not
* And the third type is a catchall for failures where the output doesn't conform to the above to types? 

If that is correct (and it might well not be), some questions:

1. What exactly is `instance`? It is a jsonpath to the field in question? In the self describing JSON? Or the schema? Can we name this so it is less ambiguous?
2. How comes the first type of error has a `message` property but the second type doesn't?
3. Could we do something to make it easier to identify which of the three types each object in the array is (e.g. add an additional enum field to all three types) to make it easier to query the array. (So an analyst can write a query that says, if the item in the array is of the second type, fetch the field value...)
4. I find the name of the single field in the third item type, `etc`, a bit confusing. Is there some context or a convention there that I'm not aware of that makes this name much more self-explanatory than I realise?

schemas/com.snowplowanalytics.badrows/format_violation/jsonschema/1-0-0
yalisassoon: I think the more we can add the better

schemas/com.snowplowanalytics.badrows/format_violation/jsonschema/1-0-0
yalisassoon: Agree

schemas/com.snowplowanalytics.badrows/schema_violations/jsonschema/1-0-0
yalisassoon: Can you elaborate @BenFradet ?

schemas/com.snowplowanalytics.badrows/enrichment_failures/jsonschema/1-0-0
```diff
@@ -0,0 +1,115 @@
+{
```
yalisassoon: My hope was that for any enrichment failures we would have the ability to query all the fields in the event that were not generated at enrichment. Have I misunderstood? Or is that not the case?

schemas/com.snowplowanalytics.badrows/storage_loader_failure/jsonschema/1-0-0
yalisassoon: Yes - think so! Was your initial proposal that this be the enriched event tsv? 
I'd love to brainstorm what the format of this should be, and how that relates to the format of the data outputted at enrichment failure. (Would be nice if it was a comparable format, but lacking the fields that would be computed at enrichment.)

yalisassoon: Also - we want these failures to be very easy to recover i.e. to be able to write a simple process where our different loaders can quickly retry loading the bad data. Not sure how this requirement impacts the design decisions here about how the data is stored... (Maybe we attach two copies to the bad row - one that reflects the format of the data prior to loading to the specific storage target format, and another that is the "canonical" format?)

schemas/com.snowplowanalytics.badrows/schema_violations/jsonschema/1-0-0
BenFradet: it will depend on what information is going to be made available in the next iglu client release

schemas/com.snowplowanalytics.badrows/format_violation/jsonschema/1-0-0
```diff
@@ -0,0 +1,87 @@
+{
```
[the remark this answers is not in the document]
yalisassoon: Let's rename to `collector_payload_format_violation`

schemas/com.snowplowanalytics.badrows/tracker_protocol_violations/jsonschema/1-0-0
```diff
@@ -0,0 +1,104 @@
...
+        },
+        "messages": {
+          "type": "array",
+          "description": "List of failure messages associated with the tracker protocol violations",
+          "items": {
+            "type": "object",
+            "description": "Individual tracker protocol violation",
+            "properties": {
+              "key": {
+                "type": "string",
```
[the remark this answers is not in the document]
yalisassoon: Let's do the above but also add an additional field `value` that captures the actual value that caused the violation.

schemas/com.snowplowanalytics.badrows/enrichment_failures/jsonschema/1-0-0
yalisassoon: As discussed with @BenFradet:

For failures at the first three stages of the pipeline (up to Tracker Protocol Failure) the payload will be the raw HTTP payload.

For schema validation and enrichment errors the payload will be the tracker protocol + collector metadata.

For load failures the payload will be the enriched event.

schemas/com.snowplowanalytics.badrows/schema_violations/jsonschema/1-0-0
yalisassoon: Understood!

schemas/com.snowplowanalytics.badrows/enrichment_failures/jsonschema/1-0-0
yalisassoon: So to clarify - if the payload is the tracker protocol + collector metadata then querying the individual fields should be relatively straightforward.

schemas/com.snowplowanalytics.badrows/storage_loader_failure/jsonschema/1-0-0
yalisassoon: @BenFradet to check with @chuwy

schemas/com.snowplowanalytics.badrows/schema_violations/jsonschema/1-0-0
```diff
@@ -0,0 +1,167 @@
...
+                    "type": "string",
+                    "description": "Failure message associated with this schema violation",
+                    "maxLength": 256
+                  }
+                },
+                "required": [ "schemaKey", "instance", "igluRepositories", "message" ],
+                "additionalProperties": false
+              },
+              {
+                "properties": {
```
poberherr: can we add `app_id` here ? 
The main intend of this is:
- It's right now hard to determine who are the emitters putting bad data into your schema collection.
To be able to escalate to the appropriate tool / team, I think mainly interesting is the `app_id`
To clarify `event_name` which is referred here as `schemaKey` ?

poberherr:
Is this `PR` stale ?

BenFradet:
no, I'm just working on other things atm

schemas/com.snowplowanalytics.badrows/format_violation/jsonschema/1-0-0
chuwy: Is it going to be `[aws, gcp]`? What about mini? Is it platform-dependent?

schemas/com.snowplowanalytics.badrows/schema_violations/jsonschema/1-0-0
chuwy: Here's my attempt to formalize an error coming from Iglu client: https://gist.github.com/chuwy/f03247b2c399d134d428aa81863bdfc4, as the `schema_violation` is pretty much owned by Iglu client. 

Few moments to bear in mind:

* It has no Iglu URI (or `schemaKey`), which is however super-critical in a bad row itself. I belive it should be included in a top-level property defined by bad row and not returned by Iglu client (as it generally doesn't make sense for callee (client) to return the argument that already was passed to it)
* We don't have *much* control over `Invalid data` format, which is most important part of that schema.

schemas/com.snowplowanalytics.badrows/storage_loader_failure/jsonschema/1-0-0
chuwy: Unfortunately, there's an important corner case that we must be aware of. Despite the fact that data in 99% of cases was already enriched and we *can* assume the enriched format there's also a chance that something sends rubbish to PubSub/Kafka topic or S3 bucket and Loader simply cannot read it. I'm not sure whether this should be another member of union type

```
StorageLoaderFailure = AbsolutelyInvalidData | ProblemWithEnrichedData
```

Or maybe even totally separate type of bad row (quite undesirable), but we cannot **always** assume an enriched format and pretty much nothing apart from custom loader's message.

schemas/com.snowplowanalytics.badrows/storage_loader_failure/jsonschema/1-0-0
chuwy: Depends on our decision regarding "totally invalid data" problem I mentioned above.

schemas/com.snowplowanalytics.badrows/format_violation/jsonschema/1-0-0
BenFradet: yes it's platform-dependent

schemas/com.snowplowanalytics.badrows/enrichment_failures/jsonschema/1-0-0
```diff
@@ -0,0 +1,115 @@
...
+                    "description": "Failure message associated with this enrichment failure",
+                    "maxLength": 256
+                  }
+                },
+                "required": [ "enrichment", "message" ],
+                "additionalProperties": false
+              },
+              {
+                "properties": {
+                  "etc": {
```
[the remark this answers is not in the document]
BenFradet: yes, etc is to be defined/decided

schemas/com.snowplowanalytics.badrows/format_violation/jsonschema/1-0-0
BenFradet: I think rather than aws or gcp, we might want to specify the technology used, e.g. `[kinesis, pubsub, kafka, nsq, and one for batch, spark?]`, wdyt?

chuwy: I don't have a strong opinion on this one and just wanted to raise that for Mini AWS/GCP dichotomy would be irrelevant. At the same time isn't this information derivable from `processor`?

BenFradet: you mean `artifact`? yes it is, I think let's remove it

schemas/com.snowplowanalytics.badrows/storage_loader_failure/jsonschema/1-0-0
BenFradet: Since this one is completely unopinionated, we could have `Payload = EnrichedEvent | SomeOtherThing`

schemas/com.snowplowanalytics.badrows/format_violation/jsonschema/1-0-0
```diff
@@ -0,0 +1,87 @@
...
+      "properties": {
+        "timestamp": {
+          "type": "string",
+          "description": "Timestamp at which the payload was collected",
+          "format": "date-time"
+        },
+        "headers": {
+          "type": "object",
+          "description": "List of the headers part of this collector payload"
+        },
```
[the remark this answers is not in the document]
BenFradet: yup, chose not to

schemas/com.snowplowanalytics.badrows/schema_violations/jsonschema/1-0-0
BenFradet: yes there'll be [the full tracker protocol](https://github.com/snowplow/snowplow/wiki/snowplow-tracker-protocol)

BenFradet:
@yalisassoon @chuwy pushed new versions which are a lot more feature complete, feel free to review :+1:

schemas/com.snowplowanalytics.snowplow.badrows/schema_violations/jsonschema/1-0-0
```diff
@@ -0,0 +1,303 @@
...
+                }
+              ]
+            }
+          ]
+        }
+      },
+      "required": [ "timestamp", "schemaKey", "messages" ],
+      "additionalProperties": false
+    },
+    "payload": {
```
poberherr: I think that's  an interesting choice;
So if I understand correctly,  the whole payload of the event is inlined here, which makes sence, so the shredded events are still searchable; for me this is perfect since I'd be still able to search for the app_id and sort ideally errors by `app_id` , `event_name` etc.

miike: I think this is part of the payload (e.g., the core event but not contexts / derived_contexts).

BenFradet: yes that's right

poberherr:
Hmm then I have a general question, just to clarify if I missunderstood something:

Currently the bad row format schema is like this: ( or how is this schema called ? schredded event schema ? ) 

``````{
  "_index": "bad",
  "_type": "bad",
  "_id": "<id>",
  "_version": 1,
  "_score": null,
  "_source": {
    "line": "<base64 blob>",
    "errors": [
      {
        "level": "error",
        "message": "error: "
      }
    ],
    "failure_tstamp": "2019-03-28T13:10:27.019Z"
  },
  "fields": {
    "failure_tstamp": [
      "2019-03-28T13:10:27.019Z"
    ]
  },
  "sort": [
    1553778627019
  ]
}```

my main problem is that this is not searchable; 
So this will be replaced by `schemas/com.snowplowanalytics.snowplow.badrows/schema_violations/jsonschema/1-0-0` is that correct?

**Comment by BenFradet** [comment]
This will be replaced by one data fitting one of the schema defined in this pr according to the failure that happened, this is detailed in the rfc: https://discourse.snowplowanalytics.com/t/a-new-bad-row-format/2558

**Comment by poberherr** [comment]
Yes understood; 
But as I also mentioned in the RFC, from a user perspective, pulling in some properties form the main event to the `context` events would be extremely helpful. I can see that this will be probably a pain to implement but as pointed out [here](https://github.com/snowplow/iglu-central/pull/882/files#r269219576) pulling in the app_id would solve a lot of problems.
While the debugging with the suggested `schema_violation` is better, it would still be hard to figure out if multiple user apps / platforms use the same unstructured events producing errors.

**Comment by BenFradet** [comment]
The `app_id` will be part of a schema violation: https://github.com/snowplow/iglu-central/pull/882/files#diff-e552d9f744e44db7e869d0771da0b645R136

**Comment by poberherr** [comment]
:bowing_man: thanks for pointing it out, appreciated

File: schemas/com.snowplowanalytics.snowplow.badrows/schema_violations/jsonschema/1-0-0 [hunk 21]
```diff
@@ -0,0 +1,303 @@
...
+          "format": "date-time"
+        },
+        "schemaKey": {
+          "type": "string",
+          "description": "Schema which led to the issue",
+          "pattern": "^iglu:[a-zA-Z0-9-_.]+/[a-zA-Z0-9-_]+/[a-zA-Z0-9-_]+/[0-9]+-[0-9]+-[0-9]+$"
+        },
+        "messages": {
+          "description": "Schema violations",
+          "anyOf": [
```
yalisassoon: These schema violation error types are brilliant Ben: so specific - and I think, collectively, exhaustive.

schemas/com.snowplowanalytics.snowplow.badrows/schema_violations/jsonschema/1-0-0
```diff
@@ -0,0 +1,303 @@
...
+      },
+      "required": [ "timestamp", "schemaKey", "messages" ],
+      "additionalProperties": false
+    },
+    "payload": {
+      "type": "object",
+      "description": "The payload that resulted in schema violations",
+      "properties": {
+        "trackerProtocolPayload": {
+          "type": "object",
```
yalisassoon: Love this @BenFradet - being able to query the top level fields in the trackerProtocolPayload is going to make debugging _much_ easier

schemas/com.snowplowanalytics.snowplow.badrows/enrichment_failures/jsonschema/1-0-0
```diff
@@ -0,0 +1,309 @@
...
+          }
+        }
+      }
+    },
+    "payload": {
+      "type": "object",
+      "description": "The payload that resulted in enrichment failures",
+      "properties": {
+        "trackerProtocolPayload": {
+          "type": "object",
```
yalisassoon: Once an event is through to enrichment, validation has been completed. Wouldn't it then be possible to make the `unstruct_event` and `contexts` fields available for querying? There might be useful information about e.g. mobile app version that is buried in a mobile-specific context, for example.

schemas/com.snowplowanalytics.snowplow.badrows/storage_loader_failure/jsonschema/1-0-0
```diff
@@ -0,0 +1,573 @@
...
+            "data": {
+              "type": "object",
+              "description": "Data associated with the particular storage loader failure"
+            }
+          },
+          "required": [ "schema", "data" ],
+          "additionalProperties": false
+        },
+        "enrichedEvent": {
+          "type": "object",
```
yalisassoon: It looks like the `enrichedEvent` object is missing properties for `unstruct_event`, `contexts` and `derived_contexts` I think?

schemas/com.snowplowanalytics.snowplow.badrows/enrichment_failures/jsonschema/1-0-0
BenFradet: true, I think I missed those, will add them

schemas/com.snowplowanalytics.snowplow.badrows/storage_loader_failure/jsonschema/1-0-0
BenFradet: true, added

schemas/com.snowplowanalytics.snowplow.badrows/enrichment_failures/jsonschema/1-0-0
```diff
@@ -0,0 +1,311 @@
...
+                    "type": "string",
+                    "description": "Format of the enrichment which produced this failure",
+                    "pattern": "^[a-zA-Z0-9-_]+$"
+                  },
+                  "version": {
+                    "type": "string",
+                    "description": "Version of the enrichment which produced this failure",
+                    "pattern": "^[0-9]+-[0-9]+-[0-9]+$"
+                  },
+                  "identifier": {
```
chuwy: What we're going to put into `identified` until we don't a separate id for it?

schemas/com.snowplowanalytics.snowplow.badrows/enrichment_failures/jsonschema/1-0-0
```diff
@@ -0,0 +1,311 @@
...
+    },
+    "payload": {
+      "type": "object",
+      "description": "The payload that resulted in enrichment failures",
+      "properties": {
+        "trackerProtocolPayload": {
+          "type": "object",
+          "description": "The tracker protocol payload that resulted in enrichment failures",
+          "properties": {
+            "app_id": { "type": ["string", "null"] },
```
chuwy: These don't look like tracker protocol properties. More like `enrichedEvent` (though it is not technically enriched yet) and I'd expect 134 properties here, but see only about 80.

chuwy: In case this is still supposed to be tracker protocol - what happens if new tracker (e.g. with implemented 1-0-5 [payload_data](https://github.com/snowplow/iglu-central/tree/master/schemas/com.snowplowanalytics.snowplow/payload_data/jsonschema) with new `foo` property) sends an event to a legacy enricher (that is aware only of 1-0-4). Will enricher just strip away this `foo` in a bad row?

schemas/com.snowplowanalytics.snowplow.badrows/enrichment_failures/jsonschema/1-0-0
```diff
@@ -0,0 +1,311 @@
...
+            "derived_tstamp": { "type": ["string", "null"] },
+            "event_vendor": { "type": ["string", "null"] },
+            "event_name": { "type": ["string", "null"] },
+            "event_format": { "type": ["string", "null"] },
+            "event_version": { "type": ["string", "null"] },
+            "true_tstamp": {
+              "type": ["string", "null"],
+              "format": "date-time"
+            },
+            "unstruct_event": { "type": ["object", "null"] },
```
chuwy: Object with `schema` and `data` properties (same for contexts)

schemas/com.snowplowanalytics.snowplow.badrows/enrichment_failures/jsonschema/1-0-0
BenFradet: I don't know yet this is something to figure out or make not required

schemas/com.snowplowanalytics.snowplow.badrows/enrichment_failures/jsonschema/1-0-0
BenFradet: yes they're taken from the tracker protocol payload transformation that's why there are 80 and not more

BenFradet: no enricher just won't care

schemas/com.snowplowanalytics.snowplow.badrows/enrichment_failures/jsonschema/1-0-0
chuwy: I'd go with duplicated `$vendor/$name/jsonschema/$version` string until we have proper id. Because making the field required means it will be a new MODEL.

schemas/com.snowplowanalytics.snowplow.badrows/enrichment_failures/jsonschema/1-0-0
chuwy: Not sure I following. It means we'll not have it them in a bad row?

chuwy: Can we maybe find more descriptive name then?

I actually think raw tracker protocol probably would be a better fit because it represents raw/original/untouched event. But maybe it will also be harder to query for analysts.

schemas/com.snowplowanalytics.snowplow.badrows/enrichment_failures/jsonschema/1-0-0
```diff
@@ -0,0 +1,311 @@
...
+                    "required": [ "error" ],
+                    "additionalProperties": false
+                  }
+                ]
+              }
+            },
+            "required": [ "enrichment", "message" ],
+            "additionalProperties": false
+          }
+        }
```
BenFradet: missing:

```json
            "required": [ "timestamp", "messages" ],
            "additionalProperties": false
```

BenFradet:
Updated the schemas, here's what changed:

- rm storage loader failure => better left to Anton
- collector payload format violation:
  - payload is a simple string, we can't assume structure regarding what's in the incoming stream
  - more detailed failure structure
- enrichment failures:
  - payload is a partially enriched event
- schema violations:
  - added cases to the messages (not json, not sd, schema criterion)
  - payload is a partially enriched event
- tracker protocol violation:
  - small modifications to the payload to fit the implem
  - more detailed failures reusing iglu's client error
- add adapter failure

schemas/com.snowplowanalytics.snowplow.badrows/adapter_failures/jsonschema/1-0-0
```diff
@@ -0,0 +1,244 @@
...
+                  "value": {
+                    "type": [ "string", "null" ],
+                    "description": "Stringified representation of the value which is not json"
+                  },
+                  "error": {
+                    "type": "string",
+                    "description": "Json parsing issue"
+                  }
+                },
+                "required": ["properties", "error" ],
```
benjben: There should not be `properties`.

schemas/com.snowplowanalytics.snowplow.badrows/tracker_protocol_violations/jsonschema/1-0-0
```diff
@@ -0,0 +1,427 @@
...
+                  "value": {
+                    "type": [ "string", "null" ],
+                    "description": "Stringified representation of the value which is not json"
+                  },
+                  "error": {
+                    "type": "string",
+                    "description": "Json parsing issue"
+                  }
+                },
+                "required": ["properties", "error" ],
```
benjben: There should not be `"properties"`.

schemas/com.snowplowanalytics.snowplow.badrows/schema_violations/jsonschema/1-0-0
```diff
@@ -0,0 +1,437 @@
...
+                  "value": {
+                    "type": [ "string", "null" ],
+                    "description": "Stringified representation of the value which is not json"
+                  },
+                  "error": {
+                    "type": "string",
+                    "description": "Json parsing issue"
+                  }
+                },
+                "required": ["properties", "error" ],
```
benjben: There should not be `properties`.

PR #905, opened 2019-02-28
Add marketing_campaign_* fields to com.sendgrid schemas

### LINKED ISSUE #904: Add marketing_campaign_* fields to com.sendgrid schemas

dilyand:


dilyand:
@BenFradet Do you mean multiple commits within the same PR or separate PRs?

I'm also assuming that each commit will be closing a separate issue rather than have one for all?

Regarding the version, I think it should be at least `1-1-0` but not a simple addition. We're adding fields but we have `additionalProperties: true` in `1-0-0`, which means that some of these properties might have been part of historical events but with a different format. Eg, something that is a string now might have been an integer. We cannot guarantee that all old events will pass validation against these new schemas.

I went for `2-0-0` rather than `1-1-0` because the Redshift DDLs have meanwhile also changed. `1-0-0` DDLs still have the `LZO` encoding while the new DDLs have `ZSTD`.

BenFradet:
yes I meant mutlitple commits in the same PR.

@chuwy can help us regarding your other qs

chuwy:
Hey @dilyand, I agree with `2-0-0` approach. The main disadvantage is a new table, but it is a lot safer from my POV. Redshift's compression type though is irrelevant here as it doesn't break anything and we can even patch such DDLs.

schemas/com.sendgrid/processed/jsonschema/2-0-0
```diff
@@ -0,0 +1,88 @@
...
+    },
+    "send_at": {
+      "description": "To schedule a send request for a large batch of emails, use the send_at parameter which will send all emails at approximately the same time. send_at is a UNIX timestamp.",
+      "type": "integer",
+      "maximum": 2147483647,
+      "minimum": 0
+    },
+    "marketing_campaign_id": {
+      "description": "For emails sent through our Marketing Campaigns feature, we add Marketing Campaigns specific parameters to the Event Webhook. Both marketing_campaign_name and marketing_campaign_id are displayed as unique arguments in the event data.",
+      "type": "integer"
```
benjben: The `marketing_campaign_*` fields are always set, they can't be `null`?

benjben:
Awesome that you added the descriptions of the fields @dilyand !

schemas/com.sendgrid/processed/jsonschema/2-0-0
dilyand: The SendGrid [docs](https://sendgrid.com/docs/for-developers/tracking-events/event/#custom-arguments) say:

> Both `marketing_campaign_name` and  `marketing_campaign_id` are displayed as unique arguments in the event data.

I took that to mean that they are always set.

benjben: Ok, I asked because the issue [here](https://github.com/snowplow/snowplow/issues/4002) mentions *optional* `marketing_campaign_*` fields.

What about `marketing_campaign_version` and `marketing_campaign_split_id` ?

dilyand: Ah, yes, they are optional. That is why the schema does not have any required properties. So if they are missing, their values will indeed be `null`.

benjben: In my understanding `required` and `null` were 2 different things:
- `required`: if a field is not required, it's possible that it doesn't appear at all in the JSON.
- `null`: if `null` is not in the type (e.g. `"type": ["null", "integer"]), the value of the field can never be `null`.

Isn't it the case?

dilyand: You've got it right. If it's not required, then it can be missing from the JSON and the event will still pass validation.

If it is in the JSON, then it must match the type, which in this case is `integer`.

I meant that if it's missing, it will pass validation, but the value of the field will be `null` downstream. For example when loading into Redshift this column is nullable: https://github.com/snowplow/iglu-central/pull/905/files#diff-b05f60993433e51ea2a62f00526a47f3R20 .

Does that make sense?

benjben: Ok so in the JSON the field will not exist unless it has a non-null value, is that right?

dilyand: Yes, that is my assumption, based on the SendGrid docs.

benjben: Ok cool thanks!

BenFradet:
superseded by #942

PR #920, opened 2019-03-07
Screen view updates

mhadam:
These are the required schema changes for the mobile screen tracking updates.


schemas/com.snowplowanalytics.mobile/screen/jsonschema/1-0-0
```diff
@@ -0,0 +1,41 @@
...
+            "type": "string",
+            "maxLength": 100,
+            "description": "The type of screen that was viewed e.g feed / carousel."
+        },
+        "screen_session_id": {
+            "type": "string",
+            "maxLength": 100,
+            "description": "An ID that can be used in lieu of the name property, usually regenerated upon each session."
+        },
+        "screenview_id": {
```
dilyand: Isn't that redundant, since there already is a unique `event_id` generated for each screenview?

dilyand:
Hey @mhadam , a lot of the fields in the `screen` context seem to duplicate fields from the `screen_view` event. Wouldn't it be better to have a schema for the event that only has event-related data, and then push everything else off into the context?

schemas/com.snowplowanalytics.mobile/screen/jsonschema/1-0-0
colmsnowplow: @dilyand screen view id is conceptually equivalent to page view id. It's both in the screen view event and attached as a context to all events on the screen (if that's enabled).

schemas/com.snowplowanalytics.snowplow/screen_view/jsonschema/1-0-1
```diff
@@ -0,0 +1,61 @@
...
+			"type": "string",
+            "maxLength": 100,
+            "description": "The name of the screen viewed."
+		},
+		"type": {
+            "type": "string",
+            "maxLength": 100,
+            "description": "The type of screen that was viewed e.g feed / carousel."
+        },
+        "screen_session_id": {
```
colmsnowplow: Perhaps there's a better name for this that avoids confusing it with the session id?

schemas/com.snowplowanalytics.mobile/screen/jsonschema/1-0-0
dilyand: That makes sense.

schemas/com.snowplowanalytics.snowplow/screen_view/jsonschema/1-0-1
architgoyal1: +1

colmsnowplow: screen_instance_id maybe? This is still a flawed descriptor though would be keen to come up with something more accurate

schemas/com.snowplowanalytics.mobile/screen/jsonschema/1-0-0
```diff
@@ -0,0 +1,41 @@
...
+        },
+        "type": {
+            "type": "string",
+            "maxLength": 100,
+            "description": "The type of screen that was viewed e.g feed / carousel."
+        },
+        "screen_session_id": {
+            "type": "string",
+            "maxLength": 100,
+            "description": "An ID that can be used in lieu of the name property, usually regenerated upon each session."
```
architgoyal1: Why would the screen ID be regenerated each session? I would assume there is a 1:1 mapping of screen IDs to screen name and screen name would persist past the session?

schemas/com.snowplowanalytics.snowplow/screen_view/jsonschema/1-0-1
architgoyal1: I am a little confused with this field in general:

Why would the screen ID be regenerated each session? I would assume there is a 1:1 mapping of screen IDs to screen name and screen name would persist past the session?

If this is an ID that is generated each session then we could just use the session_id out of the client_session_context?

colmsnowplow: There are two things here - one is a screen ID which is like a page view id - new UUID for every screen view event.

This one is to account for apps which don't do a good job of naming their screens. It gives a screen ID to the screen which doesn't change (per session. There's no way to retain the ID after the app closes). So let's say you haven't assigned screen names, and the user journey is:

Screen 1 [set of events a] -> screen 2 -> screen 1 [set of events b]

This field allows you to aggregate across all events which occurred on screen 1 in the session, regardless of how many times the user left and came back to the screen within that session.

That's how I understand it anyway.

schemas/com.snowplowanalytics.snowplow/screen_view/jsonschema/1-0-1
```diff
@@ -0,0 +1,61 @@
...
+            "type": "string",
+            "maxLength": 100,
+            "description": "A screenview ID of the previous screenview."
+        },
+        "previous_type": {
+            "type": "string",
+            "maxLength": 100,
+            "description": "The screen type of the previous screenview."
+        },
+        "transition_type": {
```
architgoyal1: (Piggybacking off @dilyand 's comment) In my view this is the only property that is really specific to the screen view event, all the other properties can be in the screen entity that we will recommend that all mobile users enable.

Just for my understanding, is this field intended for values such as "button_click" and "swipe"? ie the action that was performed to lead us to this screen?

schemas/com.snowplowanalytics.snowplow/screen_view/jsonschema/1-0-1
colmsnowplow: So actually this should be screen_id but that's confusing too. 

screen_name_id? aux_screen_name maybe?

architgoyal1: > There's no way to retain the ID after the app closes

Ah, this field makes sense to me then!

Perhaps a name to imply that it is transient such as `screen_temp_id`

dilyand: I think `screen_name_id` makes most sense to me. It's an ID that's meant to be used instead of a name.

schemas/com.snowplowanalytics.snowplow/screen_view/jsonschema/1-0-1
dilyand: I think everything from `screenview_id` till the end belongs in the event rather than the screen entity. 'Previous' suggests some sort of temporal relationship. That makes sense for events, which follow in chronological order; but not so much for screens, which are not temporally bound.

schemas/com.snowplowanalytics.snowplow/screen_view/jsonschema/1-0-1
colmsnowplow: Yeah my concern on that one is that it suggests that it's a 1:1 map to name, which it isn't. It's a 1:1 map to name + session_id.

`persisting_screen_id` is another option but shares this problem.

schemas/com.snowplowanalytics.snowplow/screen_view/jsonschema/1-0-1
colmsnowplow: > Just for my understanding, is this field intended for values such as "button_click" and "swipe"? ie the action that was performed to lead us to this screen?

I _think_ transition_type refers to an actual technical aspect of apps. There are different ways of transitioning from one screen to another, and this captures that as far as I know. I'm not sure if they're predicated on the user action like 'button click' or 'swipe', or if they're something like 'View Controller view loaded' vs 'Carousel screen transition' (I made those up but just trying to illustrate)

schemas/com.snowplowanalytics.snowplow/screen_view/jsonschema/1-0-1
architgoyal1: I think we should be careful to show that it _doesn't_ persist. I'm trying to imagine a world where a user didn't do a good job of naming screens and relied on `screen_xxxx_id` thinking it will be a persisting ID. I think we need to make clear that it doesn't really replace screen name as it _wont_ persist.

colmsnowplow: Agreed. The tricky part is balancing that with the fact that it does persist in comparison to the screen_id.

It persists for the session, it doesn't persist forever....

What about either `screen_in_session_id` or `in_session_screen_id`?

dilyand: `in_session_screen_id` gets my +1

architgoyal1: likewise +1

schemas/com.snowplowanalytics.snowplow/screen_view/jsonschema/1-0-1
architgoyal1: Ah ok makes sense!

mhadam:
> Hey @mhadam , a lot of the fields in the `screen` context seem to duplicate fields from the `screen_view` event. Wouldn't it be better to have a schema for the event that only has event-related data, and then push everything else off into the context?

@dilyand The reasoning here deals with performing queries from what I can tell - not super familiar here but that's what I understood. You could aggregate events more easily based on contexts and access any value. It's definitely overkill but perhaps worth it if it simplifies working with the data.

schemas/com.snowplowanalytics.snowplow/screen_view/jsonschema/1-0-1
mhadam: The motivation here is true autotracking where an app designer doesn't even have to be careful around how their screens are named. So ideally they'd name everything well but I thought, if I'm dealing with all these view controllers, I can round them up when the app starts and assign an ID from the start. But the problem is that these won't be the same across users and sessions since they're randomly generated.

I've tried to figure if there's any problem with the way I'm generating and associating the (currently called) screen_session_id, and I think perhaps the best name is `screen_instance_id`, and perhaps more accurately `viewcontroller_instance_id`. I'm using associated objects in Obj-C, where you're able to attach an object to another object. So there no way to really associate an id in a definite way across all devices, basically the best I can do is an id that persists as long as an object exists.

It's conceivable that a class could get instantiated and then deallocated, so the `instance_id` would change there. Maybe another field for the class name of the viewcontroller would be appropriate?

I think these schemas should be general, but perhaps there should be platform specific names like `viewcontroller_instance_id`, and `viewcontroller_class_name`.

schemas/com.snowplowanalytics.snowplow/screen_view/jsonschema/1-0-1
mhadam: Don't have much to add here, but figured it'd be good to say I don't have any autotracking for transition type. I'm not completely sure how to accomplish that, and it'd be something that could be added down the line if there's a decent way to go about it.

mhadam: > I think everything from `screenview_id` till the end belongs in the event rather than the screen entity. 'Previous' suggests some sort of temporal relationship. That makes sense for events, which follow in chronological order; but not so much for screens, which are not temporally bound.

Ok - but like I mentioned in another comment, is there any point where the querying is super simplified if you're able to discriminate events by screen details and the previous screen specifics are available right there? I think the intention is making screen transitions really accessible even if it means some redundancy.

dilyand: @mhadam I agree that some tradeoffs would be acceptable if they make working with the data easier. But I think there are no great benefits from redundancy in this case. On BigQuery or Snowflake, we only have a single `atomic.events` table anyway, so we're just duplicating columns in that table. On Redshift, if you use both schemas, you'll still end up with two additional shredded tables, and even if you can maybe do some queries with one join instead of two, I don't think it's worth it. I'm much in favour of keeping things separate and consistent with the concept of events vs entities that we're trying to push in our most recent training materials.

But we should take more opinions on this. @colmsnowplow , @architgoyal1 any definitive thoughts? @chuwy you might want to chime in as well?

colmsnowplow: My two cents: firstly, the context is optional. So IMO everything should remain in the event regardless.

> But I think there are no great benefits from redundancy in this case. On BigQuery or Snowflake, we only have a single atomic.events table anyway, so we're just duplicating columns in that table.  On Redshift, if you use both schemas, you'll still end up with two additional shredded tables, and even if you can maybe do some queries with one join instead of two, I don't think it's worth it.

I see the reasoning here, but I think it's worth highlighting a slightly different perspective - thinking of the benefit less in terms of query efficiency and more in terms of path to value for the user. 

If the data I want to aggregate by is only in one row, I need to account for that before I can gain insights. So I need a 1:many join, a window function, or I need to aggregate then join then aggregate again. If I wasn't proficient at SQL the way you and I are @dilyand, that's not trivial. The simplest option (1:many join) opens the door to problems with duplication, for example. So at the simplest we're asking them to get their head around how to test for duplication in the first step of their query before they get to learning about users.

If that data is in the context though, it's pretty straightforward. I have a 1:1 join then a group by. Hard to get wrong and I'll be confident in the insights I deliver within a few minutes of getting data. 

So IMO the question of what should go in the context should be answered by focusing on what the user is likely going to want to do - I don't think `transition_type` is a likely candidate for aggregations but I do think the rest are.

alexanderdean:
This is a great thread!

schemas/com.snowplowanalytics.snowplow/screen_view/jsonschema/1-0-1
mhadam: I'm not really sure if this is better handled by documenting it really clearly so the behind-the-scenes stuff is clear or whether it's better to offer more schema fields. Wondering what y'all think of supplemental platform specific fields. IMO, they offer a clearer idea of what's actually being captured for auto-tracking:

- For Android:
  - `activity_id`
  - `activity_name`
  - `fragment_id`
  - `fragment_name`

- For iOS:
  - `viewcontroller_id`
  - `viewcontroller_class_name`
  - `viewcontroller_name`

And (as we already have in the schema draft) required fields that during auto-tracking arbitrarily take info from the above fields:
  - `screen_id`
  - `name`

colmsnowplow: I think this likely confuses the issue for most consumers of the data. Think product manager rather than developer.

mhadam: ok agreed, it seems like tracking screenviews through activities is standard on android anyway

mhadam:
Last details I want to close up:

- `id` : should we make this 36 character length? (that's the size of the UUID generated on both Android/iOS)

- the rest: longer than 100? 128? Not sure if there's any rationale to follow around lengths.

schemas/com.snowplowanalytics.snowplow/screen_view/jsonschema/1-0-1
dilyand: On closer inspection, all of the context fields are already included in the event. What is the purpose of the context then?

dilyand:
@mhadam I think we should enforce lengths only when there is a very good reason for it. With the UUID there is a good reason -- it's effectively a double-check that weird strings don't get in. For the rest, my experience has been that it's better to be a bit permissive rather than end up with a recovery job on your hands, because something was a character too long.

schemas/com.snowplowanalytics.snowplow/screen_view/jsonschema/1-0-1
colmsnowplow: The event marks a screen view event. The context is attached to all events that happen on screen, like the web page context.

dilyand: In that case why don't we make the same as the web page context? All we need is the `screenview_id`.

colmsnowplow: > If the data I want to aggregate by is only in one row, I need to account for that before I can gain insights. So I need a 1:many join, a window function, or I need to aggregate then join then aggregate again.

If I want to aggregate the events that happen on a screen, then ideally the data I want to aggregate by isn't in only one row.

dilyand: No, I'm saying the context could literally be just the `screenview_id`, attached to every event on the screen, including the `screen_view` event. Just like the web page context works. The data will then be in all rows.

colmsnowplow: Sure. What if I need to aggregate non-screenview events across screen types, or all screen views of the same screen?

The web context works really well for a page views based model, but has these same issues for parallel tasks. That just doesn't come up often since most web analytics starts with page views.

dilyand: Got it.

colmsnowplow: I mean, it's just my opinion here - my initial reaction is that it's valuable/makes life easier. 🙂

mhadam:
Hoping the current files I've committed are good, let me know if there's anything I should change.

dilyand:
LGTM.

PR #933, opened 2019-04-19
[Temporary] Opencage enrichment

benjben:
This PR is temporary and is created so that Toni can develop integration tests for OpenCage enrichment with the associated JSON schemas.


chuwy:
Moved it to R108 as R105 has been released already and I'm working on R106 atm.

PR #943, opened 2019-05-21
Fix focus_form events being rejected by enrichment

borismo:
# Background
I noticed a lot of the following errors in the Stream Enricher "bad" Kinesis Stream:
```
error: object instance has properties which are not allowed by the schema: ["type"]
    level: "error"
    schema: {"loadingURI":"#","pointer":""}
    instance: {"pointer":""}
    domain: "validation"
    keyword: "additionalProperties"
    unwanted: ["type"]
```
They happen mostly on `focus_form` events.
# Investigation
I took a look at [their schema](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/focus_form/jsonschema/1-0-0) and noticed they indeed do not accept an `type` property. However the schema does include an `elementType` property.
The [JavaScript Tracker](https://github.com/snowplow/snowplow-javascript-tracker/blob/master/core/lib/core.ts#L836) seems to send a `type` property for form events, but no `elementType`.
# Proposed solution
Therefore it seems to me that replacing the property in the schema would align with other form events and fix this problem.

Does that sound right?

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/focus_form/jsonschema/1-0-1
+++ b/schemas/com.snowplowanalytics.snowplow/focus_form/jsonschema/1-0-1
@@ -1,0 +1,47 @@
+{
+	"$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+	"description": "Schema for the focus of a form field",
+	"self": {
+		"vendor": "com.snowplowanalytics.snowplow",
+		"name": "focus_form",
+		"format": "jsonschema",
+		"version": "1-0-0"
+	},
+
+	"type": "object",
+	"properties": {
+		"formId": {
+			"type": "string",
+			"minLength": 0,
+			"maxLength": 64
+		},
+		"elementId": {
+			"type": "string",
+			"minLength": 0,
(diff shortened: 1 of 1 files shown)


bramelfrink:
Travis complains that the focus_form schema is not valid:
`1. Error: JSON Schema [iglu:com.snowplowanalytics.snowplow/focus_form/jsonschema/1-0-0] doesn't conform path [com.snowplowanalytics.snowplow/focus_form/jsonschema/1-0-1]`

If I understand correctly, you renamed `elementType` to `type` in the schema?

borismo:
> Travis complains that the focus_form schema is not valid

80c729f fixed it 🙂
> If I understand correctly, you renamed `elementType` to `type` in the schema?

Correct.

snowplowcla:

:memo: Please visit https://github.com/snowplow/snowplow/wiki/CLA to learn more and sign.

borismo:
> you'll need to sign a Contributor License Agreement

I signed it!

snowplowcla:
@borismo  has signed the Software Grant and Corporate Contributor License Agreement

GroverBI:
When will this be fixed?

PR #959, opened 2019-06-11
Storage Loader Failure Related Schemas

aldemirenes:


schemas/com.snowplowanalytics.snowplow.badrows/snowflake_bad_row/jsonschema/1-0-0
```diff
@@ -0,0 +1,38 @@
...
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "invalid_timestamp": {
+      "type": "object",
+      "description": "Information about the invalid timestamp error",
+      "properties": {
+        "timestamp_type": {
```
chuwy: I'd call this `timestamp` and another property `value`.

schemas/com.snowplowanalytics.snowplow.badrows/snowflake_bad_row/jsonschema/1-0-0
```diff
@@ -0,0 +1,38 @@
...
+  "description": "Schema for a Snowflake Loader specific bad row",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow.badrows",
+    "name": "snowflake_bad_row",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "invalid_timestamp": {
```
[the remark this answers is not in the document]
aldemirenes: I think we can convert `invalid_timestamp` to `invalid_timestamps` as an array, what do you think ?

schemas/com.snowplowanalytics.snowplow.badrows/snowflake_bad_row/jsonschema/1-0-0
```diff
@@ -0,0 +1,38 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for a Snowflake Loader specific bad row",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow.badrows",
+    "name": "snowflake_bad_row",
```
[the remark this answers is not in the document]
aldemirenes: Name should be only `snowflake_loader` then ?

schemas/com.snowplowanalytics.snowplow.badrows/snowflake_bad_row/jsonschema/1-0-0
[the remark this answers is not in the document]
aldemirenes: I think we can add new property `max_raw_length_exceeded`. Do you think it is okay ?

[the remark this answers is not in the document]
chuwy: 👍

schemas/com.snowplowanalytics.snowplow.badrows/snowflake_bad_row/jsonschema/1-0-0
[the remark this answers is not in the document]
chuwy: Yep, or even `snowflake`. Up to you

schemas/com.snowplowanalytics.snowplow.badrows/snowflake/jsonschema/1-0-0
```diff
@@ -0,0 +1,566 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for a Snowflake Loader specific bad row",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow.badrows",
+    "name": "snowflake",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "anyOf": [
```
chuwy: I think this should be `oneOf`. `anyOf` means that both schemas could be applied, e.g. object could be with `original`, `errors` *and* `failures`, `enrichedEvent` properties, whereas we clearly want it to be only one of these sets.

schemas/com.snowplowanalytics.snowplow.badrows/snowflake/jsonschema/1-0-0
```diff
@@ -0,0 +1,566 @@
...
+    {
+      "type": "object",
+      "description": "Model for storage loader failures with unknown data",
```
chuwy: `...unkown data, e.g. invalid enriched event`

schemas/com.snowplowanalytics.snowplow.badrows/snowflake/jsonschema/1-0-0
```diff
@@ -0,0 +1,566 @@
...
+            "type": "string"
+          }
+        }
+      },
+      "required": ["original", "errors"],
+      "additionalProperties": false
+    },
+    {
+      "type": "object",
+      "description": "Model for storage loader failure with known data",
```
chuwy: `...i.e. valid enrcied event`

schemas/com.snowplowanalytics.snowplow.badrows/snowflake/jsonschema/1-0-0
```diff
@@ -0,0 +1,566 @@
...
+      "properties": {
+        "failures": {
+          "type": "array",
+          "description": "Snowflake related failures",
+          "items": {
+            "anyOf": [
```
chuwy: `oneOf`?

schemas/com.snowplowanalytics.snowplow.storage/snowflake_config/jsonschema/2-0-0
```diff
@@ -0,0 +1,120 @@
...
+            "format": "uri"
+        },
+        "stage": {
+            "type": "string"
+        },
+        "stageUrl": {
+            "type": "string",
+            "format": "uri"
+        },
+        "badOutputUrl": {
```
chuwy: I think this should be required. If we're going to make the Transformer work on with FS, do you think it would validate `file:///some/path`-like URIs?

schemas/com.snowplowanalytics.snowplow.storage/snowflake_config/jsonschema/2-0-0
```diff
@@ -0,0 +1,120 @@
...
+    "description": "Snowplow Snowflake storage configuration",
+    "self": {
+        "vendor": "com.snowplowanalytics.snowplow.storage",
+        "name": "snowflake_config",
+        "format": "jsonschema",
+        "version": "2-0-0"
+    },
+    "type": "object",
+    "properties": {
+        "name": {
```
chuwy: Can we add human-readable descriptions please? I wasn't pedantic enough when designing first versions.

schemas/com.snowplowanalytics.snowplow.storage/snowflake_config/jsonschema/2-0-0
aldemirenes: Yeah, I think so. I will try to look at the FS support in the following days whether adding it would be easy or not. I can be more precise about this after looking at it.

yalisassoon:
Looks great @aldemirenes ! A few suggestions from me:

1. Can we have two separate schemas for the two error types please? They are so different that it will be much easier to treat them as different types, partition them separately and query them separately.
2. Should we add a processor object to both schemas, with the following two data points, indicating where the bad row was generated? This should make these bad row types consistent with those that [Ben defined](https://github.com/snowplow/iglu-central/tree/bad-rows/schemas/com.snowplowanalytics.snowplow.badrows)
  - artifact
  - version
3. Can we rename the "original" object to "payload". I _think_ this is more consistent with the naming in the other fields, and is clearer in terms of an analyst understanding that this was the payload that the loader tried to work with. Not sure if @chuwy agrees?

Thanks!

schemas/com.snowplowanalytics.snowplow.badrows/loader_failure_with_unkown_data/jsonschema/1-0-0
```diff
@@ -0,0 +1,45 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for storage loader failures with unknown data, e.g. invalid enriched event",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow.badrows",
+    "name": "loader_failure_with_unknown_data",
```
chuwy: Can we come up with a more concise name? At least `enriched_parsing_error` or `loader_parsing_error`, 3 underscores in an entity name is my personal upper bound limit.

schemas/com.snowplowanalytics.snowplow.badrows/snowflake_specific_failure/jsonschema/1-0-0
```diff
@@ -0,0 +1,561 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for a Snowflake Loader specific bad row",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow.badrows",
+    "name": "snowflake_specific_failure",
```
chuwy: If we remove `specific` part, will `snowflake_failure` mean that failure is not specific to snowflake? Also, the difference between "failure" and "error" is subtle, but this one is certainly not "failure".

schemas/com.snowplowanalytics.snowplow.badrows/loader_failure_with_unkown_data/jsonschema/1-0-0
aldemirenes: I think we can go with `enriched_parsing_error`

schemas/com.snowplowanalytics.snowplow.badrows/snowflake_specific_failure/jsonschema/1-0-0
aldemirenes: You are right, `specific` part can be removed. Then, is `snowflake_error` okay ?

chuwy: Yes, fine by me.

schemas/com.snowplowanalytics.snowplow.badrows/loader_failure_with_unkown_data/jsonschema/1-0-0
chuwy: I think we can do better, but if not - let's stick with `enriched_parsing_error`

aldemirenes: After second thought, I think `enriched_parsing_error` could be confused with `enrichment_failures` which Ben is creating. I think going with `loader_parsing_error` would be better.

chuwy:
Closing in favor of https://github.com/snowplow/iglu-central/pull/962

PR #960, opened 2019-06-13
ST-27: Improve error handling in stream tee - schema

knservis:
Bad row format for relays (to be used with stream tee first)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+Release 106 (2019-07-03)
+------------------------
+Add com.snowplowanalytics.snowplow.badrows/relay_failure/jsonschema/1-0-0 (#974)
+
 Release 105 (2019-06-29)
 -----------------------
 Add com.snowplowanalytics.snowplow/gdpr/jsonschema/1-0-0 (#965)
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ limitations under the License.
 [travis]: https://travis-ci.org/snowplow/iglu-central
 [travis-image]: https://travis-ci.org/snowplow/iglu-central.png?branch=master
 
-[release-image]: http://img.shields.io/badge/release-105-orange.svg?style=flat
+[release-image]: http://img.shields.io/badge/release-106-orange.svg?style=flat
 [releases]: https://github.com/snowplow/iglu-central/releases
 
 [license-image]: http://img.shields.io/badge/license-Apache--2-blue.svg?style=flat
--- a/schemas/com.snowplowanalytics.snowplow.badrows/relay_failure/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow.badrows/relay_failure/jsonschema/1-0-0
@@ -1,0 +1,882 @@
+{
+    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "description": "Schema for a bad row resulting from a relay failure",
+    "self": {
+        "vendor": "com.snowplowanalytics.snowplow.badrows",
+        "name": "relay_failure",
+        "format": "jsonschema",
+        "version": "1-0-0"
+    },
+    "type": "object",
+    "properties": {
+        "failure": {
+            "type": "object",
+            "description": "Information regarding the relay failure",
+            "properties": {
+                "timestamp": {
+                    "type": "string",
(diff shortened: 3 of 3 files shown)


knservis:
Hey @yalisassoon thanks for the review:
> We cannot assume that all the events feeding Stream Tee will be valid Snowplow events: there is a possibliity for example that there are some random entries on the Stream that have been generated by another process. So we need a way to fail these events and just report the raw line of data, so that someone debugging can spot that those failures were related to the stream that Stream Tee has been set to consume to contain some non-Snowplow data. (This might be the case if e.g. the microservice was configured on the wrong stream e.g. the raw rather than the enriched stream.) I expect this would cause a different type of error, one with a very different payload object. (Which would be one string field instead.) 

That was the intention of the field "eventRaw" here https://github.com/snowplow/iglu-central/pull/960/files#diff-b5d26a61baa9ec035f259fddf579c2daR834
Which is supposed to be included if the "eventStruct" is not.

@chuwy and @aldemirenes is that how it was done in loaders?

> Ideally it would be nice to think about the different types of failure and what individual data points we can pull out for each to make debugging easier. However, this might only become clear after a few months of running Stream Tee, so maybe we can leave the failure object as-is, and then evolve it in future versions?

Other than the message text and the stack trace I currently can not imagine what other information to add..

knservis:
@yalisassoon Thanks for our chat today. So my understanding is that it may be better to use two schemas (essentially split the schema here into those that have the message encoded and those that don't).

chuwy:
In loaders, Scala Analytics SDK represents an ultimate source of thuth - everything that couldn't be parsed by it - goes into "parsing error bad row". SDK also provides an information about the failure (instead of traceback) - what field has an unexpected format. @aldemirenes submitted https://github.com/snowplow/iglu-central/pull/963/commits/d98308f377317b3a20ae03c514b647c3f956883b representing this generic parsing failure

knservis:
TL;DR @yalisassoon I would leave things as they are as I don't see the bad errors on loaders being sufficiently diverse in structure or use to justify multiple schemas. 

Here is my understanfing of what is happening over at loaders:
When parsing an enriched event the errors are an array of string for each payload:
https://github.com/snowplow/iglu-central/blob/5c12a462974b46e9c6327fe74faf355c0b656780/schemas/com.snowplowanalytics.snowplow.badrows/loader_parsing_error/jsonschema/1-0-0#L17-L20

When the event has been successfully parsed the errors that can occur have to do with some ci=olumn snowflake does nt like (presumably schema mismatch) for which you can get the coolumn, value and relevant message:
https://github.com/snowplow/iglu-central/blob/5c12a462974b46e9c6327fe74faf355c0b656780/schemas/com.snowplowanalytics.snowplow.badrows/snowflake_error/jsonschema/1-0-0#L23-L27

and a deduplication error:
https://github.com/snowplow/iglu-central/blob/5c12a462974b46e9c6327fe74faf355c0b656780/schemas/com.snowplowanalytics.snowplow.badrows/snowflake_error/jsonschema/1-0-0#L39-L45
with as simple "message"

in addition, in both cases you get all the event fields (as it has been parsed). I am assuming that this can help with reprocessing. 

As these errors are known and on distinct stages so adding all those in a large error schema could become confusing. 

In the case of stream tee we may parse or not (based on config) and we may get an error that has stopped parsing or transmission, but we are not reprocessing using the bad stream. The error has to do with an individual event that out of a batch had an issue and should not be reprocessed (e.g. size rules violation or invalid structure). 

To be honest even if the message was successfully parsed I would suggest not having the fields of the event but just the undecoded event as I currently cannot see any other use other than debugging and alerting for those. 

I would add the analytics sdk version in the artefact name possibly. 

@aldemirenes and @chuwy let me know if you think otherwise.

yalisassoon:
Understood @knservis - let's proceed with your suggeseted schema.

alexanderdean:
Hey @knservis - I took a look as requested but I don't really have the context to take a design view on this. Happy to be readded as "reviewer" when it's ready for release.

knservis:
@alexanderdean I think we are ready for release

alexanderdean:
Please get ticks from @yalisassoon and @chuwy (they are on as reviewers) then @ me. You are missing various things to cut a new Iglu Central release (CHANGELOG etc).

knservis:
@alexanderdean How does it look now?

alexanderdean:
LGTM!

PR #966, opened 2019-06-29
Mailchimp profile updates will contain undocumented ip_signup param

razius:
Payload data for profile updates will include an undocumented ip_signup
param which will make the event enriching fail due to validation errors.


snowplowcla:
@razius  has signed the Software Grant and Corporate Contributor License Agreement

PR #972, opened 2019-07-01
Add schemas for an application's self-reported status

dilyand:


dilyand:
An initial set of schemas meant to cover the data emitted from an application self-reporting its status. The set consists of a minimalistic `status_report` event and contexts that add details about the client, pipeline, application and the application status.

Key questions:
- Is this a good MVP set of schemas for implementing status reports from within applications such as the BigQuery Loader?
- Do the fields of each schema make sense?
- Any suggestions for fields that should be required? (The schemas are very loose at the moment.)
- Any suggestions for field limitations, such as `min`, `max`, etc?
- Any concerns that some of the data will not be possible to be reported from within the application?

schemas/com.snowplowanalytics.monitor/application/jsonschema/1-0-0
```diff
@@ -0,0 +1,112 @@
...
+      "description": "The ID of the application instance; a UUID",
+      "type": "string",
+      "maxLength": 36
+    },
+    "instance_name": {
+      "description": "The name of the application instance, eg com-acme-prod1-rdb-loader-1",
+      "type": ["string", "null"],
+      "maxLength": 256
+    },
+    "inputs": {
```
yalisassoon: Hey @dilyand - when I discussed this with @jbeemster we realised that applications would only have one input. (I.e. an app will only consume a single stream.) However, it could have many outputs.

For that reason we thought have dedicated `input_id` and `input_type` string fields in teh schema, but have an `outputs` field that was an array of objects where each object contains an `id` field (UUID)  and `type` field (string)

yalisassoon: Looking further down the schema that's exactly how you've specified the `outputs` field

schemas/com.snowplowanalytics.monitor/application_status/jsonschema/1-0-0
```diff
@@ -0,0 +1,133 @@
...
+      "description": "The start of the period to which this status applies",
+      "type": ["string", "null"],
+      "format": "date-time"
+    },
+    "periodEnd": {
+      "description": "The end of the period to which this status applies",
+      "type": ["string", "null"],
+      "format": "date-time"
+    },
+    "inAppStats": {
```
yalisassoon: When we developed the schema we distinguished between data points that would be emitted from the app with data points that could be generated by monitoring tools outside the app looking at the app.

It looks like in this schema the data points that would be self-reported by the app are nested in the `inAppStats` object, which I think is a nice approach. I'm curious why we wouldn't then nest the fields that would be recorded outside the app in an equivalent object? 

To be clear - I don't have a view on whether grouping and nesting these two categories of fields is a good thing to do - but I think whatever approach we take with one group we should also take with the other.

What do you think?

schemas/com.snowplowanalytics.monitor/application_status/jsonschema/1-0-0
```diff
@@ -0,0 +1,133 @@
...
+    "numNodes": {
+      "description": "The number of nodes (instances, containers, pods, etc) on which the application is running in the period to which this status applies",
+      "type": ["integer", "null"]
+    },
+    "nodeType": {
+      "description": "The type of nodes being used, e.g. EC2, Dataflow, Pod, etc",
+      "type": ["string", "null"]
+    },
+    "nodeSpecs": {
+      "description": "Detailed specifications for the nodes being used",
```
yalisassoon: @jbeemster - the way @dilyand has designed this schema allows for different node types and specification, whereas when we were brainstorming we assumed they would all be of the same type.

Is the approach @dilyand has taken safer? Or can we assume that the node types within an application will always be the same. (In which case this is over-engineering?)

schemas/com.snowplowanalytics.monitor/application_status/jsonschema/1-0-0
```diff
@@ -0,0 +1,133 @@
...
+      "description": "The average memory usage in the period to which this status applies; in MB",
+      "type": ["integer", "null"]
+    },
+    "avgNetworkIn": {
+      "description": "The average volume of inbound network traffic in the period to which this status applies; in Bytes per second",
+      "type": ["integer", "null"]
+    },
+    "avgNetworkOut": {
+      "description": "The average volume of outbound network traffic in the period to which this status applies; in Bytes per second",
+      "type": ["integer", "null"]
```
yalisassoon: Some other fields we considered here:

* `last_scale_up_tstamp`
* `last_scale_down_tstamp`

schemas/com.snowplowanalytics.monitor/status_report/jsonschema/1-0-0
```diff
@@ -0,0 +1,21 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for a status report event emitted from an application",
+  "self": {
+    "vendor": "com.snowplowanalytics.monitor",
+    "name": "status_report",
```
yalisassoon: @dilyand you prefer the name `status_report` to `report_my_status`?

I'm not tied to `report_my_status` particularly but keen to name it something that makes it clear this is the event - `report_status` sounds like a noun rather than a verb...

schemas/com.snowplowanalytics.monitor/status_report/jsonschema/1-0-0
```diff
@@ -0,0 +1,21 @@
...
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+
+  "type": "object",
+  "properties": {
+    "id": {
+      "description": "The ID of the status report; a UUID",
+      "type": "string",
```
yalisassoon: Do we need an ID for the event? (It will automatically be assigned an event ID and timestamp if it is a Snowplow event.)

yalisassoon:
Hey @dilyand have added my comments - would a quick call to discuss be helpful?

schemas/com.snowplowanalytics.monitor/application/jsonschema/1-0-0
dilyand: I thought it was better to structure the inputs in the same way as the outputs.

- If we only have one input, then the array will just have one value. I don't see a downside.
- If in future there's a use case with more than one input, then we'll just add another value to the array and won't need to upgrade the schema.
- This also supports a use case with an arbitrary number of inputs, potentially changing at different stages of the app execution.
- It's more consistent.

But happy to reconsider if this structure will make it harder to process and visualise the data.

schemas/com.snowplowanalytics.monitor/application_status/jsonschema/1-0-0
dilyand: The main reason I bundled these in the `inAppStats` array is to support reporting stats by tracker version. We don't know in advance how many tracker versions will be in use, so we need a flexible structure that can hold an arbitrary number of them.

If we do not need to slice by tracker version, then there is not real need for the `inAppStats` object and each field can just be a top-level property.

I did not take the same approach for the stats that are 'outside the app' because there we cannot guarantee that we can slice by tracker version. This schema is meant to not be specific to loaders. So, in some upstream process events that failed to be processed might not contain information about the tracker.

I think the best approach now would be to unnest these 'internal' stats.

schemas/com.snowplowanalytics.monitor/application_status/jsonschema/1-0-0
dilyand: Oh, it looks like I'd missed them. I'll add them to the schema.

schemas/com.snowplowanalytics.monitor/status_report/jsonschema/1-0-0
dilyand: No strong feelings, but I followed the Tracking Conventions drafted by Implementation Engineering:

> Custom event names to read `(object_)action` (with the exception of prefixes in next section), for example:
>- `button_click`
>- `nativead_impression`
>- `view`
>- Note: NOT `view_event`

So: `status_report`.

schemas/com.snowplowanalytics.monitor/status_report/jsonschema/1-0-0
dilyand: No immediate need, it's more that we can't have an empty schema for this event. But to make the case for a dedicated ID, in future we might want to report other events from inside the app and be able to assign them to a specific status report. This would act like `page_view_id` does on web.

schemas/com.snowplowanalytics.monitor/application/jsonschema/1-0-0
yalisassoon: You've won me over! Thanks

schemas/com.snowplowanalytics.monitor/application_status/jsonschema/1-0-0
yalisassoon: Agree

schemas/com.snowplowanalytics.monitor/status_report/jsonschema/1-0-0
yalisassoon: 👍

schemas/com.snowplowanalytics.monitor/status_report/jsonschema/1-0-0
yalisassoon: 👍

schemas/com.snowplowanalytics.monitor/application/jsonschema/1-0-0
```diff
@@ -0,0 +1,112 @@
...
+      "description": "The category of the application, eg loader",
+      "type": ["string", "null"],
+      "maxLength": 256
+    },
+    "version": {
+      "description": "The version of the application, eg 1.22.1",
+      "type": ["string", "null"],
+      "maxLength": 32
+    },
+    "instance_id": {
```
jbeemster: instance_id is a bit overloaded from AWS as the id of an EC2 "instance" - can this be just "id" and for "instance_name" just "name"?

dilyand: `instance_id` can be just `id` but we already have a `name` property:

```
"name": {
      "description": "The name of the application, eg rdb-loader",
      (...)
    },
"instance_name": {
      "description": "The name of the application instance, eg com-acme-prod1-rdb-loader-1",
      (...)
    }
```

I guess we can get rid of `name=rdb-loader` if it will always be clear what application that is from the (currently so called) `instance_name`?

jbeemster: Maybe name becomes app_name and instance_name becomes name?

schemas/com.snowplowanalytics.monitor/application/jsonschema/1-0-0
```diff
@@ -0,0 +1,112 @@
...
+          }
+        },
+        "required": ["id", "type"],
+        "additionalProperties": false
+      }
+    },
+    "cloud": {
+      "description": "The cloud on which the application is running",
+      "type": ["string", "null"],
+      "enum": ["aws", "gcp", "azure", "kafka"]
```
jbeemster: Kafka is not a cloud is it?

schemas/com.snowplowanalytics.monitor/application/jsonschema/1-0-0
dilyand: I think it helps to look at the fully qualified names:

```
com.snowplowanalytics.monitor.application.name
com.snowplowanalytics.monitor.application.instanceName
```

vs 

```
com.snowplowanalytics.monitor.application.appName
com.snowplowanalytics.monitor.application.name
```

I think what we have now works better that way. WDYT?

(Note to self: There are a few rogue snake_cases in these schemas.)

schemas/com.snowplowanalytics.monitor/application/jsonschema/1-0-0
dilyand: Right. I wanted to make sure the schemas can be used outside of those three clouds as well and Kafka is the most popular option for that. But maybe `private` will be better? Or even `other` -- including any private cloud and on-prem?

schemas/com.snowplowanalytics.monitor/application/jsonschema/1-0-0
```diff
@@ -0,0 +1,121 @@
...
+      "type": ["string", "null"],
+      "maxLength": 256
+    },
+    "version": {
+      "description": "The version of the application, eg 1.22.1",
+      "type": ["string", "null"],
+      "maxLength": 32
+    },
+    "instanceId": {
+      "description": "The ID of the application instance; a UUID",
```
chuwy: Can be `format: uuid`. We don't need `maxLength` in that case.

schemas/com.snowplowanalytics.monitor/application/jsonschema/1-0-0
```diff
@@ -0,0 +1,121 @@
...
+    "inputs": {
+      "description": "The inputs consumed by the application, such as streams etc",
+      "type": ["array", "null"],
+      "items": {
+        "description": "Details about the input",
+        "type": "object",
+        "properties": {
+          "id": {
+            "description": "The ID of the input; a UUID",
+            "type": "string",
```
chuwy: `format: uuid`

schemas/com.snowplowanalytics.monitor/pipeline/jsonschema/1-0-0
```diff
@@ -0,0 +1,26 @@
...
+    "vendor": "com.snowplowanalytics.monitor",
+    "name": "pipeline",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+
+  "type": "object",
+  "properties": {
+    "id": {
+      "description": "The ID of the pipeline; a UUID",
```
chuwy: `format: uuid`

schemas/com.snowplowanalytics.monitor/application/jsonschema/1-0-0
```diff
@@ -0,0 +1,121 @@
...
+      "maxLength": 64
+    },
+    "fabric": {
+      "description": "The infrastructure on which the application is deployed, eg EC2, Lambda, Kubernetes, Dataflow, ECS, Cloud Functions, etc",
+      "type": ["string", "null"],
+      "maxLength": 256
+    },
+    "tags": {
+      "description": "Tags attached to the application instance",
+      "type": ["array", "null"],
```
chuwy: Is this format of `[{"key1": "value1"}, {"key2": "value2"}]` dictated by producer? I feel that simple object with strinig keys and string values would be easier to parse.

dilyand: Do you mean:

```
"tags": {
      "description": "Tags attached to the application instance",
      "type": ["array", "null"],
      "items": {
        "description": "A key-value pair",
        "type": "string"
        }
```

and then pass data in as `["key1: value1", "key2: value2"]`?

Originally, it was:

```
"tags": {
      "description": "Tags attached to the application instance",
      "type": ["array", "null"],
      "items": {
        "description": "A key-value pair",
        "type": "object",
        "additionalProperties": true
        }
```

which would allow for `[{"key1": "value1"}, {...}]` but then the `keys` cannot be known in advance to create a table in Redshift.

dilyand:
Moved the schemas out of Iglu Central: https://github.com/snowplow-proservices/com.snowplowanalytics-schema-registry/pull/194

PR #979, opened 2019-07-04
R107

chuwy:


chuwy:
Hey @dilyand, @yalisassoon this release includes schemas from previous PRs you reviewed ([Snowflake bad rows](https://github.com/snowplow/iglu-central/pull/979/files#diff-cf48a80a4739dda7c8b63464a4f29f40) for Yali and [Sendgrid patches](https://github.com/snowplow/iglu-central/pull/979/files#diff-b5de960b0cb30a1048e060684491af55) for Dilyan). Could you please give your green ticks so Alex could merge it?

dilyand:
@chuwy Just to double check: for SendGrid we're patching schemas that may be already in use. Do we have a good process to let Support know about the implications?

chuwy:
Good question, @dilyand. I didn't notice any breaking changes when reviewed it, so upgrade should not have any negative impact and hence I don't think there's anything Support need to be aware of. But also, the only channel that I can think of is a group chat channel.

dilyand:
Won't tables have to be upgraded to accommodate the expanded `event_id` column size from `VARCHAR(48)` to `VARCHAR(4096)`?

chuwy:
> Support are aware of the impact of patched schemas and have a plan how to deal with affected pipelines.

Does it mean however we need to hold it on?

dilyand:
```
Leon  [8 minutes ago]
I'd be happier if we could wait until we have an overview of how many pipelines will be affected and have a plan to migrate those

Leon  [7 minutes ago]
@Yulia is already working on the first part which we hope to have soon.

Leon  [7 minutes ago]
Would it be possible to wait for that?
```

leonmaas:
Analysis: only one pipeline would be affected but in a good way; this would actually help them.
From our point of view we are ready to go.

dilyand:
The only thing I'm not sure about then is what Mutator will do (or, rather, not do). Because it will see shredded types that it already knows about: `2-0-0` so it won't change anything in the DDL. AFAICS, BQ DDLs just have a property type but no limits on length, so that should be fine. Still, best for @chuwy to confirm this.

chuwy:
That's right for Mutator, it doesn't care about lengths.

alexanderdean:
Missing a green tick from @yalisassoon .

@chuwy does the Redshift loader truncate columns by default? Otherwise I imagine the widening of `sg_event_id` in this PR would break loads?

yalisassoon:
Have approved the PR, but good to understand the answer to @alexanderdean question:

> @chuwy does the Redshift loader truncate columns by default? Otherwise I imagine the widening of sg_event_id in this PR would break loads?

If the loader doesn't truncate columns, we should be  ready with SQL table migration scripts to update the column lengths appropriately.

chuwy:
@alexanderdean, @yalisassoon yes, RDB Loader truncates columns:

https://github.com/snowplow/snowplow-rdb-loader/blob/master/src/main/scala/com/snowplowanalytics/snowplow/rdbloader/loaders/RedshiftLoadStatements.scala#L171

(I'll update the release date for 8th July)

PR #992, opened 2019-07-10
Rename focus form's 'elementType' field to 'type' (closes #991)

### LINKED ISSUE #991: Rename focus form's 'elementType' field to 'type'
Change form and focus form events collect the same data and are tracked via the same method. Currently the schema for focus_form events has an elementType field but no type field, where the change form event uses type - resulting in legitimate events going to bad rows.

Minor schema upgrade should fix this issue.

colmsnowplow:


colmsnowplow:
Taking a different approach to resolving - doesn't require a new schema. Closing.

PR #1003, opened 2019-08-27
focus_form: use correct property name for type

jethron:
The JS tracker uses `type` to be consistent with the change_form and submit_form schemas: https://github.com/snowplow/snowplow-javascript-tracker/blob/ed751aaf8486f72d0abc0f0f880754c3dd61aa65/core/lib/core.ts#L859

Otherwise valid events using v1-0-0 of this focus_form schema will be going to bad because the schema is incorrect vs what's implemented in the tracker (due to `"additionalProperties": false`).

Not sure if this should be a new version or not given the circumstance? Is it possible someone would be using the `elementType` property somehow in prod, not with the JS tracker?

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/focus_form/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow/focus_form/jsonschema/1-0-0
@@ -24,7 +24,7 @@
 			"type": "string",
 			"enum": ["INPUT", "TEXTAREA", "SELECT"]
 		},
-		"elementType": {
+		"type": {
 			"type": ["string", "null"],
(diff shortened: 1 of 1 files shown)


colmsnowplow:
Hey @jethron,

Thanks for raising! 

Although a schema change is the cleaner conceptual solution, we decided that the solution with the least overhead on existing implementations is to have the tracker send data to conform to the schema which already exists.

[That issue on the JS tracker](https://github.com/snowplow/snowplow-javascript-tracker/issues/731) is resolved in the next release - for which QA is just wrapping up as we speak. This release got delayed a bit due to issues with the testing framework and a change to the deployment process but we're almost ready to go now. 🙂

colmsnowplow:
PS. Sorry for the bother of having opened this PR!

PR #1013, opened 2019-10-31
R110

chuwy:
- [x] Finalize against snowplow-badrows property-based tests
- [x] Make **all** properties requried (since circe expresses absence via `null`) (cancelled)


schemas/com.snowplowanalytics.snowplow.badrows/enrichment_failures/jsonschema/1-0-0
```diff
@@ -0,0 +1,331 @@
...
+              "type": [ "string", "null" ],
+              "description": "Network user id associated with this payload",
+              "format": "uuid"
+            }
+          },
+          "required": [ "vendor", "version", "loaderName", "encoding" ],
+          "additionalProperties": false
+        },
+        "enriched": {
+          "description": "The partially enriched event that resulted in schema violations",
```
benjben: This will happen when we validate the context added by each enrichment or this is something else?

schemas/com.snowplowanalytics.snowplow.badrows/collector_payload_format_violation/jsonschema/1-0-0
```diff
@@ -0,0 +1,97 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for a format violation bad row",
```
benjben:
```suggestion
  "description": "Schema for a format violation in the collector payload",
```

schemas/com.snowplowanalytics.snowplow.badrows/loader_iglu_error/jsonschema/2-0-0
```diff
@@ -0,0 +1,657 @@
...
+          }
+        },
+        "required": ["schemaKey", "error"],
+        "additionalProperties": false
+      }
+    },
+    "payload": {
+      "type": "object",
+      "description": "The enriched event that resulted in a storage loader failure",
+      "properties": {
```
benjben: Out of curiosity, is it not possible to specify that the value of an object should match another schema ? e.g. [atomic event](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/atomic/jsonschema/1-0-0) here

schemas/com.snowplowanalytics.snowplow.badrows/loader_runtime_error/jsonschema/1-0-1
```diff
@@ -0,0 +1,524 @@
...
+            "maxLength": 512
+          },
+          "version": {
+            "type": "string",
+            "description": "Version of the artifact responsible for the creation of enrichment failures",
+            "pattern": "^(\\d+\\.\\d+\\.\\d+.*)$",
+            "maxLength": 32
+          }
+        },
+        "required": [ "artifact", "version" ],
```
benjben: Why not `processor` ? It's just a nice have?

schemas/com.snowplowanalytics.snowplow.badrows/tracker_protocol_violations/jsonschema/1-0-0
```diff
@@ -0,0 +1,373 @@
...
+      "properties": {
+        "timestamp": {
+          "type": "string",
+          "description": "Timestamp at which the failure occurred",
+          "format": "date-time"
+        },
+        "vendor": {
+          "type": "string",
+          "description": "Vendor of the adapter that processed this payload, here com.snowplowanalytics.snowplow",
+          "maxLength": 64
```
benjben:
```suggestion
          "enum": [ "com.snowplowanalytics.snowplow"]
```

schemas/com.snowplowanalytics.snowplow.badrows/tracker_protocol_violations/jsonschema/1-0-0
```diff
@@ -0,0 +1,373 @@
...
+        },
+        "version": {
+          "type": "string",
+          "description": "Version of the adapter that processed this payload, here tp2",
+          "maxLength": 16
```
benjben:
```suggestion
         "enum": [ "tp2"]
```

schemas/com.snowplowanalytics.snowplow.badrows/enrichment_failures/jsonschema/1-0-0
```diff
@@ -0,0 +1,331 @@
...
+    "payload": {
+      "type": "object",
+      "properties": {
+        "raw": {
+          "type": "object",
+          "description": "The raw event extracted from collector payload",
+          "properties": {
+            "vendor": {
+              "type": "string",
+              "description": "Vendor of the adapter that processed this payload, here com.snowplowanalytics.snowplow",
```
benjben:
```suggestion
              "description": "Vendor of the adapter that processed the collector payload containing the event to enrich",
```

schemas/com.snowplowanalytics.snowplow.badrows/adapter_failures/jsonschema/1-0-0
```diff
@@ -0,0 +1,243 @@
...
+                    "maxLength": 64
+                  },
+                  "value": {
+                    "type": [ "string", "null" ],
+                    "description": "Stringified representation of the value which did not meet expectations"
+                  },
+                  "expectation": {
+                    "type": "string",
+                    "description": "Expectation which was not met",
+                    "maxLength": 256
```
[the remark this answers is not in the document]
chuwy: Not really. But my reasoning was that:

1. Nothing should really validate bad rows against their schemas
2. Most of our current warehouses don't have a notion of varchar size
3. One that have (Redshift) truncates these properties (and doesn't support JSON arrays, so not sure it even makes sense)
4. In the end, no matter what value we choose - there's always a chance it isn't enough
5. In absolute most of the cases this should be enough

TL;DR all these `maxLength`s are 99% informational

schemas/com.snowplowanalytics.snowplow.badrows/adapter_failures/jsonschema/1-0-0
```diff
@@ -0,0 +1,243 @@
...
+                  }
+                },
+                "required": ["field", "expectation" ],
```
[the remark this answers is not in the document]
chuwy: Actually, it's always `null` as per circe encoding. My current plan is to switch add all properties to `required`. Absence should be expressed via `null`

schemas/com.snowplowanalytics.snowplow.badrows/enrichment_failures/jsonschema/1-0-0
```diff
@@ -0,0 +1,331 @@
...
+            "type": "object",
+            "description": "Individual enrichment failure or null in case of validation",
+            "properties": {
+              "enrichment": {
+                "type": ["object", "null"],
+                "description": "Information needed to identify an enrichment",
+                "properties": {
+                  "schemaKey": {
+                    "type": "string",
+                    "description": "The schema coordinates for this enrichment"
```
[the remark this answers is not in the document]
chuwy: The configuration. Will make it cleaner.

schemas/com.snowplowanalytics.snowplow.badrows/enrichment_failures/jsonschema/1-0-0
chuwy: No, this is a payload we had after (failed) enrichments.

schemas/com.snowplowanalytics.snowplow.badrows/loader_iglu_error/jsonschema/2-0-0
chuwy: No, unfortunately. It is often requested by users and would be a killer feature for a project like bad rows (we re-use a lot of schemas), but we never got there.

schemas/com.snowplowanalytics.snowplow.badrows/loader_runtime_error/jsonschema/1-0-1
chuwy: To make it backward-compatible with 1-0-0. I released RDB Loader with this bad row that didn't have a `processor`, so we need to remain compatible with it.

schemas/com.snowplowanalytics.snowplow.badrows/adapter_failures/jsonschema/1-0-0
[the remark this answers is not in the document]
benjben: > Nothing should really validate bad rows against their schemas
I had never thought about that!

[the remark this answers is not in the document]
chuwy: I agree, this is not obvious. But if you think about it:

1. Property-based tests in snowplow-badrows make us very confident that data conforms these schemas
2. If a bad row is in invalid in its consumer (e.g. was written not by snowplow-badrows) then it should be something "completely invalid", like missing required property (maybe multiple ones) rather than "slightly invalid" like exceeded string length
3. Both "completely invalid" and "slightly invalid" bad rows would end up in some kind of dead end and we likely are not interested in reasons and don't expect this to happen at all in a common case.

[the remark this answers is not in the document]
benjben: Makes sense!

chuwy:
Hi @alexanderdean can you merge please these schemas for Snowplow R118 and BigQuery Loader 0.2.0.

PR #1015, opened 2019-11-27
Add amp_id and amp_web_page schemas

colmsnowplow:
Schemas for default tracking via AMP tracker v2. amp_id tracks AMP unique id (CLIENT_ID) and maps to domain_userid (where cross-domain linking picks up the duid).

amp_web_page aims at tracking their parallel of the page view ID. Unfortunately the implementation of a globally unique pvid appears to be broken (issue: https://github.com/ampproject/amphtml/issues/25794), so until that's fixed we track a pageViewId which is unique to combination of url and date, plus the timestamp and url (no available means of tracking date only).

Users will need to combine these in modeling until such a time as ampPageViewId64 becomes available - at that point we can decide to track all of them, or just the ampPageViewId64.


colmsnowplow:
Happy to discuss reasoning before upload & merge.

schemas/com.snowplowanalytics.amp/amp_id/jsonschema/1-0-0
```diff
@@ -0,0 +1,30 @@
...
+  },
+  "description": "Schema for AMP user identification",
+  "properties": {
+    "ampClientId": {
+      "type": "string",
+      "description": "AMP client ID - unique to AMP user",
+      "maxLength": 128
+    },
+    "domainUserId": {
+      "type": [
```
chuwy: Do we really express an optionality of property by `null` and absence of property? We'd like to pick a single approach, e.g. have `type: [string, null]` *and* have it in `required` or have `type: string` and not list in `required`. That's totally okay though if we don't control the data.

colmsnowplow: Ah so maybe this is my schema rustiness combined with a slightly atypical case. We will always have the field, but we have no control over whether or not it contains a value. 

So actually, it seems I should be requiring the field and allowing nulls?

colmsnowplow:
> Unfortunately the implementation of a globally unique pvid appears to be broken

Actually they've responded to that issue with a potential solution. If it works I'll rationalise that schema down to just that one unique ID (which would be FAR better).

colmsnowplow:
Realised that a) I missed a couple of things and b) I hadn't followed the naming standards. 

Rebased to fix that and address your comment @chuwy. Apologies for creating more work than was needed, can I get another pass on this.

@yalisassoon as an FYI - these are my proposals for default contexts to be sent with every v2 AMP event. I'll be checking in with you on a related design decision before finalising the PR to google/AMP shortly.

Finally, @alexanderdean are you still the maintainer of this repo/the one who merges? If so I'll add you as reviewer once any further change requests are sorted. :)

schemas/com.snowplowanalytics.amp/amp_id/jsonschema/1-0-0
colmsnowplow: Fixed in the rebase now, I believe.

alexanderdean:
> Finally, @alexanderdean are you still the maintainer of this repo/the one who merges? If so I'll add you as reviewer once any further change requests are sorted. :)

Thanks, just add me at the last moment @colmsnowplow !

schemas/com.snowplowanalytics.amp/amp_id/jsonschema/1-0-0
dilyand: > We'd like to pick a single approach, e.g. have `type: [string, null]` *and* have it in `required` or have `type: string` and not list in `required`. 

If a property is not required but its type is simply `type: string`, then sending that property with a `null` value will fail to validate. So for a truly optional field (ie it can be missing, or null), we need *both* `type: [string, null]` and not required.

schemas/com.snowplowanalytics.amp/amp_id/jsonschema/1-0-0
```diff
@@ -0,0 +1,40 @@
...
+        "string",
+        "null"
+      ],
+      "description": "Snowplow domain userid - tracked if available",
```
dilyand:
```suggestion
      "description": "Snowplow domain_userid - tracked if available",
```

schemas/com.snowplowanalytics.amp/amp_id/jsonschema/1-0-0
```diff
@@ -0,0 +1,40 @@
...
+      "maxLength": 128
+    },
+    "userId": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "description": "User-defined userid - tracked if available",
```
dilyand:
```suggestion
      "description": "User-defined user_id - tracked if available",
```

schemas/com.snowplowanalytics.amp/amp_id/jsonschema/1-0-0
```diff
@@ -0,0 +1,40 @@
...
+      "maxLength": 255
+    }
+  },
+  "additionalProperties": false,
```
dilyand: Are we sure AMP won't add more properties on a whim? Or is this data coming from our tracker, meaning we control the fields?

schemas/com.snowplowanalytics.amp/amp_id/jsonschema/1-0-0
```diff
@@ -0,0 +1,40 @@
...
+    "version": "1-0-0"
+  },
+  "description": "Schema for AMP user identification",
+  "properties": {
+    "ampClientId": {
+      "type": "string",
+      "description": "AMP client ID - unique to AMP user",
+      "maxLength": 128
+    },
+    "domainUserId": {
```
dilyand:
```suggestion
    "domainUserid": {
```

sql/com.snowplowanalytics.amp/amp_id_1.sql
```diff
@@ -0,0 +1,26 @@
...
+    "schema_name"    VARCHAR(128)  ENCODE ZSTD NOT NULL,
+    "schema_format"  VARCHAR(128)  ENCODE ZSTD NOT NULL,
+    "schema_version" VARCHAR(128)  ENCODE ZSTD NOT NULL,
+    "root_id"        CHAR(36)      ENCODE RAW  NOT NULL,
+    "root_tstamp"    TIMESTAMP     ENCODE ZSTD NOT NULL,
+    "ref_root"       VARCHAR(255)  ENCODE ZSTD NOT NULL,
+    "ref_tree"       VARCHAR(1500) ENCODE ZSTD NOT NULL,
+    "ref_parent"     VARCHAR(255)  ENCODE ZSTD NOT NULL,
+    "amp_client_id"  VARCHAR(128)  ENCODE ZSTD NOT NULL,
+    "domain_user_id" VARCHAR(128)  ENCODE ZSTD,
```
dilyand: I've suggested a change in the schema, so that this field matches the `atomic` spelling. But the DDL will have to be regenerated if accepted.

schemas/com.snowplowanalytics.amp/amp_web_page/jsonschema/1-0-0
```diff
@@ -0,0 +1,20 @@
+{
+	"$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+	"self": {
+		"vendor": "com.snowplowanalytics.amp",
```
dilyand: Does this use tabs instead of spaces? It looks way wider than the other schema.

schemas/com.snowplowanalytics.amp/amp_id/jsonschema/1-0-0
colmsnowplow: I believe it can't be `null` value unless something goes wrong. But yeah I'll revert to both not required and nullable as this is the safer case.

schemas/com.snowplowanalytics.amp/amp_id/jsonschema/1-0-0
colmsnowplow: Yes. We control it.

schemas/com.snowplowanalytics.amp/amp_web_page/jsonschema/1-0-0
colmsnowplow: Weird. Will fix.

colmsnowplow:
@alexanderdean @chuwy has suggested we can include this as part of R110 release. Not sure if the process still requires your review first, but if so it's ready now!

alexanderdean:
@colmsnowplow - I don't need to review this PR, I will just review and publish the R110 branch when it is finally complete...

chuwy:
Closing in favor of https://github.com/snowplow/iglu-central/pull/1013

PR #1030, opened 2020-03-02
client_session schema update to resolve type issue

ejensen:
Extracts the single `type` string out of an array containing a single value.
This resolves issues with JSONSchema parsers fail when type arrays contain a single value:
> Union Type Schemas must contain two or more Simple Type Schemas

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-1
+++ b/schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-1
@@ -33,9 +33,7 @@
 			"enum": [ "SQLITE", "COOKIE_1", "COOKIE_3", "LOCAL_STORAGE", "FLASH_LSO" ]
 		},
 		"firstEventId": {
-			"type": [
-				"string"
-			],
+			"type": "string",
 			"format": "uuid"
 		}
 	},
--- a/schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-1
+++ b/schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-1
@@ -30,6 +30,7 @@
 			"format": "uuid"
 		},
 		"storageMechanism": {
+			"type": "string",
 			"enum": [ "SQLITE", "COOKIE_1", "COOKIE_3", "LOCAL_STORAGE", "FLASH_LSO" ]
 		},
 		"firstEventId": {


paulboocock:
Hi @ejensen 
I'm going to cherry-pick your changes into another branch so we can generate the required jsonpath and ddl files to accompany this change.
As firstSessionId isn't required we will need the type to stay as an array but it should allow nulls. We'll bump the schema to 1.0.2 for this change.

ejensen:
@paulboocock The property that was modified was `firstEventId` which is not a required field. It can be excluded from the body, instead of sending `null` in the body.

The `firstSessionId` property you mentioned is required, and already has a type array that includes `null`.

Replacing `firstEventId`'s type array with a single value for `firstEventId` is a non-breaking change that resolves the schema parsing issue.
Making a new version (1.0.2) would require [all the client SDKs](https://github.com/search?q=org%3Asnowplow+iglu%3Acom.snowplowanalytics.snowplow%2Fmobile_context%2Fjsonschema%2F1-0-1&type=Code) to update in order to get the fix.

paulboocock:
Apologies, I meant `firstEventId` in my earlier message (edited!).

I'd prefer not to introduce 1.0.2 as well, for the exact reason you mention. 
We generally specify `null` as a type on unrequired attributes - I believe this to be simply for clarity. I'm running this by my colleague to get confirmation on which direction we should go in.

paulboocock:
Also, I added a comment to https://github.com/snowplow/snowplow-android-tracker/issues/362
Perhaps you also have visibility of this @ejensen to help me appreciate the severity? Whilst I appreciate the validation error presented here, I'm curious how this is manifesting itself for you?

ejensen:
@paulboocock We are using [Fivetran](https://fivetran.com/docs/events/snowplow) to transfer events into our data warehouse. Fivetran's JSON schema parser is reporting the schema validation error (shown above) for the client session context data. Resulting in the session context data being dropped. 

Without the event session data being transferred to the data warehouse. We've fallen back to manual sessionization.

To verify that this schema change fixes the issue, we forked the schema and the Snowplow SDKs (to use the forked schema). With these changes, the session context passed Fivetran's validation and the data made it to the warehouse

paulboocock:
Thank you for the context. It might also be worth bringing this up with Fivetran as producing an error in this case seems rather strict in my opinion, a warning would likely suffice and the data should still be "load-able". The Snowplow loaders can deal with this fine.

In regards to specifing `null` as a type option, this is an old convention we use generally use in Snowplow schemas. We do this to prefer "explicit absense". It's a fairly weak requirement but we do this as it has some benefits for schema-inference and weird use cases like templating.
I'd hope that Fivetran will work with the schema proposed in #1032 - as the field was already nullable (not in required) we can do an in-place mutation of this schema (not increasing version number). 

This will go out in the next Iglu Central release @ejensen, I'll update again when I have a more concrete timeframe.

PR #1032, opened 2020-03-09
Patch com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-1

paulboocock:
Fixes issues mentioned in #1030

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-1
+++ b/schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-1
@@ -11,32 +11,41 @@
 	"properties": {
 		"userId": {
 			"type": "string",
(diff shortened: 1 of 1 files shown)


paulboocock:
Not sure if we can edit this in place. Can make 1.0.2 if editing 1.0.1 isn't a suitable solution.

paulboocock:
> Not sure if we can edit this in place. Can make 1.0.2 if editing 1.0.1 isn't a suitable solution.

Atfer speaking to @chuwy it seems an in-place mutation of this schema is ok. It’s not a problem in this particular case because field was nullable anyway (not in required). 
We should always avoid in-place mutations but due to the knock on consequences of having to adjust every SDK, an in-place mutation seems acceptable on this already nullable field.

PR #1034, opened 2020-03-12
AppsFlyer Push API v2.0

joaolcorreia:
+ Added schema for attribution events for AppsFlyer Push API v2.0
+ Validated with example hits on Snowplow Mini

Example payloads:
https://support.appsflyer.com/hc/en-us/articles/207034356-Push-API-V2-0-real-time-attribution-event-reporting


schemas/com.appsflyer/attribution/jsonschema/1-0-0
```diff
@@ -0,0 +1,566 @@
...
+        "maxLength": 1024
+    }, 
+    "api_version": {
+        "description": "API Version e.g. 2.0",
+        "type": ["string", "null"],
+        "maxLength": 10
+    }, 
+    "app_id": {
+        "description": "App ID",
+        "type": ["string", "null"],
```
paulboocock: `app_id` is a required field but is shown here as a allowing nulls. I imagine this should be `"type": "string"` assuming `app_id` is always available from AppsFlyers Push API V2.

joaolcorreia: Fixed. Im not sure if they always send app_id so I removed it.

Issue #1035, opened 2020-03-20
Patch schemas/dev.amp.snowplow/amp_page_ping/jsonschema/1-0-0


colmsnowplow:
It has become apparant that the values for page ping data from AMP, while normally compatible as integers, may be decimals.

No tracking methods which use this schema are yet in use, so it's safe to patch the schema in-place to replace all integer fields as numbers.

PR #1036, opened 2020-03-20
Patch schemas/dev.amp.snowplow/amp_page_ping/jsonschema/1-0-0 (closes…

colmsnowplow:
… #1035)


colmsnowplow:
The method that uses this schema is not yet in use, so I'm hoping it's ok to edit in place.

PR #1039, opened 2020-03-26
R112

### LINKED ISSUE #18: Add iglu:com.appsflyer/attribution/jsonschema/1-0-0
http://support.appsflyer.com/entries/69796693-Accessing-AppsFlyer-Attribution-Conversion-Data-from-the-SDK-Deferred-Deep-linking-

### LINKED ISSUE #1031: Patch com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-1
`firstSessionId` is an array but only contains a single item, which is invalid. `firstSessionId` should also list `"null"` as it is an optional property of the client_session schema.

`storageMechanism` also contains no `type` attribute, this should be `string`.

### LINKED ISSUE #1035: Patch schemas/dev.amp.snowplow/amp_page_ping/jsonschema/1-0-0

### LINKED ISSUE #1038: Patch com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-1
Specifies that the networkType enum type is a string.
This should be a non-breaking change since it is only adding more specificity to the schema.
A type specification of enums is required for integrations such as Fivetran:

> If you want to store an enum value, add a type associated with it. All the enum values should be of same data type

paulboocock:
Contains 3 patches and 1 new schema. 3 community contributions and 1 for the AMP tracker.

~Add iglu:com.appsflyer/attribution/jsonschema/1-0-0 (close #18)~
Patch schemas/dev.amp.snowplow/amp_page_ping/jsonschema/1-0-0 (closes #1035)
Patch com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-1 (close #1038)
Patch com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-1 (close #1031)


paulboocock:
Side note: We need to ensure the mirrors are in the correct mode before publishing this so that the patch for #1035 replicates to the mirrors.

paulboocock:
**dev.amp.snowplow/amp_page_ping/jsonschema/1-0-0**
This is currently unused by the AMP tracker so is safe to patch in place. Future AMP tracker release will make use of this.

**com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-1 and com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-1**
Contain no structural changes that would effect loading, only better labelling of attributes (for instance marking enums as strings, as per the specification and adding descriptions).

paulboocock:
Removed Add iglu:com.appsflyer/attribution/jsonschema/1-0-0 #18 due to inconsistencies with the AppsFlyer API. This leads to the data failing to validate against this schema.

PR #1042, opened 2020-04-02
Update query string encoding for snowplow-badrows

benjben:
`$.data.payload.raw.parameters` went from

```json
"parameters": {
  "type": "object",
   "description": "Tracker protocol parameters"
}
```

to

```json
            "parameters": {
              "type": [ "array", "null" ],
              "description": "Query string of the collector payload containing this event",
              "items": {
                "type": "object",
                "properties": {
                  "name": {
                    "type": "string",
                    "description": "Name of the querystring parameter",
                    "maxLength": 512
                  },
                  "value": {
                    "type": [ "string", "null" ],
                    "description": "Possible value for the querystring parameter",
                    "maxLength": 512
                  }
                },
                "required": [ "name", "value" ],
                "additionalProperties": false
              }
            },
```

For `SchemaViolations` and `EnrichmentFailures`

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow.badrows/enrichment_failures/jsonschema/2-0-0
+++ b/schemas/com.snowplowanalytics.snowplow.badrows/enrichment_failures/jsonschema/2-0-0
@@ -1,0 +1,483 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for a bad row resulting from enrichment failures",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow.badrows",
+    "name": "enrichment_failures",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "failure": {
+      "type": "object",
+      "description": "Information regarding the enrichment failures",
+      "properties": {
+        "timestamp": {
+          "type": "string",
+          "description": "Timestamp at which the failure occurred",
+          "format": "date-time"
+        },
+        "messages": {
+          "type": "array",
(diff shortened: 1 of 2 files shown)


paulboocock:
The R112 release has already been pushed to production. This will need to move to R113, updated the milestone to reflect this.

benjben:
We missed the slot, too bad! Thanks for the notice @paulboocock

benjben:
Following https://github.com/snowplow-incubator/snowplow-badrows/issues/39 `v_collector` and `v_etl` got updated to `"type": ["string", "null"]`

chuwy:
Closing in favor of https://github.com/snowplow/iglu-central/pull/1046

benjben:
@chuwy So here is the exhaustive list of updates in both `SchemaViolations` and `EnrichmentFailures` :

1. `$.data.payload.raw.parameters` went from `object` to `array` of `object` each with `name` and `value`
2. `v_collector` and `v_etl` got updated from `"type": "string"` to `"type": ["string", "null"]`
3. `dvce_ismobile` got updated from `"type": ["string", "null"]` to `"type": ["integer", "null"]`

chuwy:
Just spotted, @benjben, the SchemaVers do not correspond https://github.com/snowplow/iglu-central/pull/1042/files#diff-f9bd73efbb5a753902d043346691ca17R8. I'll fix this in R113, but please check it with igluctl next time.

benjben:
Oh no, good catch! Sorry about that @chuwy !

I'll use `igluctl` next time. We should add it to Travis so that schemas for PRs are automatically checked.

chuwy:
>  We should add it to Travis so that schemas for PRs are automatically checked.

They're checked and it would not allow to push them, but I don't know why we have no Travis checks in the PR.

PR #1046, opened 2020-04-22
R113

chuwy:
Bad rows + R119

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,12 @@
+Release 113 (2020-04-24)
+------------------------
+Add com.snowplowanalytics.snowplow.badrows/recovery_error/jsonschema/1-0-0 (#1044)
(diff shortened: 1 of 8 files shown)


colmsnowplow:
@chuwy can we document what impact the two patches would have on existing pipelines/why we think it's safe please?

benjben:
Hey @colmsnowplow , without this patch, `EnrichmentFailures 1-0-0` and `SchemaViolations 1-0-0` emitted by enrich are not valid according to their schema, so any consumer that would try to validate them would fail. The patch is only there to make these 2 bad rows valid. I can't think of any impact on existing (QA) pipelines.

chuwy:
To add to that: none of current consumers actually validate the data against existing schemas, but if they did - bad rows would be invalid. What we're doing is making sure that in future if consumers start validate data or DDLs were derived out of these schemas - data would be valid.

PR #1051, opened 2020-05-05
R114

peel:
Regex filtering in recovery config

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+Release 114 (2020-05-25)
+------------------------
+Add com.snowplowanalytics.snowplow/recoveries/jsonschema/3-0-0 (#1050)
+
 Release 113 (2020-04-24)
 ------------------------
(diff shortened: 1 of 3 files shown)


chuwy:
The patch is safe because `2-0-0` has never been used anywhere and recovery job 0.2.0 is not released yet. @colmsnowplow, @jbeemster could you please have a look.

jbeemster:
If this has never been used anywhere presumably it would also be fairly simple to just increment the schema version to avoid patching?

chuwy:
> it would also be fairly simple to just increment the schema version to avoid patching?
That's true. I think we should make this a rule a thumb to create patches only when there's no other way to fix production software.

@peel could you please rework this into `3-0-0`? We should just abandon `2-0-0`

peel:
@chuwy Done

chuwy:
Fixed the CHANGELOG entry. 

@jbeemster I think its ready now.

peel:
Hey guys, what's the plan on this one? Can it be released?

PR #1055, opened 2020-06-17
Add com.maxmind/anonymous_ip/jsonschema/1-0-0

miike:
Schema for https://github.com/snowplow/scala-maxmind-iplookups/pull/133

I haven't included IP address in this schema as I'm not sure whether it belongs here or not. I think there is a legitimate use case for it in certain circumstances (e.g., where data is shredded and you may not want to run an expensive join back to atomic.events) but this matter less in BQ / Snowflake.

cc @dilyand @chuwy


chuwy:
Also, I think we can live without JSONPaths and SQL definitions now as RDB Loader has official automgirations.

miike:
Let's go with leaving IP address out.

The uses cases are mostly Redshift specific e.g., find all IP ranges that are proxies and filter this as part of another query (without joining back to events). If we extrapolate this to the other Maxmind schemas e.g., precision, connection type etc and each of these had an IP address associated with the schema it would not make much sense.

PR #1058, opened 2020-07-07
Generic bad row

benjben:


schemas/com.snowplowanalytics.snowplow.badrows/generic/jsonschema/1-0-0
```diff
@@ -0,0 +1,65 @@
...
+            "maxLength": 32
+          }
+        },
+        "required": [ "artifact", "version" ],
+        "additionalProperties": false
+      },
+      "payload": {
+        "type": "object",
+        "properties": {
+          "event": {
```
chuwy, line 34: If it was just `payload` (like in other [unstructured payloads](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow.badrows/loader_parsing_error/jsonschema/2-0-0#L12-L15)) string we could use common [`RawPayload`](https://github.com/snowplow-incubator/snowplow-badrows/blob/master/src/main/scala/com.snowplowanalytics.snowplow.badrows/Payload.scala#L49)

schemas/com.snowplowanalytics.snowplow.badrows/generic/jsonschema/1-0-0
```diff
@@ -0,0 +1,65 @@
+{
+    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "description": "Generic bad row",
+    "self": {
+      "vendor": "com.snowplowanalytics.snowplow.badrows",
+      "name": "generic",
```
chuwy, line 6: Don't have anything against just `generic`, but it feels that for consistency we can use at least a noun instead of adjective: generic error, generic failures or something of that kind?

schemas/com.snowplowanalytics.snowplow.badrows/generic/jsonschema/1-0-0
```diff
@@ -0,0 +1,65 @@
+{
+    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "description": "Generic bad row",
```
chuwy, line 3:
```suggestion
    "description": "Generic, very unstructured bad row that can be emitted by any Snowplow component",
```

schemas/com.snowplowanalytics.snowplow.badrows/generic/jsonschema/1-0-0
benjben: I actually hesitated to reuse `RawPayload`, I hadn't done it only because of the comment that was saying that this is for payloads received by a collector, but you're right it makes more sense to reuse it and update the comment.

schemas/com.snowplowanalytics.snowplow.badrows/generic/jsonschema/1-0-0
benjben: Hehe I also hesitated for this, I had first used `generic_error` and then switched to just `generic`. Switching back to `generic_error`

schemas/com.snowplowanalytics.snowplow.badrows/generic/jsonschema/1-0-0
chuwy: Yeah let's fix that "collector" part - some loader bad rows also have it

chuwy:
Closing in favor of https://github.com/snowplow/iglu-central/pull/1063

PR #1062, opened 2020-07-10
Add org.ietf/http_client_hints/jsonschema/1-0-0 (closes #1061)

### LINKED ISSUE #1061: Add org.ietf/http_client_hints/jsonschema/1-0-0
Chrome 84 will add a new feature calls User Agent Client Hints. This is in preparation for freezing the User Agent string in the near future.

This schema will allow us to collect the new Client Hints as outlined here: https://tools.ietf.org/html/draft-ietf-httpbis-client-hints-15

The values are available in JavaScript as detailed here: https://wicg.github.io/ua-client-hints/#getters

This will be added to the JavaScript 2.15.0 release.

paulboocock:
Required before 2.15.0 of the JavaScript can be released. 

The `isMobile` and `brands` properties will always be populated. The other 5 properties are deemed "High Entropy" and will be optionally requested, therefore they might not be present when sending this context.

I appreciate that `brands` is a little complex to deal with but the browser is allowed to respond with mulitple brands. For example, Chrome currently responses with Chrome, v84 and Chromium, v84 as two separate entities. It's expected Edge will also adopt this feature soon and will respond with Edge, v84 and Chromium, v84. There is no guarenteed order to these so there is little I can do to select a single one and it seems the best action is to capture them all and attempt to deal with them downstream.

### DIFF
--- a/schemas/org.ietf/http_client_hints/jsonschema/1-0-0
+++ b/schemas/org.ietf/http_client_hints/jsonschema/1-0-0
@@ -1,0 +1,55 @@
+{
+  "$schema" : "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "self" : {
+    "vendor" : "org.ietf",
+    "name" : "http_client_hints",
+    "version" : "1-0-0",
+    "format" : "jsonschema"
+  },
+  "type" : "object",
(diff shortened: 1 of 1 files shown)


chuwy:
Closing in favor of https://github.com/snowplow/iglu-central/pull/1063

PR #1083, opened 2020-10-09
RDB loader redesign : schema for SQS messages

benjben:


schemas/com.snowplowanalytics.snowplow.storage.rdbshredder/output/jsonschema/1-0-0
```diff
@@ -0,0 +1,68 @@
...
+      "min_timestamp": {
+        "description": "Minimum timestamp in shredded data",
+        "type": "string",
+        "format": "date-time"
+      },
+      "max_timestamp": {
+        "description": "Maximum timestamp in shredded data",
+        "type": "string",
+        "format": "date-time"
+      },
```
benjben, line 26: I think that `s3_path`, `min_timestamp` and `max_timestamp` should be with each element of the array, so in the end we would have nothing else than an array of shredded types

chuwy:
I think it's been cherry-picked already. Closing.

PR #1087, opened 2020-12-07
Patch form tracking schemas

### LINKED ISSUE #1086: Patch com.snowplowanalytics.snowplow/focus_form/jsonschema/1-0-0
Required to fix https://github.com/snowplow/snowplow-javascript-tracker/issues/872

paulboocock:
Resolves #1086


schemas/com.snowplowanalytics.snowplow/focus_form/jsonschema/1-0-1
```diff
@@ -0,0 +1,45 @@
...
+		"elementType": {
+			"description": "The type of input element",
+			"type": ["string", "null"],
+			"enum": ["button", "checkbox", "color", "date", "datetime", "datetime-local", "email", "file", "hidden", "image", "month", "number", "password", "radio", "range", "reset", "search", "submit", "tel", "text", "time", "url", "week"]
+		},
+		"elementClasses": {
+			"description": "An array of the classes on the element",
+			"type": ["array", "null"],
+			"items": {
+				"type": "string"
```
cksnp, line 35: Presumably this could use a description too?

cksnp (approved):
Approving, but I can't help wondering: Instead of removing `maxLength` on this one, would it not be better to add `maxLength` in a new version of the other two form schemas? If the defined `maxLength` is too small, maybe they can all be bumped adding an adequate length?

chuwy (approved):
Generally looks good, but:

1. I agree with Costas - it's always better to create new one. I assume that however we want to fix _existing_ tracking (and it is a _big_ problem). In case of existing tracking new schema won't help. I just don't think I can asses the impact
2. Instead of removing `maxLength` entirely I'm all for just overprovisioning it. We often still bounded by some restriction (i.e. 65K in Redshift), I think it's better to make it explicit. Maybe unreasonably big, but explicit.

But assuming those two points were taken in consideration.

paulboocock:
Based on these comments, I'm going to make some changes to this PR. I agree that being more explicit around the length would be beneficial.

paulboocock:
I've realised from your comments, I could make some better choices to resolve this in a way that brings nice improvements/consistency to the form tracking data.

So, I've updated `focus_form` with explicit lengths and patched the existing schema. I've also updated the other form schemas to now include the same explicit lengths. I see this is a considerable improvement as the three very related form tracking features will now behave the same for a specific form (rather than some events passing validation and some failing for the same form). 

I've also tidied up the schemas with descriptions. Lastly, as these are 1-0-0 schemas that I'm patching, I've updated the .sql files too - that seemed like good practice to me.

sql/com.snowplowanalytics.snowplow/change_form_1.sql
```diff
@@ -1,43 +1,29 @@
...
+    "schema_vendor"   VARCHAR(128)   ENCODE ZSTD NOT NULL,
+    "schema_name"     VARCHAR(128)   ENCODE ZSTD NOT NULL,
+    "schema_format"   VARCHAR(128)   ENCODE ZSTD NOT NULL,
+    "schema_version"  VARCHAR(128)   ENCODE ZSTD NOT NULL,
+    "root_id"         CHAR(36)       ENCODE RAW  NOT NULL,
+    "root_tstamp"     TIMESTAMP      ENCODE ZSTD NOT NULL,
+    "ref_root"        VARCHAR(255)   ENCODE ZSTD NOT NULL,
+    "ref_tree"        VARCHAR(1500)  ENCODE ZSTD NOT NULL,
+    "ref_parent"      VARCHAR(255)   ENCODE ZSTD NOT NULL,
+    "element_id"      VARCHAR(4096)  ENCODE ZSTD NOT NULL,
```
chuwy, line 17: I don't think we can merge this (`element_id` and `form_id` are swapped). If user creates the table and uses existing JSONPath (I don't see them in the PR) it means that they will be loading into wrong columns. If we create JSONPath - it means that users who already have tables deployed will be loading into wrong columns. This is the exact the case where patches are super dangerous.

It's okay to have different encodings and even constraints, but order is super crucial here. Especially because almost all clients with `form` schemas still don't have automigrartions.

chuwy (changes requested):
Hey @paulboocock! We cannot patch SQLs/JSONPaths with order changed, unfortunately.

sql/com.snowplowanalytics.snowplow/change_form_1.sql
paulboocock: Fixed the order of this and `submit_form`. You learn something new every day 😄 Also ensured they are all consistent with the existing JSONPaths.

chuwy (approved):

chuwy:

PR #1093, opened 2020-12-11
Add datamodeling config schema

adatzer:


schemas/com.snowplowanalytics.datamodeling/config/jsonschema/1-0-0
```diff
@@ -0,0 +1,115 @@
...
+        "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+        "self": {
+                "vendor": "com.snowplowanalytics.datamodeling",
+                "name": "config",
+                "version": "1-0-0",
+                "format": "jsonschema"
+        },
+        "type": "object",
+        "properties": {
+                "schema": {
```
colmsnowplow, line 11: Schema and data shouldn't explicitly be in the schema. They're the 'self-describing' part. The schema describes only the properties of the `data` part of the Self-describing object.

('schema' is in fact encapsulated by the 'self' portion of the schema itself.

schemas/com.snowplowanalytics.datamodeling/config/jsonschema/1-0-0
```diff
@@ -0,0 +1,115 @@
...
+                                },
+                                "playbooks": {
+                                        "type": "array",
+                                        "items": {
+                                                "type": "object",
+                                                "properties": {
+                                                        "playbook": {
+                                                                "type": "string",
+                                                                "minLength": 1,
+                                                                "maxLength": 64
```
colmsnowplow, line 44: I think across the board let's go with longer maxLength, or just no maxLength at all.

schemas/com.snowplowanalytics.datamodeling/config/jsonschema/1-0-0
```diff
@@ -0,0 +1,115 @@
...
+                                                        },
+                                                        "dependsOn": {
+                                                                "type": "array",
+                                                                "items": {
+                                                                        "type": "string",
+                                                                        "minLength": 1,
+                                                                        "maxLength": 64
+                                                                },
+                                                                "minItems": 0,
+                                                                "maxItems": 64
```
colmsnowplow, line 54: Same for maxItems in arrays - no need to impose the restriction in the schema if it doesn't exist in practice.

schemas/com.snowplowanalytics.datamodeling/config/jsonschema/1-0-0
```diff
@@ -0,0 +1,115 @@
...
+                                                        }
+                                                },
+                                                "required": [
+                                                        "playbook",
+                                                        "dependsOn"
+                                                ],
+                                                "additionalProperties": false
```
colmsnowplow, line 61: I would probably say additionalProperties = true. 

additionalProperties means that if false, sending additional keys will render the data invalid. If true, additional data can be present, but nothing will be done with it. I think the latter is a better descriptor of the situation (unless your validation script imposes it?)

schemas/com.snowplowanalytics.datamodeling/config/jsonschema/1-0-0
```diff
@@ -0,0 +1,115 @@
...
+                                },
+                                "dagName": {
+                                        "type": ["string", "null"],
+                                        "minLength": 1,
+                                        "maxLength": 64
+                                },
+                                "sqlRunner": {
+                                        "type": ["string", "null"],
+                                        "pattern": "^(0|[1-9][0-9]{0,1})\\.(0|[1-9][0-9]{0,1})\\.(0|[1-9][0-9]{0,1})(-rc[0-9]{1,2}){0,1}$",
+                                        "maxLength":13
```
colmsnowplow, line 98: I wouldn't impose a max length here either.

paulboocock: We've just had a chat on another PR about maxLengths. It seems that having a maxLength is preferred because it is more explicit (for example, in an event/entity schema how to create the Redshift table) even if that maxLength has to be a huge overestimation (up to Redshift max of 65k but probably something like 4096 is fine?).

adatzer:
Made a separate comment, because it refers to more than one from the above.
The schema reflects what the validation checks for. We can certainly change that there, but at the moment:
1. The validation logic both expects and checks for `minLength` and `maxLength` for all strings except for the ones with regex pattern (`sqlRunner`). 
2. The same with arrays, expects and checks for `minItems` and `maxItems`.
3. In the playbook object, it does check that there are no additional properties than `"playbook"` and `"dependsOn"`. So if additional properties, the config will be indeed invalid.

schemas/com.snowplowanalytics.datamodeling/config/jsonschema/1-0-0
colmsnowplow: Sure, I agree with that perspective from the point of view that when you're tracking data, there is a hard limitation somewhere (so for a string, redshift imposes a max length, and that should be represented in the schema).

When something genuinely doesn't have any such restriction it seems more wrong than right to specify one in the schema for me. Not something I care enough about to object strongly, so if we want to add a v long max length it's fine. Just philosophically seems odd.

colmsnowplow:
> The validation logic both expects and checks for minLength and maxLength for all strings except for the ones with regex pattern (sqlRunner).
> The same with arrays, expects and checks for minItems and maxItems.
> In the playbook object, it does check that there are no additional properties than "playbook" and "dependsOn". So if additional properties, the config will be indeed invalid.

OK that makes sense then. As long as we're setting the allowable limit to be quite high.

schemas/com.snowplowanalytics.datamodeling/config/jsonschema/1-0-0
paulboocock: Philosophically I also agree and think that as this is a config, it should probably be removed.

colmsnowplow:
Just to summarise for clarity as some convos happened offline.

- Since the consumer of the config does actually perform validation of the maxLength restrictions etc. (but doesn't use this schema to do so), then it does make most sense to reflect those in the schema. If we decide to change them we can release a new version of both script and schema - but we don't think we'll need to.

- The 'schema' 'data' keys should be removed - the factotum one was created this way a long time ago for reasons unknown right now - but it's the exception not the rule.

So changing the second point is all that needs to happen before this can go into a release I think. (plus creating an issue and rebase for 1:1:1 rule as per release guide) :)

paulboocock (approved):
LGTM

colmsnowplow (approved):

chuwy:

PR #1095, opened 2020-12-17
R119

chuwy:
Contains:

* ~One schema (1-0-0, no migrations)~
* ~3 patches - safe as no ordering or types have been changed, just `string` constraints which won't have any effect on existing tables~
* ~4 JSONPath for existing schemas (1 widely used, but was added after Redshift migrations) + 4 SQL files (same) + 1 SQL (`com.maichimp/unsubscribe_1.sql` - we had the DDL, but it contains a typo)~

Updated: since we couldn't release this before Travis CI goes down - we had to rework R119 into CI/CD-only release. Everything from original PR will be in R120


schemas/com.snowplowanalytics.snowplow.storage.rdbshredder/shredding_complete/jsonschema/1-0-0
```diff
@@ -0,0 +1,86 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for information about the shredded data, for loader to read",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow.storage.rdbshredder",
```
benjben, line 5: In the code you used `com.snowplowanalytics.snowplow.storage.rdbloader`. I think that loader is better if we consider that shredder is a subpart of loader.

schemas/com.snowplowanalytics.snowplow.storage.rdbshredder/shredding_complete/jsonschema/1-0-0
```diff
@@ -0,0 +1,86 @@
...
+    "name": "shredding_complete",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "base": {
+      "description": "Blob storage path to the root of the batch",
+      "type": "string",
+      "maxLength": 1024
```
benjben, line 15: We could add a check that it starts with `s3://`, `s3a://` or `s3n://`

schemas/com.snowplowanalytics.snowplow.storage.rdbshredder/shredding_complete/jsonschema/1-0-0
```diff
@@ -0,0 +1,86 @@
...
+    },
+    "compression": {
+      "description": "File compression type",
+      "enum": ["GZIP", "NONE"]
+    },
+    "types": {
+      "description": "Set of Iglu URIs and the format they were shredded into",
+      "type": "array",
```
benjben, line 23: We can add that the minimum length of this array is 1.

benjben (approved):
Just a couple of additional checks that we could add and the name of the `vendor` that we probably want to update, otherwise looks great :+1:

schemas/com.snowplowanalytics.snowplow.storage.rdbshredder/shredding_complete/jsonschema/1-0-0
chuwy: I think that `format: uri` is what we need here. At some point RDB Loader will be working with GCS and Snowflake.

schemas/com.snowplowanalytics.snowplow.storage.rdbshredder/shredding_complete/jsonschema/1-0-0
chuwy: Why? It's totally okay to not have any shredded types in a batch. Even more, it's totally okay to not have any _events_ in a batch.

schemas/com.snowplowanalytics.snowplow.storage.rdbshredder/shredding_complete/jsonschema/1-0-0
chuwy: I actually was thinking about that, but decided to go with shredder just because we already have one (unused) schema with this vendor. However, after thinking more, I agree that semantic meaning is more important than consistency with legacy schemas.

schemas/com.snowplowanalytics.snowplow.storage.rdbshredder/shredding_complete/jsonschema/1-0-0
benjben: I was thinking that shredder should not emit a message if it didn't process any data, but actually you're right, it's better to have a message that says 0 than not having a message, as not having a message can indicate a problem.

paulboocock (approved):
LGTM

chuwy:
Force-pushed to include #1096.

paulboocock (approved):

cksnp (approved):

chuwy:
Over to you, @jbeemster!

PR #1097, opened 2020-12-28
Migrate to GitHub Actions

oguzhanunlu:
Migration to GitHub Actions


oguzhanunlu:
@jbeemster , this is ready for review. Note that I left the top commit intentionally so that you can see how I tested CD. That commit is to be deleted once the review is done

Right now,
- for a commit, only lint is performed
- for a pr, both lint and mirror are performed
- for a tag, lint + mirror + deploy are performed

See [this run](https://github.com/snowplow/iglu-central/runs/1682041733?check_suite_focus=true) for an example continuous deployment. I deleted the example tag (`r-migration-test`) to keep the repository clean.

chuwy (approved):
Logic and commit look good to me. I'd just clarify that by "mirror" you mean "dev mirror" as we also have Iglu Server mirrors.

oguzhanunlu:
I just updated my comment above and updated the step name to clarify which mirror it is. Does that look good @chuwy ?

oguzhanunlu:
Just squashed dev mirror naming related changes and reworded commit message, hence the force push

jbeemster (approved):
All looks good to me!

One thing that might be nice to change is to have the logic of when to do a release live in a separate git action.  Rather than having to figure out via bash if it is a release tag you could just have CD done based on tags which would simplify your CI and CD flows.  We could not do this easily in Travis (hence these workarounds) but actions does let you logically split these apart.

Not a blocker though!

oguzhanunlu:
@chuwy @jbeemster I've revised the logic in this PR to address Josh's feedback on more modular approach.

- CI action runs `lint` and then `dev mirror` steps on 
  - any pull request
  - any push to `master` branch
The previous `on: [push, pull_request]` approach results in duplicate runs for each push to a PR, hence the optimization 
- CD action runs `deploy` step on tags that start with `r`

It is tested using dry run approach, you can check actions history to see how it behaves.

jbeemster:
LGTM!

chuwy:

PR #1101, opened 2021-02-08
R120

chuwy:


.github/scripts/constants.sh
```diff
@@ -1,8 +1,7 @@
 # Igluctl
 INSTALL_DIR="/tmp"
-VALIDATOR_V="0.7.1"
-VALIDATOR_ZIP="igluctl_${VALIDATOR_V}.zip"
-VALIDATOR_URI="http://dl.bintray.com/snowplow/snowplow-generic/${VALIDATOR_ZIP}"
+VALIDATOR_V="0.7.2"
+VALIDATOR_URI="https://github.com/snowplow-incubator/igluctl/releases/download/${VALIDATOR_V}/igluctl"
```
jbeemster, line 4: Woohoo no more Bintray!

jbeemster (approved):

paulboocock (approved):

cksnp (approved):

.github/scripts/requirements.sh
```diff
@@ -0,0 +1,16 @@
...
+set -e
+
+# Constants
+source $GITHUB_WORKSPACE/.github/scripts/constants.sh
+
+function install_igluctl() {
+    if [ ! -f "${IGLUCTL}" ]
+    then
+        echo "igluctl not found, downloading..."
+        wget "${VALIDATOR_URI}" -P "${INSTALL_DIR}"
```
jbeemster, line 11: Just to confirm this is downloading the JAR directly and we are certain it will be executable?  Would it be worth adding a `chmod +x igluctl` here just in case a binary gets uploaded that does not have the executable bit set?

chuwy: Hey @jbeemster! No, it won't be executable, we run it everywhere with `java -jar ${IGLUCTL}`, so we're certain it will execute it.

jbeemster: Easy done.

Issue #1102, opened 2021-02-08
Infrastructure: replace Travis badge with GA badge in README


oguzhanunlu:
After successful migration to GitHub Actions in R120, README still has the travis badge. We can replace that with another badge for GitHub Actions

PR #1103, opened 2021-02-08
R121

chuwy:
It has been reviewed in original PR that should have contained these schemas (https://github.com/snowplow/iglu-central/pull/1095), but ended up CI/CD only.

The only changed I adde is https://github.com/snowplow/iglu-central/issues/1102


paulboocock (approved):

cksnp (approved):

stdfalse (approved):
LGTM! It's a pity though that for the patches schemas appropriate tables in Redshift won't be widened automatically. @chuwy, do we have plans to address this in the loader?

chuwy:
> do we have plans to address this in the loader?

For patches specifically - no, unfortunately. We'd like to make patches less and less frequent and reserve for very narrow use cases. For additions however yes it's planned for near future.

I've updated the date in CHANGELOG, over to @jbeemster

PR #1104, opened 2021-02-15
Correct 'repository' references in bad rows schemas

miike:
There's a spelling error in a few of the bad rows schemas - the error exists in the schemas themselves but the upstream components have the correct spelling.

I'm not sure this is the best way to correct - by overriding the schema or if it's better to just go with a new model version for each of the schemas, though this would then result in having to update both the upstream and downstream dependencies.

For downstream dependencies I don't think there's many but https://github.com/snowplow-incubator/snowplow-badrows will be one of them.

cc @chuwy @benjben


chuwy:
Oh, good catch, @miike.

So this part of 

It's also strange because in snowplow-badrows we have property-based testing

Issue #1109, opened 2021-04-22
Add com.snowplowanalytics.mobile/remote_config/jsonschema/1-0-0


AlexBenny:
This is used in the remote configuration feature for the mobile trackers.
It's the definition of how the remote configuration has to be in order to be correctly processed by the tracker.
It doesn't represent data to send in the pipeline.

colmsnowplow:
@chuwy since this is a schema for a configuration object (which is used to configure the iOS/Android trackers themselves), I'm unsure if its rightful place is among the schemas for events and entities. Do you have an opinion on where in Iglu Central such a schema should live?

chuwy:
Hey @colmsnowplow! By "place among the schemas for events and entities" do you mean the whole Iglu Central or `com.snowplowanalytics.mobile` namespace? Either way I don't think we were ever trying to separate config and snowplow entities. And I think it's a good place. [com.snowplowanalytics.snowplow](https://github.com/snowplow/iglu-central/tree/master/schemas/com.snowplowanalytics.snowplow) might be a bad example, but it contains all possible schemas. I'm happy as long as we do at least a slight effort for not putting everything into that single vendor.

colmsnowplow:
I meant both really - just wanted a sanity check on how to organise any delineation between configs and tracking-related schemas.

If you're happy with how it's proposed, so am I! TY. :)

paulboocock:
```
"description": "The configuration file for the Snowplow mobile trackers.",
  "type": "object",
  "properties": {
    "configurationVersion": {
```

Do we need a configurationVersion? Doesn't the schema version give us that information already?

AlexBenny:
Sorry, I've noticed your comment just now, I guess I don't receive notifications from this repo (!?).

The configurationVersion is used by the tracker to check either it has to apply the downloaded configuration. If the version number is different it applies the changes otherwise it doesn't.

You are talking about the formatVersion that was redundant with the schema but I already removed that part when Anton raised a couple of comments in the PR.

PR #1118, opened 2021-06-14
Add mobile remote configuration schema

AlexBenny:
As specified in #1109 :
This is used in the remote configuration feature for the mobile trackers.
It's the definition of how the remote configuration has to be in order to be correctly processed by the tracker.
It doesn't represent data to send in the pipeline.

### DIFF
--- a/schemas/com.snowplowanalytics.mobile/remote_config/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.mobile/remote_config/jsonschema/1-0-0
@@ -1,0 +1,169 @@
+{
+  "$schema" : "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "self" : {
+    "vendor" : "com.snowplowanalytics.mobile",
+    "name" : "remote_config",
+    "version" : "1-0-0",
+    "format" : "jsonschema"
+  },
+  "title": "Snowplow Mobile Trackers config file",
+  "description": "The configuration file for the Snowplow mobile trackers.",
+  "type": "object",
+  "properties": {
+    "formatVersion": {
+      "description": "Version string (SemVer format) that identifies the current schema.",
+      "type": "string",
+      "pattern": "^[0-9]+\\.[0-9]+\\.[0-9a-zA-Z.+\\-]+$"
+    },
+    "configurationVersion": {
+      "description": "Version number that identifier the current configuration. It has to be increased on each update.",
+      "type": "integer"
+    },
+    "configurationBundle": {
+      "description": "The list of configurations for each tracker to configure.",
+      "type": "array",
+      "items": [
+        {
+          "type": "object",
+          "properties": {
+            "namespace": {
+              "description": "The namespace string that identifies the single tracker.",
+              "type": "string"
+            },
+            "networkConfiguration": {
+              "description": "Represents the network communication configuration allowing the tracker to be able to send events to the Snowplow collector.",
+              "type": "object",
+              "properties": {
+                "endpoint": {
+                  "description": "URL of the collector that is going to receive the events tracked by the tracker. The URL can include the schema/protocol (e.g.: http://collector-url.com). In case the URL doesn’t include the schema/protocol, the HTTPS protocol is automatically selected.",
+                  "type": "string"
+                },
+                "method": {
+                  "description": "The method used to send the requests (GET or POST).",
+                  "enum": [ "get", "post" ]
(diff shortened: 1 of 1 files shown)


AlexBenny:
Do I need to add "$schema" in the required fields?
It's needed in the tracker as it's used to check if the json received has the correct format.

schemas/com.snowplowanalytics.mobile/remote_config/jsonschema/1-0-0
```diff
@@ -0,0 +1,160 @@
...
+    "vendor" : "com.snowplowanalytics.mobile",
+    "name" : "remote_config",
+    "version" : "1-0-0",
+    "format" : "jsonschema"
+  },
+  "description": "The configuration file for the Snowplow mobile trackers.",
+  "type": "object",
+  "properties": {
+    "configurationVersion": {
+      "description": "Version number that identifier the current configuration. It has to be increased on each update.",
```
chuwy, line 13: Presumably we can add `minimum`?

chuwy (approved):
> Do I need to add "$schema" in the required fields?

Hey @AlexBenny! Not sure what you mean, but most likely - no, the schema looks good now!

schemas/com.snowplowanalytics.mobile/remote_config/jsonschema/1-0-0
AlexBenny: Technically it wouldn't need a minimum as we use it just to check if the configuration is changed.
We can eventually put it to 0 but there aren't technical reasons for that.

AlexBenny:
> Not sure what you mean, but most likely - no, the schema looks good now!

I just realised that the tracker requires the $schema field in the json data sent to the tracker in order to process it correctly.
In this case I haven't set it mandatory, so is the user creates the json data without specifying $schema the tracker wouldn't be able to process it, even if the json data is compliant with this schema.

Would it make sense if I force it as mandatory doing this?
```
...
    "required": [
        "$schema",
        "configurationVersion",
        "configurationBundle"
    ]
...
```

paulboocock:
`$schema` shouldn't be part of the schema definition. It's a keyword in JSON schema ([https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.8.1.1](https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.8.1.1)) and a core part of what every json document should contain ([http://json-schema.org/learn/getting-started-step-by-step.html#starting](http://json-schema.org/learn/getting-started-step-by-step.html#starting)). 

`$schema` should be used in json documents which should validate against a json schema definition, so it's fine to expect users to include it in their configuration files - the RFC states that it's implementation specific if it's not defined so I think it's fine to reject the file in the tracker if `$schema` is missing, even if it's not present in this schema definition.

schemas/com.snowplowanalytics.mobile/remote_config/jsonschema/1-0-0
paulboocock: You wouldn't want it to be a negative number though, so specifying a minimum probably still makes sense rather than omitting it.

AlexBenny: Why all this anger against negative numbers :D

schemas/com.snowplowanalytics.mobile/remote_config/jsonschema/1-0-0
```diff
@@ -0,0 +1,163 @@
...
+              "foregroundTimeout": {
+                "description": "The amount of time that can elapse before the session id is updated while the app is in the foreground.",
+                "type": "integer",
+                "minimum": 0
+              }
+            }
+          }
+        },
+        "required": [
+          "namespace"
```
paulboocock, line 154: Is `networkConfiguration` not required?

paulboocock (approved):
LGTM bar my one comment

schemas/com.snowplowanalytics.mobile/remote_config/jsonschema/1-0-0
AlexBenny: it's not required. If you don't set the networkConfiguration in the configuration bundle, the tracker will use the related namespace to remove the tracker instance if it exists. It's an implicit and rudimental way to clean instances.

paulboocock: In that case, all looks good!

PR #1119, opened 2021-06-16
Release R123

### LINKED ISSUE #1109: Add com.snowplowanalytics.mobile/remote_config/jsonschema/1-0-0

paulboocock:
A single schema required for Remote Configuration for the mobile trackers v2.1.0

Approved Schema PR: https://github.com/snowplow/iglu-central/pull/1118

### Changelog
Add com.snowplowanalytics.mobile/remote_config/jsonschema/1-0-0 (close #1109)

### DIFF
--- a/schemas/com.snowplowanalytics.mobile/remote_config/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.mobile/remote_config/jsonschema/1-0-0
@@ -1,0 +1,163 @@
+{
+  "$schema" : "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "self" : {
+    "vendor" : "com.snowplowanalytics.mobile",
+    "name" : "remote_config",
+    "version" : "1-0-0",
+    "format" : "jsonschema"
+  },
+  "description": "The configuration file for the Snowplow mobile trackers.",
+  "type": "object",
+  "properties": {
+    "configurationVersion": {
+      "description": "Version number that identifies the current configuration. It has to be increased on each update.",
+      "type": "integer",
+      "minimum": 0
+    },
+    "configurationBundle": {
+      "description": "The list of configurations for each tracker to configure.",
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "namespace": {
+            "description": "The namespace string that identifies the single tracker.",
+            "type": "string"
+          },
+          "networkConfiguration": {
+            "description": "Represents the network communication configuration allowing the tracker to be able to send events to the Snowplow collector.",
+            "type": "object",
+            "properties": {
+              "endpoint": {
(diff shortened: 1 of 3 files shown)


istreeter (approved):

schemas/com.snowplowanalytics.mobile/remote_config/jsonschema/1-0-0
```diff
@@ -0,0 +1,163 @@
...
+          "networkConfiguration": {
+            "description": "Represents the network communication configuration allowing the tracker to be able to send events to the Snowplow collector.",
+            "type": "object",
+            "properties": {
+              "endpoint": {
+                "description": "URL of the collector that is going to receive the events tracked by the tracker. The URL can include the schema/protocol (e.g.: http://collector-url.com). In case the URL doesn’t include the schema/protocol, the HTTPS protocol is automatically selected.",
+                "type": "string"
+              },
+              "method": {
+                "description": "The method used to send the requests (GET or POST).",
```
cksnp, line 36: Since this is not required, perhaps the default should be added in the description?

schemas/com.snowplowanalytics.mobile/remote_config/jsonschema/1-0-0
```diff
@@ -0,0 +1,163 @@
...
+              }
+            },
+            "required": [
+              "endpoint"
+            ]
+          },
+          "trackerConfiguration": {
+            "description": "Represents the configuration of the tracker and the core tracker properties. The TrackerConfiguration can be used to setup the tracker behaviour indicating what should be tracked in term of automatic tracking and contexts/entities to track with the events.",
+            "type": "object",
+            "properties": {
```
cksnp, line 47: Presumably none of these properties are required if `trackerConfiguration` exists?

cksnp (approved):
Two comments but I *think* none of them are blockers, so approving.

schemas/com.snowplowanalytics.mobile/remote_config/jsonschema/1-0-0
AlexBenny: I think it depends where we enforce the default value. At the moment the default value is set by the tracker. A different tracker version could change the default value without invalidate the correctness of the schema.
Same point would value for all the settings of the other configurations (Especially TrackerConfiguration). Unless we think that a new major version of the tracker (with breaking changes in the default values) would force an update of this schema too.

schemas/com.snowplowanalytics.mobile/remote_config/jsonschema/1-0-0
AlexBenny: yep. None of these are required.

PR #1125, opened 2021-07-20
Batch Alert

chuwy:
Very applied down-to-the-earth schema with some very hard decisions.

This would be used in https://github.com/snowplow/snowplow-rdb-loader/issues/458 for webhook-based alerting when RDB Loader fails to load a folder.

On vendor/name:

* Initially I thought to make it [`com.snowplowanalytics.snowplow.storage/rdbloader_alert`](https://github.com/snowplow/iglu-central/tree/master/schemas/com.snowplowanalytics.snowplow.storage). There are mostly configs in that namespace, but also important `shredding_complete` which is used in the RDB Loader
* But then realised we have [com.snowplowanalytics.monitoring.batch](https://github.com/snowplow/iglu-central/tree/master/schemas/com.snowplowanalytics.monitoring.batch) and RDB Loader already [produces](https://github.com/snowplow/snowplow-rdb-loader/blob/ad7e781e3eccf6e0bc29180fd96c066f85b8e6dc/modules/loader/src/main/scala/com/snowplowanalytics/snowplow/rdbloader/dsl/Monitoring.scala#L52-L53) events with those schemas, just as Snowplow events (that we're not using for a long time), not as webhook payloads. I also think it makes sense to call it "batch" because RDB Loader is essentially a batch loader. The only problem with that namespace is that beside of this new schema it's mostly dead and unused
* Another option on name would be `loader_alert` - I don't think we plan to issue it from any other apps


paulboocock:
nit:  Commit is `Add com.snowplowanalytics.snowplow.monitoring.batch/alert/jsonschema/1-0-0` but should be `Add com.snowplowanalytics.monitoring.batch/alert/jsonschema/1-0-0`

chuwy:
Nice catch!

schemas/com.snowplowanalytics.monitoring.batch/alert/jsonschema/1-0-0
```diff
@@ -0,0 +1,40 @@
...
+    "vendor": "com.snowplowanalytics.monitoring.batch",
+    "name": "alert",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "folder": {
+      "description": "A folder that caused an alert",
+      "type": "string",
```
oguzhanunlu, line 14: If `folder` is optional, I'd expect `["string", "null"]` here, wdyt?

dilyand:
> Another option on name would be `loader_alert` - I don't think we plan to issue it from any other apps

What if loaders get merged into a single app down the line?

Btw, this more general name will probably also need a `target` name, if the same app can load into multiple data WHs.

***

On the namespace, we also have https://github.com/snowplow-proservices/com.snowplowanalytics-schema-registry/tree/master/schemas/com.snowplowanalytics.monitor.storage and https://github.com/snowplow-proservices/com.snowplowanalytics-schema-registry/tree/master/schemas/com.snowplowanalytics.monitor

schemas/com.snowplowanalytics.monitoring.batch/alert/jsonschema/1-0-0
```diff
@@ -0,0 +1,40 @@
...
+      "format": "uri",
+      "maxLength": 1024
+    },
+    "version": {
+      "type": "string",
+      "description": "Application's version"
```
oguzhanunlu, line 20: just a nit: an uninitiated person would need to check top description to understand that it is rdb loader being referred by `Application`

schemas/com.snowplowanalytics.monitoring.batch/alert/jsonschema/1-0-0
oguzhanunlu: I noticed that `folder` isn't required hence optional, is there a strategy on picking this approach vs making it accept `"null"` ?

istreeter:
The `shredding_complete` schema uses ["base" instead of folder](https://github.com/snowplow/iglu-central/blob/fa04d40df5ff7d2527f29a6e64d3b5ab24c5d325/schemas/com.snowplowanalytics.snowplow.storage/shredding_complete/jsonschema/1-0-1#L12).  Do you agree these are the same thing, or do they represent something slightly different?  Should we be consistent in naming across schemas?

Instead of `version`, how about having a `processor` block to match what we have in bad rows?  It address the comment that `version` could be misunderstood, it addresses the concern about whether this schema could be used by another artifact, and it makes us more consistent across schemas.

chuwy:
> Do you agree these are the same thing, or do they represent something slightly different?

Funny, I had `base` at some point (when schema was very different), then made a decision to rename and cannot remember the reason. But generally yes, I agree they're the same thing and loader joins tables on this property.

> Instead of version, how about having a processor block to match what we have in bad rows?

I'm less sure about this one (cc @oguzhanunlu, @dilyand). I think this alert is specificially about RDB Loader (soon to be called Batch Loader). I don't foresee it to be used with for example Shredder or BigQuery Loader. On the other hand, I agree that it costs nothing.

schemas/com.snowplowanalytics.monitoring.batch/alert/jsonschema/1-0-0
chuwy: Yeah, you're right. I expressed optionality by not mentioning it in `required` because circe encoders by default just skip the field, but I agree it's better to be explicit.

dilyand:
> I think this alert is specificially about RDB Loader (soon to be called Batch Loader). I don't foresee it to be used with for example Shredder or BigQuery Loader. On the other hand, I agree that it costs nothing.

If it costs nothing, then I think more general gives us more flexibility, while more specific gives us less.

benjben:
> I think this alert is specificially about RDB Loader

This schema seems very generic. Let's say that tomorrow we want to add webhook alerting to enrich, we could very well reuse the same (if we use `processor` as Ian suggested). Any reason to tie it up to RDB loader and not make it for all apps ?

chuwy:
Just in case - I'm all for adding an app name. As Dilyan pointed out - if it costs nothing it's worth doing. Below is just for discussion.

> Only folder is tied to RDB loader, but given that this is optional we could imagine putting it as a tag.

I'm very much against this. `folder` (or `base`) when present is super important. It signals that a particular folder has a problem and I definitely can imagine pipeline operators querying it / grouping by it.

Also `folder` is what makes the alert specific to _Batch_ Loader, i.e. it operates on folder/batches, not records, not streams. This also a reason why I tend to call it `application` or similar rather than `processor` as in bad rows. `processor` produces bad rows because it processes records/events, whereas here it's just a plain application.

istreeter:
I will add one one more reason why `folder` (or `base`) should not be a tag.  The tags should be configured entirely by the user in the configuration of the application.  I think we could keep this distinction.... tags are user-configured, other fields are application-set.  It makes things messy if the application adds a tag.

I am swayed by your argument about `processor` being something specific and not appropriate here.

chuwy:
Closing in favor of #1127

PR #1127, opened 2021-08-23
R126

chuwy:


schemas/com.snowplowanalytics.monitoring.batch/alert/jsonschema/1-0-0
```diff
@@ -0,0 +1,47 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Webhook payload describing an alert coming from RDB Loader",
+  "self": {
+    "vendor": "com.snowplowanalytics.monitoring.batch",
```
paulboocock, line 5: What does `.batch` represent on the vendor? Is it a schema specific to batch processing with Snowplow? What would a similar alert from the BigQuery Loader be called?
I'm also not sure what properties of this schema make it particularly "batch-y"? Why not `com.snowplowanalytics.monitoring`?

paulboocock (commented):
Looks good, just one question on the vendor name.

schemas/com.snowplowanalytics.monitoring.batch/alert/jsonschema/1-0-0
chuwy: It's all about `folder`!

In future we see the current RDB and Snowflake loaders to be merged into a single Batch Loader, where architetecture is inherently batch-y: one component (shredder or transformer) prepares the data in one folder (batch of data), another component (loader) copies it into warehouse.

chuwy: But in fact I'd lie to you if I told you it was an easy decision. I don't like it very much, but most other options included creation of a new namespace, which I also don't like.

paulboocock: Hardest thing in computer science, right? ;-)

paulboocock (approved):

istreeter (approved):

PR #1132, opened 2021-09-13
Add com.snowplowanalytics.mobile/application_lifecycle/jsonschema/1-0-0 (close #1131)

### LINKED ISSUE #1131: Add com.snowplowanalytics.mobile/application_lifecycle/jsonschema/1-0-0
The native mobile trackers v2.3 will introduce a new entity automatically tracked by the tracker.
All the events will be added with this entity that indicates whether or not the event has been tracked when the app was visible to the user (foreground state rather than background state).

AlexBenny:
As explained in the related issue, here is the schema for application_lifecycle entity.
The entity is attached to all the events by the tracker based on the app state (foreground or background).

For reference iOS PR: snowplow/snowplow-objc-tracker#637

### DIFF
--- a/schemas/com.snowplowanalytics.mobile/application_lifecycle/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.mobile/application_lifecycle/jsonschema/1-0-0
@@ -12,8 +12,7 @@
   "properties": {
     "isVisible": {
       "description": "Indicates if the app is in foreground state (true) or background state (false)",
-      "type": "boolean",
-      "default": true
+      "type": "boolean"
     },
     "index": {
       "description": "Represents the foreground index or background index (tracked with com.snowplowanalytics.snowplow application_foreground and application_background events.",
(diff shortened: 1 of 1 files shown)


schemas/com.snowplowanalytics.mobile/application_lifecycle/jsonschema/1-0-0
```diff
@@ -0,0 +1,27 @@
...
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+
+  "type": "object",
+  "properties": {
+    "isVisible": {
+      "description": "Indicates if the app is in foreground state (true) or background state (false)",
+      "type": "boolean",
+      "default": true
```
chuwy, line 16: I certainly don't have anything against `default`, but just in case - it won't be used anywhere (Iglu doesn't support it explicitly). Thus tables will be generated as `NOT NULL` and if user doesn't provide it - the loading will fail. Some more context:

* https://github.com/snowplow/iglu/issues/248
* https://github.com/snowplow/iglu-scala-client/issues/76

We can use it as purely information, although I afraid it encourages bad behavior.

chuwy (changes requested):
Happy to give a green tick if you think `default` is justified.

schemas/com.snowplowanalytics.mobile/application_lifecycle/jsonschema/1-0-0
AlexBenny: I agree 👍  I don't even know where I got it. That value is always set in the tracker so no need to specify this here.

chuwy (approved):
LGTM! I'll squash the commits in final R127 PR.

chuwy:
Closing in favor of https://github.com/snowplow/iglu-central/pull/1139

PR #1134, opened 2021-09-15
Add iglu:com.snowplowanalytics.oss/oss_context/jsonschema/1-0-1 (close #1133)

### LINKED ISSUE #1133: Add com.snowplowanalytics.oss/oss_context/jsonschema/1-0-1
Application telemetry has a requirement to track instances with a unique ID, as per https://github.com/snowplow/stream-collector/issues/167 .

The schema should be handling both terraform deployment telemetry and application telemetry.

voropaevp:
Creating PR to make schema available for testing via the `http://iglucentral-dev.com.s3-website-us-east-1.amazonaws.com`.


istreeter:
I want to put some explanation here for benefit of reviewers.  There were two different ways this schema could have been prepared:

*Option 1*: Make it completely consistent with [the tracking that is already implemented in the open source terraform modules](https://github.com/snowplow-devops/terraform-snowplow-telemetry/blob/v0.2.0/main.tf#L14).  This would mean calling it `com.snowplowanalytics.oss/oss_context`.  Note the schema for the terraform tracking was put in a private iglu server, not iglu-central, but now we want to make it publicly visible.

The advantage of option 1 is that we (Snowplow Analytics) do not have to change any of our data modelling jobs, and all events will continue to load into the same table as before.

To make option 1 work, I think we would have to release the original schema as `1-0-0` first, and then release the new schema as `1-0-1` because it adds nullable fields compared to the schema in the private repo.  We cannot release the new one as `1-0-0` because it would break our pipeline.

*Option 2*: Re-design the schema, and do not try to make it consistent with the original tracking.

The advantages of option 2 are:
- We can keep the convention of putting snowplowy schemas in a `com.snowplowanalytics.snowplow.*` vendor.
- Opportunity to rename some fields if we want to, and schema name if we want to.

The disadvantages of option 2 are that we (Snowplow Analytics) would have telemetry events landing in a different table, which is slightly inconvenient for us.  And we would need to release a new version of oss terraform modules using the new schema name.

chuwy:
My _very_ humble opinion here: I think we should be after idealistic Option 2. If the only advantage of Option 1 is data landing to the same table - we always have ways to make that happen even with Option 2, although in a hacky way: recovery job or JS enrichment. It's more of a problem for existing data, but presumably there's not much of it collected so far.

schemas/com.snowplowanalytics.oss/oss_context/jsonschema/1-0-0
```diff
@@ -0,0 +1,94 @@
...
+      ],
+      "maxLength": 256,
+      "description": "A unique identifier for the VM instance. Unique for each instance of the app running within a module"
+    },
+    "appGeneratedId": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "pattern": "[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}",
```
chuwy, line 57: I'd recommend `"format": "uuid": https://json-schema.org/understanding-json-schema/reference/string.html#resource-identifiers

That way some DB will be able to to deduce `UUID` type if they support it

schemas/com.snowplowanalytics.oss/oss_context/jsonschema/1-0-0
```diff
@@ -0,0 +1,94 @@
...
+      "description": "Automatically generated UUID on every service restart"
+    },
+    "cloud": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "maxLength": 256,
+      "description": "The cloud AWS or GCP, none for on-premise"
```
chuwy, line 66: Could be `"enum": ["aws", "gcp", "azure", null]`?

If we go the `enum` way - we still should preserve nullable `string` type, because Iglu does not always infer the right type for enums.

schemas/com.snowplowanalytics.oss/oss_context/jsonschema/1-0-0
```diff
@@ -0,0 +1,94 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Telemetry schema",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow.telemetry",
+    "name": "heartbeat",
```
chuwy, line 6: Metadata must match the filepath.

chuwy (changes requested):
1. I left some comments on the name - I think `heartbeat` is preferable, but don't want to block. This isn't a strong opinion, but regardless - filename and metadata must match
2. Two comments that I left: presumably if we stick with `oss_context` - we want to keep them as is. If we go after a completely new schema - I'd suggest to use `format` and `enum`

schemas/com.snowplowanalytics.oss/oss_context/jsonschema/1-0-0
```diff
@@ -0,0 +1,94 @@
...
+        "string",
+        "null"
+      ],
+      "maxLength": 256,
+      "description": "The region the services are running in"
+    },
+    "applicationName": {
+      "type": [
+        "string",
+        "null"
```
benjben, line 79: It will always be set

schemas/com.snowplowanalytics.oss/oss_context/jsonschema/1-0-0
```diff
@@ -0,0 +1,94 @@
...
+      ],
+      "maxLength": 256,
+      "description": "The name of the application as supplied by the terraform script OR reflected from the build"
+    },
+    "applicationVersion": {
+      "type": [
+        "string",
+        "null"
```
benjben, line 87: It will always be set

voropaevp:
Adding both the current version used by OSS terraform modules and the updated version and the updated version that will be used for application telemetry.

voropaevp:
Moving the previous schema version to the separate PR.

chuwy:
Cherry-picked into https://github.com/snowplow/iglu-central/pull/1139. Closing.

PR #1139, opened 2021-09-20
R127

chuwy:
The release:

* Adds two new mobile schemas
* Makes `com.snowplowanalytics.oss/oss_context/jsonschema/1-0-0` public
* Adds `com.snowplowanalytics.oss/oss_context/jsonschema/1-0-1`
* Patches `com.snowplowanalytics.monitoring.batch/alert/jsonschema/1-0-0` - the patch is safe because schema could not be used with warehouses


istreeter (approved):

istreeter:
This is not critical, but would be nice to also bump igluctl to 0.8.0 in the github action.

paulboocock (approved):

chuwy:
> This is not critical, but would be nice to also bump igluctl to 0.8.0 in the github action.

Discussed this with Ian and decided to do another igluctl-only release later.

schemas/com.snowplowanalytics.mobile/deep_link/jsonschema/1-0-0
```diff
@@ -0,0 +1,27 @@
...
+	"type": "object",
+	"properties": {
+		"url": {
+			"type": "string",
+			"description": "URL in the received deep-link",
+			"format": "uri",
+			"maxLength": 4096
+		},
+		"referrer": {
+			"type": "string",
```
jbeemster, line 19: This property is not required - should it be nullable in that case?

schemas/com.snowplowanalytics.mobile/application_lifecycle/jsonschema/1-0-0
```diff
@@ -0,0 +1,26 @@
...
+  },
+
+  "type": "object",
+  "properties": {
+    "isVisible": {
+      "description": "Indicates if the app is in foreground state (true) or background state (false)",
+      "type": "boolean"
+    },
+    "index": {
+      "description": "Represents the foreground index or background index (tracked with com.snowplowanalytics.snowplow application_foreground and application_background events.",
```
jbeemster, line 18: This property is not required - should it be nullable?

schemas/com.snowplowanalytics.mobile/deep_link_received/jsonschema/1-0-0
```diff
@@ -0,0 +1,27 @@
...
+	},
+	"type": "object",
+	"properties": {
+		"url": {
+			"type": "string",
+			"description": "URL in the received deep-link",
+			"format": "uri",
+			"maxLength": 4096
+		},
+		"referrer": {
```
jbeemster, line 18: This property is not required - should it be nullable?

schemas/com.snowplowanalytics.monitoring.batch/alert/jsonschema/1-0-0
```diff
@@ -36,10 +36,7 @@
     },
     "tags": {
       "description": "Set of key value pairs providing additional information",
-      "type": "object",
-      "additionalProperties": {
-        "type": "string"
-      }
+      "type": "object"
```
jbeemster, line 39: @chuwy where is this schema used?  Assuming this cannot break any load processes / schema evolution through this removal?

schemas/com.snowplowanalytics.oss/oss_context/jsonschema/1-0-0
```diff
@@ -0,0 +1,59 @@
...
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Context schema for OSS terraform modules, VM instances and application events",
+  "self": {
+    "vendor": "com.snowplowanalytics.oss",
+    "name": "oss_context",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
```
jbeemster, line 11: This schema is not setting any required properties - seems to be something we do on all other schemas?

schemas/com.snowplowanalytics.oss/oss_context/jsonschema/1-0-1
```diff
@@ -0,0 +1,64 @@
...
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Context schema for OSS terraform modules, VM instances and application events",
+  "self": {
+    "vendor": "com.snowplowanalytics.oss",
+    "name": "oss_context",
+    "format": "jsonschema",
+    "version": "1-0-1"
+  },
+  "type": "object",
+  "properties": {
```
jbeemster, line 11: This schema is not setting any required properties - seems to be something we do on all other schemas?

schemas/com.snowplowanalytics.monitoring.batch/alert/jsonschema/1-0-0
jbeemster: Saw the comment on the PR - no issues.

chuwy:
Hey @jbeemster,

Regarding all your comments about `required` and/or `null`. For downstream consumers (Loaders) there's no difference between: `{"foo": 42, "bar": null}` and `{"foo": 42}`. In both cases `bar` will be a nullable column.

For schemas there are two ways though to express optionality and it mostly depends on the producers themselves. The rule of thumb that I use to decide between "explicit null" and "missing key" is:

1. If it's a context sent over the network - use "missing key" approach, it can save some bites
2. If it's a some kind of configuration schema - use "explicit key", it will make templating easier and increase readability

Otherwise - rely on producer's preference.

PR #1169, opened 2021-09-21
Add properties for memory, battery, and storage info into mobile context (close #1168)

### LINKED ISSUE #1168: Add properties for memory, battery, and storage info into mobile context
To be able to track more information about mobile devices as context along with events, we need additional properties in the schema. We can extend the existing mobile_context schema.

So far, we identified the following properties:

* memory
   * physicalMemory – Total physical system memory in bytes
   * systemAvailableMemory – Available memory on the system in bytes (Android only)
   * appAvailableMemory – Amount of memory in bytes available to the current app (iOS only)
* battery
   * batteryLevel – Remaining battery level as an integer percentage of total battery capacity
   * batteryState – Battery state for the device
   * lowPowerMode – A Boolean indicating whether Low Power Mode is enabled (iOS only)
* storage
   * availableStorage – Bytes of storage remaining
   * totalStorage – Total size of storage in bytes

matus-tomlein:
* Extended the mobile_context schema with new properties
* Added "null" types for existing optional properties
* Bumped mobile_context to version 1.0.2

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-2
+++ b/schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-2
@@ -1,0 +1,95 @@
+{
+	"$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+	"description": "Schema for mobile contexts",
+	"self": {
+		"vendor": "com.snowplowanalytics.snowplow",
+		"name": "mobile_context",
+		"format": "jsonschema",
+		"version": "1-0-2"
+	},
+	"type": "object",
+	"properties": {
+		"osType": {
+			"type": "string"
+		},
+		"osVersion": {
+			"type": "string"
+		},
+		"deviceManufacturer": {
+			"type": "string"
+		},
+		"deviceModel": {
+			"type": "string"
+		},
+		"carrier": {
+			"type": ["string", "null"]
+		},
+		"networkType": {
+			"type": ["string", "null"],
+			"enum": ["mobile", "wifi", "offline"]
+		},
+		"networkTechnology": {
+			"type": ["string", "null"]
+		},
+		"openIdfa": {
+			"type": ["string", "null"]
+		},
+		"appleIdfa": {
+			"type": ["string", "null"]
+		},
+		"appleIdfv": {
+			"type": ["string", "null"]
+		},
+		"androidIdfa": {
+			"type": ["string", "null"]
+		},
+		"physicalMemory": {
+			"type": ["integer", "null"],
+			"minimum": 0,
+			"maximum": 9223372036854775807,
+			"description": "Total physical system memory in bytes"
+		},
+		"systemAvailableMemory": {
+			"type": ["integer", "null"],
+			"minimum": 0,
+			"maximum": 9223372036854775807,
+			"description": "Available memory on the system in bytes (Android only)"
+		},
+		"appAvailableMemory": {
+			"type": ["integer", "null"],
+			"minimum": 0,
+			"maximum": 4294967295,
+			"description": "Amount of memory in bytes available to the current app (iOS only)"
+		},
+		"batteryLevel": {
+			"type": ["integer", "null"],
+			"minimum": 0,
+			"maximum": 100,
+			"description": "Remaining battery level as an integer percentage of total battery capacity"
+		},
+		"batteryState": {
+			"type": ["string", "null"],
+			"enum": ["unplugged", "charging", "full"],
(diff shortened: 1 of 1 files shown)


schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-2
```diff
@@ -0,0 +1,90 @@
...
+		},
+		"carrier": {
+			"type": ["string", "null"]
+		},
+		"networkType": {
+			"type": ["string", "null"],
+			"enum": ["mobile", "wifi", "offline"]
+		},
+		"networkTechnology": {
+			"type": ["string", "null"]
```
AlexBenny, line 32: This field (and some others) was `"type": "string"`. I wonder if the added "null" option can be a breaking change.
Happy to have a feedback on this by the experts.

schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-2
```diff
@@ -0,0 +1,90 @@
...
+			"type": ["integer", "null"],
+			"minimum": 0,
+			"description": "Total physical system memory in bytes"
+		},
+		"systemAvailableMemory": {
+			"type": ["integer", "null"],
+			"minimum": 0,
+			"description": "Available memory on the system in bytes (Android only)"
+		},
+		"appAvailableMemory": {
```
AlexBenny, line 56: I wonder if it would make sense to add a prefix in the name to state clearly that this field is platform specific. Something like `iOSAppAvailableMemory`. I don't think we have done anything like that before but it may help to avoid confusion when the data is extracted. Thoughts?

schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-2
```diff
@@ -0,0 +1,90 @@
...
+		},
+		"appleIdfv": {
+			"type": ["string", "null"]
+		},
+		"androidIdfa": {
+			"type": ["string", "null"]
+		},
+		"physicalMemory": {
+			"type": ["integer", "null"],
+			"minimum": 0,
```
AlexBenny, line 48: As you asked on Slack... 
> the properties I added for memory size in bytes can get quite high, is that useful in that case?

Honestly, I don't know. I don't think I always specified the integer max value. I see many different styles in the other schemas. Btw, I guess that specify a max value wouldn't hurt.

AlexBenny (approved):
Looks great to me!
I've just added few notes that I would like to double check with the experts (@chuwy, @paulboocock).
I realised I don't know much of the iglu schema policies.

schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-2
matus-tomlein: If we get some experts look at this, please also tell us what is the default behavior in case no max value is specified? Some of the memory sizes in bytes are represented using unsigned 64 bit integers on the device, so we should be able to support the same range of values.

schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-2
matus-tomlein: That's a good point, I think it would be clearer. The counter argument is that maybe in the future we will be able to get the system available memory on iOS and app available memory on Android which would make it difficult to reuse these properties.

schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-2
paulboocock: No harm in adding a huge maximum from my point of view. Better to be explicit.

schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-2
paulboocock: This isn't a breaking change, you are widening the allowed values so that's fine. Even if this was in the `required` array at the bottom in 1-0-1 then I think that'd be fine as you'd be going from a NOT NULL to NULL (although I'm happy to be corrected on this, but I think we're good here).

paulboocock (approved):

schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-2
matus-tomlein: Added max values for 64 bit or 32 bit integers based on what we are getting on device.

PR #1175, opened 2021-10-11
Add version tag Metabase schema

ranquild:
At Metabase, we need to include `tag` with every snowplow event. This PR adds a schema for the tag.

### DIFF
--- a/schemas/com.metabase/tag/jsonschema/1-0-0
+++ b/schemas/com.metabase/tag/jsonschema/1-0-0
@@ -7,7 +7,6 @@
         "format": "jsonschema",
         "version": "1-0-0"
     },
-
     "type": "object",
     "properties": {
         "tag": {


ranquild:
> Hey @ranquild Could you link me to some public documentation that shows the use case for this schema? We only accept Schemas into Iglu Central which benefit the wider userbase, for example: a webhooks from a third party system. I'm not really sure how this new `tag` schema would be used, unless it's something specific to the Metabase product that others would use? If this is a schema for your own use, then hosting it in your own [Iglu Repository](https://docs.snowplowanalytics.com/docs/pipeline-components-and-applications/iglu/iglu-repositories) is the right way to go (and you don't need to wait for our review and release process for Iglu Central).

Hi @paulboocock, thanks for feedback! As Metabase is an open source product, the schema would be used by all our users that enabled anonymous tracking. It doesn't make sense for us to create a private Iglu repo as people won't be able to validate their data with the schema.

PR #1188, opened 2021-11-15
Release/r130

paulboocock:
This release adds schemas for three areas:

- New schemas for Web Video Tracking. We've introduced a new "central" Snowplow schema for common properties and then platform specific schemas for additional context data (like HTML5 Video Elements, and Youtube iFrame API).
- A new Identify event which many platforms want users to send to them (Such as Iterable, which is one of our new GTM SS destinations). This will be documented as part of our recommended tracking integration for when using Iterable via GTM SS. It's been modelled on other platforms who have a similar "Identify" API.
- A fix for YAUAA string lengths where some events are failing as the YAUAA enrichment is generating values longer than this schema allows.


schemas/org.youtube/youtube/jsonschema/1-0-0
```diff
@@ -0,0 +1,143 @@
...
+      "minimum": 0,
+      "maximum": 65535
+    },
+    "unstarted": {
+      "type": "boolean",
+      "description": "If the player hasn't started"
+    },
+    "url": {
+      "type": "string",
+      "description": "The YouTube embed URL of the media resource",
```
istreeter, line 91: format uri?

istreeter (commented):
I don't have any objections at all to these schemas.  I wrote down a few thoughts that occurred, but I don't know any context around media tracking, so it's mainly curiosity.

schemas/org.whatwg/media_element/jsonschema/1-0-0
```diff
@@ -0,0 +1,274 @@
...
+      "type": [
+        "string",
+        "null"
+      ],
+      "description": "CORS settings value of the media player",
+      "maxLength": 255
+    },
+    "currentSource": {
+      "type": "string",
+      "description": "The absolute URL of the media resource",
```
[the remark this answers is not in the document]
greg-el: Looks like it should always be a URL: https://html.spec.whatwg.org/multipage/media.html#concept-media-load-algorithm
It gets assigned equal to `urlString` (an instance of: https://url.spec.whatwg.org/#concept-url)

schemas/org.whatwg/media_element/jsonschema/1-0-0
```diff
@@ -0,0 +1,274 @@
...
+    "defaultPlaybackRate": {
+      "type": "number",
+      "description": "The default media playback rate of the player",
+      "minimum": -9007199254740991,
+      "maximum": 9007199254740991
+    },
+    "disableRemotePlayback": {
+      "type": [
+        "boolean",
+        "null"
```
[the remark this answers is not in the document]
greg-el: `disableRemotePlayback` isn't supported by all browsers: https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/disableRemotePlayback
On browsers that don't support it, it doesn't exist on the HTMLMediaElement instance.
This is probably the reason for the other nullable booleans too, but I'll check each to make sure.

schemas/com.snowplowanalytics.snowplow/media_player/jsonschema/1-0-0
```diff
@@ -0,0 +1,76 @@
...
+      "type": "number",
+      "description": "Playback rate (1 is normal)",
+      "minimum": -9007199254740991,
+      "maximum": 9007199254740991
+    },
+    "volume": {
+      "type": "number",
+      "description": "Volume level",
+      "minimum": 0,
+      "maximum": 1
```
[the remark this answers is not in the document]
greg-el: Volume has been changed to 0-100 for consistency.

schemas/nl.basjes/yauaa_context/jsonschema/1-0-3
```diff
@@ -0,0 +1,231 @@
...
+            "type": "string",
+            "maxLength": 100
+        },
+        "agentVersion": {
+            "type": "string",
+            "maxLength": 100
+        },
+        "agentVersionMajor": {
+            "type": "string",
+            "maxLength": 100
```
jbeemster, line 106: Is this not technically a breaking change for SchemaVer?  Not sure Redshift for example can handle changing the length of a field without a major version bump?

istreeter: > Is this not technically a breaking change for SchemaVer?

No, because all historic data is compatible with the new schema, and so [according to this blog post](https://snowplowanalytics.com/blog/2014/05/13/introducing-schemaver-for-semantic-versioning-of-schemas/) that makes it an `addition` (final number).

> Not sure Redshift for example can handle changing the length of a field without a major version bump?

That used to be true, but [this issue in schema-ddl](https://github.com/snowplow/schema-ddl/issues/83) addressed that problem.  Starting from RDB loader 1.2.x we can alter the column length.  Same too for Postgres Loader > 0.3.0.

schemas/com.snowplowanalytics.snowplow/media_player/jsonschema/1-0-0
```diff
@@ -0,0 +1,76 @@
...
+      "type": "boolean",
+      "description": "If the media element is muted"
+    },
+    "paused": {
+      "type": "boolean",
+      "description": "If the media element is paused"
+    },
+    "percentProgress": {
+      "type": [
+        "integer",
```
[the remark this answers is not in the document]
greg-el: This was intentional, it's part of the "progress" event where a user can pass in an array of per cent values i.e. `[10, 20, 50, 90]` to fire an event when reached. Having decimal values felt unnecessary for this.

schemas/nl.basjes/yauaa_context/jsonschema/1-0-3
istreeter: Oh, my mistake.... looks like Postgres Loader does not handle this yet, although Redshift loader does.  I will need to rethink this.

schemas/org.whatwg/media_element/jsonschema/1-0-0
```diff
@@ -0,0 +1,274 @@
...
+        "additionalProperties": false
+      }
+    },
+    "seeking": {
+      "type": "boolean",
+      "description": "If the media is in the process of seeking to a new position"
+    },
+    "src": {
+      "type": "string",
+      "description": "The 'src' HTML attribute of the media element",
```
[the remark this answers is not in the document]
greg-el: If I'm understanding this correctly, wouldn't that result in the whole of the media being sent with every event?

schemas/org.youtube/youtube/jsonschema/1-0-0
greg-el: It will always be a URL (as it will be a YouTube.com link).
https://developers.google.com/youtube/iframe_api_reference#Retrieving_video_information

schemas/nl.basjes/yauaa_context/jsonschema/1-0-3
jbeemster: Isn't that only true assuming as well that everyone using this enrichment and that updates enrich is _also_ using the latest RDB Loader which they might well not be.  If they are using an older loader version as well won't this break as it will find all the schema paths and try to push for an upgrade regardless?

istreeter: The important thing is that we can provide an upgrade path to open source users.  In this case, the path is to upgrade RDB loader to >1.2.0 and then upgrade enrich to 3.0.0 (released soon).  Alternatively, an open source user may choose to use an older version of RDB loader, and enrich 2.x.

Sooner or later though, we need to start taking advantage of the feature that our modern loaders can alter column max lengths.

schemas/org.whatwg/media_element/jsonschema/1-0-0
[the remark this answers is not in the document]
istreeter: Yes.  That's why I asked if it's possible.  I'm not saying it's a problem, I'm just wondering what would happen in this edge case.

schemas/org.whatwg/media_element/jsonschema/1-0-0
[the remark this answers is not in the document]
istreeter: I meant, you could use the `"format": "uri"` annotation, and then Enrich will validate that it is always a valid uri.  There's [an example here](https://github.com/snowplow/iglu-central/blob/master/schemas/com.clearbit.enrichment/person/jsonschema/1-0-0#L16).  It's just an idea, and it doesn't make a big difference, except that it makes certain guarantees to downstream processors that the uri has been validated.

schemas/org.whatwg/media_element/jsonschema/1-0-0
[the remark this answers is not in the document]
istreeter: I mean, I'm interested in what the tracker would do - would it send the data?  I don't have an objection to the schema.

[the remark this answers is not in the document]
paulboocock: Thats a good point. We should probably protect against this happening in the JS Tracker to ensure we don't send data urls and then a reasonable max length here for a typical http url should be fine. I'm not sure we'd want to be sending huge data url's as part of an event 😬

schemas/nl.basjes/yauaa_context/jsonschema/1-0-3
paulboocock: I have a vague recollection that we've previously increased the length of a string before, in a patch release, with the knowledge that the event would still fail to load in some warehouses but would at least now pass validation (and load into some more flexible like BigQuery). You'd also have to upgrade your enrich version (v3) to start using this new schema, and with that we could caveat that you also need to be on RDB/PG Loader x.x.x too.

paulboocock: We could also make this 2-0-0 since it'd be introduced with Enrich v3... more of a challenging upgrade path for all users when we do that from a data modelling point of view though...

schemas/com.snowplowanalytics.snowplow/identify/jsonschema/1-0-0
```diff
@@ -0,0 +1,25 @@
...
+                "id": {
+                        "description": "The user identifier.",
+                        "type": ["string", "null"],
+                        "maxLength": 256
+                },
+                "email": {
+                        "description": "The user email address.",
+                        "type": ["string", "null"],
+                        "maxLength": 256
+                }
```
cksnp, line 21: I assume both `id` and `email` may be null because someone may send one *or* the other, but in that case isn't `email` an `id` really? In which case, you'd only need a non-null `id` and the value of it would be an email address.

schemas/com.snowplowanalytics.snowplow/media_player/jsonschema/1-0-0
```diff
@@ -0,0 +1,76 @@
...
+      "description": "The percent of the way through the media",
+      "minimum": 0,
+      "maximum": 100
+    },
+    "playbackRate": {
+      "type": "number",
+      "description": "Playback rate (1 is normal)",
+      "minimum": -9007199254740991,
+      "maximum": 9007199254740991
+    },
```
cksnp, line 57: Just thinking out loud, would an event with a `playbackRate` of `9007199254740991` (or `-9007199254740991`) actually be valid in any conceivable scenario?

schemas/org.whatwg/media_element/jsonschema/1-0-0
```diff
@@ -0,0 +1,274 @@
...
+            "minimum": 0,
+            "type": "number"
+          },
+          "end": {
+            "description": "The end of the time range",
+            "maximum": 9007199254740991,
+            "minimum": 0,
+            "type": "number"
+          }
+        },
```
cksnp, line 34: Is a 0-length duration valid if the lower end is not `0`? E.g., is it valid to have a buffered duration from time unit `10` to time unit `10`?

schemas/org.whatwg/media_element/jsonschema/1-0-0
```diff
@@ -0,0 +1,274 @@
...
+    "defaultMuted": {
+      "type": "boolean",
+      "description": "If audio is muted by default"
+    },
+    "defaultPlaybackRate": {
+      "type": "number",
+      "description": "The default media playback rate of the player",
+      "minimum": -9007199254740991,
+      "maximum": 9007199254740991
+    },
```
cksnp, line 65: Same question as above

schemas/org.whatwg/media_element/jsonschema/1-0-0
```diff
@@ -0,0 +1,274 @@
...
+      ],
+      "description": "An array of time ranges played",
+      "items": {
+        "type": "object",
+        "description": "A time range",
+        "properties": {
+          "start": {
+            "type": "number",
+            "description": "The beginning of the time range",
+            "minimum": -9007199254740991,
```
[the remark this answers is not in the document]
cksnp: +1 on Ian's comment, shouldn't `minimum` be `0` for both `start` and `end`

schemas/org.youtube/youtube/jsonschema/1-0-0
```diff
@@ -0,0 +1,143 @@
...
+  "type": "object",
+  "properties": {
+    "avaliablePlaybackRates": {
+      "type": "array",
+      "description": "An array of playback rates in which the current video is available",
+      "items": {
+        "type": "number",
+        "minimum": 0,
+        "maximum": 9007199254740991
+      }
```
cksnp, line 19: Playback rates in previous schemas can have negative values too

cksnp (commented):
Some questions mostly, could be because I'm unfamiliar with how these video players operate.

schemas/com.snowplowanalytics.snowplow/identify/jsonschema/1-0-0
paulboocock: I'd be inclinded to agree with you, however many platforms split out these two fields when we pass this data to them and they behave differently depending on if you give them an email address or an id. With marketing platforms, once you give them an e-mail address they start to build a user profile, whereas because an id could be an email or could be a uuid they don't make the same assumptions. Iterable is a good example, where you can pass an e-mail and/or a userId: https://api.iterable.com/api/docs#events_track

schemas/com.snowplowanalytics.snowplow/media_player/jsonschema/1-0-0
paulboocock: Not really... but there's nothing in your browser from stopping you from trying, so it's technical a valid value. It's pretty wild that negative values are even valid, what would one expect to happen... the video to play backwards?

https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/playbackRate

schemas/org.whatwg/media_element/jsonschema/1-0-0
```diff
@@ -0,0 +1,274 @@
...
+      "enum": [
+        "aac",
+        "aacp",
+        "caf",
+        "flac",
+        "mp3",
+        "mp4",
+        "ogg",
+        "wav",
+        "webm"
```
paulboocock, line 95: I think we might want to loosen this, what happens when a new format comes along? We'll have to update this schema and I don't think this enum check adds too much to this.

schemas/com.snowplowanalytics.snowplow/media_player/jsonschema/1-0-0
cksnp: Just to explain where I'm coming from on this, I'm thinking of programming mistakes where the rate is e.g. 125 instead of 1.25, leading to (what is in fact) bad events ending up in atomic tables. But I can also see how you're basically trying to be aligned with what the browsers allow so I'll resolve this.

schemas/org.whatwg/media_element/jsonschema/1-0-0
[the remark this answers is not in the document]
greg-el: Sorry @istreeter, I misunderstood, just created a fix in the tracker that will prevent data URLs from being sent in 'src'.

schemas/org.whatwg/media_element/jsonschema/1-0-0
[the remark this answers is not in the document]
greg-el: This has been added.

schemas/org.whatwg/media_element/jsonschema/1-0-0
greg-el: Enum removed.

schemas/org.whatwg/media_element/jsonschema/1-0-0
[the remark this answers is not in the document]
greg-el: I have corrected this.

schemas/org.youtube/youtube/jsonschema/1-0-0
greg-el: These are the rates the YouTube player provides in the dropdown menu on the UI. It actually looks like the range is currently 0.25 - 2 (https://youtube-eng.googleblog.com/2017/09/variable-speed-playback-on-mobile.html )(I can't find any newer source on the range, but I tried some videos and it still appears to be the case), so the minimum can probably be changed to `0.25`, and maximum reduced to `2`? (or a bit higher in case YouTube decides to allow higher values in future.

schemas/org.whatwg/media_element/jsonschema/1-0-0
greg-el: It looks like 0-length durations are valid:
["The start of a range must be less than or equal to the end of that same range."](https://html.spec.whatwg.org/multipage/media.html#timeranges)

schemas/org.youtube/youtube/jsonschema/1-0-0
paulboocock: I think leaving the minimum at 0 does no harm. The max could change if the Youtube API won't ever respond with a bigger number. Maybe going for something like 64 or 128 is a safe bet.

greg-el: The maximum has been changed to 128.

schemas/org.whatwg/media_element/jsonschema/1-0-0
```diff
@@ -0,0 +1,265 @@
...
+        },
+        "additionalProperties": false
+      }
+    },
+    "seeking": {
+      "type": "boolean",
+      "description": "If the media is in the process of seeking to a new position"
+    },
+    "src": {
+      "type": "string",
```
paulboocock, line 198: I think `src` can be null if using:

```
  <video id="playerid" playsinline controls>
    <source src="test-video.mp4" type="video/mp4" />
  </video>
```

`currentSource`, or currentSrc in the case of the HTML API, likely won't ever be null as far as I can tell, so that should probably be the required field

schemas/com.snowplowanalytics.snowplow/media_player/jsonschema/1-0-0
```diff
@@ -0,0 +1,76 @@
...
+    "currentTime": {
+      "type": "number",
+      "description": "The current playback time",
+      "minimum": 0,
+      "maximum": 9007199254740991
+    },
+    "duration": {
+      "type": [
+        "number",
+        "null"
```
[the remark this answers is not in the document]
greg-el: Yeah, it's reported as `+Infinity` for a stream by JavaScript, it does feel like it might be missing a bit of context in the case of a live video by leaving it as null now that you mention it. Perhaps a `isLive` or `isStream` bool property could be added to be more explicit?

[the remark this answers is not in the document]
paulboocock: `duration: null, isStream: true` is a nice solution (assuming duration being null is the only property you need to look at to know it's a stream).

schemas/org.whatwg/media_element/jsonschema/1-0-0
greg-el: 'src' has been removed from required

schemas/com.snowplowanalytics.snowplow/media_player/jsonschema/1-0-0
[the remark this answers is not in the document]
greg-el: 'isLive' has been added

schemas/nl.basjes/yauaa_context/jsonschema/1-0-3
paulboocock: Resolving based on this discussion: https://snowplow.slack.com/archives/GPFKE9CG1/p1637582919008300

schemas/com.snowplowanalytics.snowplow/identify/jsonschema/1-0-0
```diff
@@ -0,0 +1,25 @@
...
+        },
+        "description": "A Snowplow identify event.",
+        "type": "object",
+        "properties": {
+                "id": {
+                        "description": "The user identifier.",
+                        "type": ["string", "null"],
+                        "maxLength": 256
+                },
+                "email": {
```
istreeter, line 17: I considered suggesting to put a regex on the email field, so that downstream consumers of this entity know it has been validated already.  But email regexes are difficult, and so it might not solve any real-life problem.

istreeter (approved):
Looks good.  All previous comments have been addressed.

schemas/com.youtube/youtube/jsonschema/1-0-0
```diff
@@ -0,0 +1,158 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Context Schema for a youtube player event",
+  "self": {
+    "vendor": "com.youtube",
```
chuwy, line 5: Doesn't match commit/ticket name (where it's still `org.`)

schemas/com.snowplowanalytics.snowplow/media_player/jsonschema/1-0-0
```diff
@@ -0,0 +1,80 @@
...
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow",
+    "name": "media_player",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "currentTime": {
+      "type": "number",
```
chuwy, line 13: Could it be floating point?

schemas/com.youtube/youtube/jsonschema/1-0-0
```diff
@@ -0,0 +1,158 @@
...
+      "description": "A string of the latest error to occur, or null if no errors",
+      "enum": [
+        "INVALID_PARAMETER",
+        "HTML5_PLAYER_ERROR",
+        "NOT_FOUND",
+        "EMBED_DISALLOWED"
+      ]
+    },
+    "loaded": {
+      "type": "number",
```
chuwy, line 64: Seems other percentage properties in other schemas are `integer`. Presumably this one is an exception?

chuwy (changes requested):
Looks good! Just one blocker - the inconsistency between commit and schema vendor. Happy to green-tick once this is fixed.

Another one is a suggestion (as I know nothing about the domain of the problem) - are we sure we use `number` and those properties are always floating point and not integers?

And a point about `com.snowplowanalytics.snowplow` vendor. It _seems_ media player schemas indeed look general (or specific to Snowplow, the platform) enough to be included there, but I'm wondering if other options were condidered? The reason I'm asking is simply that we historically were using it as a default for many things and it got a bit overcrowded, so if there are other schemas that can share a namespace - I'd think of something else. But that is very minor.

schemas/com.snowplowanalytics.snowplow/media_player/jsonschema/1-0-0
paulboocock: Yeah, it's a double in the spec we've based this off: https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/currentTime

> A double-precision floating-point value indicating the current playback time in seconds.

paulboocock:
> And a point about com.snowplowanalytics.snowplow vendor. It seems media player schemas indeed look general (or specific to Snowplow, the platform) enough to be included there, but I'm wondering if other options were condidered? The reason I'm asking is simply that we historically were using it as a default for many things and it got a bit overcrowded, so if there are other schemas that can share a namespace - I'd think of something else. But that is very minor.

We did consider other options, but I'm not sure where else it would live. The goal of the the "generic" schemas is for them to be applicable to all video playing tracking that is being done by Snowplow, so we put it in the `snowplow` vendor namespace. We could perhaps make it `com.snowplowanalytics.media` if that'd be preferrable?

schemas/com.youtube/youtube/jsonschema/1-0-0
greg-el: Good catch, changed to `integer`.

chuwy (approved):
👍

jbeemster (approved):

PR #1191, opened 2021-11-26
Add com.roku/video/jsonschema/1-0-0 (close #1189)

### LINKED ISSUE #1189: Add com.roku/video/jsonschema/1-0-0
This schema is for an entity that will be automatically added to media_player_event events (added in release r130, currently in PR) as context on the Roku tracker. It gives additional information about video playback.

It should more or less reflect the properties of the [Video player object on Roku](https://developer.roku.com/en-gb/docs/references/scenegraph/media-playback-nodes/video.md). The following properties were selected:

- videoId
- contentTitle
- contentId
- contentType
- streamUrl
- videoFormat
- isUnderrun
- isResumed
- videoFormat
- measuredBitrate
- streamBitrate
- timeToStartStreaming
- width
- height

matus-tomlein:
As explained in issue #1189 this PR adds schema for context added to media events sent from the Roku tracker. It adds extra information provided by the Roku Video player.


snowplowcla:

schemas/com.roku/video/jsonschema/1-0-0
```diff
@@ -0,0 +1,115 @@
...
+      "maxLength": 255
+    },
+    "streamFormat": {
+      "description": "Type of content (e.g., mp4, wma, mkv).",
+      "type": [
+        "string",
+        "null"
+      ],
+      "maxLength": 255
+    },
```
paulboocock, line 48: Whats the difference between these two properties? Can't figure it out from the descrition of `contentType`.

schemas/com.roku/video/jsonschema/1-0-0
```diff
@@ -0,0 +1,115 @@
...
+      "description": "Indicates whether the playback was resumed after trickplay."
+    },
+    "videoFormat": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "description": "Format of the currently playing video stream (e.g., hevc, mpeg2).",
+      "maxLength": 65535
+    },
```
paulboocock, line 85: So streamFormat is the container and videoFormat is the codec?

schemas/com.roku/video/jsonschema/1-0-0
```diff
@@ -0,0 +1,115 @@
...
+      "maxLength": 255
+    },
+    "contentId": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "description": "Played video content ID.",
+      "maxLength": 255
+    },
```
paulboocock, line 24: Whats the difference between video id and content id? Curious why this is optional too?

schemas/com.roku/video/jsonschema/1-0-0
```diff
@@ -0,0 +1,115 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Context Schema for a Roku video player event",
```
paulboocock, line 3: I wonder if linking to some Roku documentation in the description might be useful to clarify some of the fields in here?

paulboocock (commented):
Looks good. Couple of comments.

schemas/com.roku/video/jsonschema/1-0-0
matus-tomlein: Video ID is generated by the tracker when tracking for the video is initialized and remains the same for all events originating from the video. Content ID may be added by developers when setting content to be played on the video player. But it is not a required property for the video player (it is just additional metadata).

schemas/com.roku/video/jsonschema/1-0-0
matus-tomlein: That's a good idea! I added a link to the Video node documentation that this is based on.

schemas/com.roku/video/jsonschema/1-0-0
matus-tomlein: This is really confusing, the official documentation isn't very helpful (the descriptions I used were from there) so it's up to interpretation. This is how I interpret it (your container vs codec explanation was close):

- streamFormat is the container format for the video. It comes from content metadata that is passed to the player during initialization.
- videoFormat is the codec of the played stream. It is automatically inferred by the video player.
- contentType is the category of played content. It can be one of audio, episode, movie, season, series. It is passed in the content metadata during video initialization.

I tried to update the descriptions a bit, please let me know if it is understandable now.

schemas/com.roku/video/jsonschema/1-0-0
matus-tomlein: Yes, this is true. There is also a difference in that videoFormat is automatically inferred while streamFormat is passed in content metadata to the video player node.

matus-tomlein:
I had to make a small change in the schema in order to support error handling in the video playback. The way that error handling on videos works is that they are assigned an `error` state and the details of the error are described in the `errorStr` property. I had to add the optional `errorStr` property to this schema.

Also, in the error, there is not stream information and so I had to make the stream info (such as `streamUrl` and `measuredBitrate`) optional here too.

On the other hand, I added a `contentUrl` property which refers to the URL passed to video when initialized – this one is available even in error state so it can be useful to provide more context.

schemas/com.roku/video/jsonschema/1-0-0
```diff
@@ -0,0 +1,143 @@
...
+      "maximum": 65535
+    },
+    "errorStr": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "description": "A diagnostic message indicating a video play error. Refer to the Roku Video documentation for the format of the string.",
+      "maxLength": 65535
+    }
```
paulboocock, line 135: How does one know that this property will be populated? What sort of event would this entity be attached to where the error string would be populated?

schemas/com.roku/video/jsonschema/1-0-0
paulboocock: Looks good! The examples help.

schemas/com.roku/video/jsonschema/1-0-0
matus-tomlein: The entity would be attached to `media_player_event` with `event_type` equal to `"error"`. So one can use this as additional information for the event type. But if you have some other suggestions, I'm open to it. I can also add an error code property (int) to this entity if that helps checking for errors (but in theory the error code should also be inside this `errorStr`).

paulboocock: No, I think thats fine. Was mainly curious what `media_player_event` would look like.

paulboocock (approved):

PR #1192, opened 2021-11-26
Add com.roku/device_info/jsonschema/1-0-0 (close #1190)

### LINKED ISSUE #1190: Add com.roku/device_info/jsonschema/1-0-0
This schema is for an entity that will be automatically added to events as context on the Roku tracker. It gives basic information about the device, network, memory, display and more. It is quite similar to the [mobile_context entity](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-2) on mobile trackers.

It should reflect the properties of the [ifDeviceInfo object](https://developer.roku.com/en-gb/docs/references/brightscript/interfaces/ifdeviceinfo.md) available on the Roku SDK. I selected the following properties to add here:

- model
- modelDisplayName
- modelType
- osVersion
- channelClientId
- isRIDADisabled
- RIDA
- captionsMode
- audioOutputChannel
- memoryLevel
- timeSinceLastKeypress
- userCountryCode
- countryCode
- videoMode
- displayWidth
- displayHeight
- displayProperties
- connectionType
- internetStatus
- features

matus-tomlein:
As explained in issue #1990 this PR adds a schema for device context that will be automatically added to all events on the Roku tracker.

### DIFF
--- a/schemas/com.roku/device_info/jsonschema/1-0-0
+++ b/schemas/com.roku/device_info/jsonschema/1-0-0
@@ -1,0 +1,159 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Context Schema with information about Roku device (reflects the ifDeviceInfo interface: https://developer.roku.com/en-gb/docs/references/brightscript/interfaces/ifdeviceinfo.md)",
+  "self": {
+    "format": "jsonschema",
+    "name": "device_info",
+    "vendor": "com.roku",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "model": {
+      "type": "string",
+      "description": "Model name of the Roku device.",
+      "maxLength": 255
+    },
+    "modelDisplayName": {
+      "type": "string",
+      "description": "Model display name.",
+      "maxLength": 65535
+    },
+    "modelType": {
+      "type": "string",
+      "description": "Type of device.",
+      "maxLength": 255
+    },
+    "osVersion": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "description": "Major, minor, revision, and build number of the Roku OS running on the device separated by dots.",
+      "maxLength": 255
+    },
+    "channelClientId": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "description": "A unique device identifier that is different across channels.",
+      "maxLength": 255
+    },
+    "isRIDADisabled": {
+      "type": [
+        "boolean",
+        "null"
+      ],
+      "description": "Indicates whether tracking via Roku's ID for Advertisers (RIDA) is disabled on the device."
+    },
+    "RIDA": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "description": "A persistent unique identifier (UUID) for the device.",
+      "maxLength": 255
+    },
+    "captionsMode": {
+      "type": "string",
+      "description": "Determines whether global captions are turned on or off, or are in instant replay mode.",
+      "maxLength": 255
+    },
+    "audioOutputChannel": {
+      "type": "string",
+      "description": "Type of audio output.",
+      "maxLength": 255
+    },
+    "memoryLevel": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "description": "General memory level of the channel.",
+      "maxLength": 255
+    },
+    "timeSinceLastKeypress": {
+      "type": "integer",
+      "description": "The number of seconds since the last remote keypress was received.",
+      "minimum": 0,
+      "maximum": 2147483647
+    },
+    "userCountryCode": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "description": "ISO 3166-1 (2-letter) country code associated with the user's Roku account.",
+      "maxLength": 255
+    },
+    "countryCode": {
+      "type": "string",
+      "description": "Roku Channel Store associated with a user's Roku account.",
+      "maxLength": 255
+    },
+    "videoMode": {
+      "type": "string",
+      "description": "Video playback resolution.",
+      "maxLength": 255
+    },
+    "displayWidth": {
+      "type": "integer",
+      "description": "Physical width of the attached display in centimeters.",
+      "minimum": 0,
+      "maximum": 65535
+    },
+    "displayHeight": {
+      "type": "integer",
+      "description": "Physical height of the attached display in centimeters.",
+      "minimum": 0,
+      "maximum": 65535
+    },
+    "displayProperties": {
+      "type": "array",
+      "description": "List of keys for display properties of the screen.",
+      "items": {
+        "type": "string",
+        "description": "Display property of the screen.",
+        "maxLength": 255
+      }
+    },
+    "connectionType": {
+      "type": "string",
+      "description": "Type of internet connection the device is using.",
+      "maxLength": 255
+    },
+    "internetStatus": {
+      "type": [
+        "boolean",
+        "null"
+      ],
+      "description": "Internet connection status of the device."
+    },
+    "features": {
+      "type": "array",
+      "description": "List of features that the current device/firmware supports.",
+      "items": {
+        "type": "string",
+        "description": "Feature that the current device/firmware supports.",
+        "maxLength": 255
+      }
+    }
+  },
+  "required": [
+    "model",
+    "modelDisplayName",
+    "modelType",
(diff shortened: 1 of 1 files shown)


snowplowcla:

schemas/com.roku/device_info/jsonschema/1-0-0
```diff
@@ -0,0 +1,159 @@
...
+      "type": "string",
+      "description": "Type of device.",
+      "maxLength": 255
+    },
+    "osVersion": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "description": "Major, minor, revision, and build number of the Roku OS running on the device separated by dots.",
```
paulboocock, line 32: Could add some regex here if we wanted to, although we'd have to be certain Roku won't suddenly change their os versioning schema so maybe not worth the risk.

schemas/com.roku/device_info/jsonschema/1-0-0
```diff
@@ -0,0 +1,159 @@
...
+    "modelType": {
+      "type": "string",
+      "description": "Type of device.",
+      "maxLength": 255
+    },
+    "osVersion": {
+      "type": [
+        "string",
+        "null"
+      ],
```
paulboocock, line 31: Why optional? Surely there's always an osVersion?

schemas/com.roku/device_info/jsonschema/1-0-0
```diff
@@ -0,0 +1,159 @@
...
+      "description": "Indicates whether tracking via Roku's ID for Advertisers (RIDA) is disabled on the device."
+    },
+    "RIDA": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "description": "A persistent unique identifier (UUID) for the device.",
+      "maxLength": 255
+    },
```
paulboocock, line 57: Just a though: This should probably be optional in the tracker too, sort of like IDFA and such. Can make it a configarable flag like `captureRIDA` or something similar.

schemas/com.roku/device_info/jsonschema/1-0-0
```diff
@@ -0,0 +1,159 @@
...
+    "countryCode": {
+      "type": "string",
+      "description": "Roku Channel Store associated with a user's Roku account.",
+      "maxLength": 255
+    },
+    "videoMode": {
+      "type": "string",
+      "description": "Video playback resolution.",
+      "maxLength": 255
+    },
```
paulboocock, line 99: `videoMode` doesn't sound like a field I'd expect to find the resolution in. Maybe worth remapping to something more descriptive?

schemas/com.roku/device_info/jsonschema/1-0-0
```diff
@@ -0,0 +1,159 @@
...
+      "description": "Type of internet connection the device is using.",
+      "maxLength": 255
+    },
+    "internetStatus": {
+      "type": [
+        "boolean",
+        "null"
+      ],
+      "description": "Internet connection status of the device."
+    },
```
paulboocock, line 132: Since this is a boolean, maybe `isInternetConnected` is a more descriptive name? (or similar, internetStatus sounds like an enum to me)

schemas/com.roku/device_info/jsonschema/1-0-0
```diff
@@ -0,0 +1,159 @@
...
+    "features": {
+      "type": "array",
+      "description": "List of features that the current device/firmware supports.",
+      "items": {
+        "type": "string",
+        "description": "Feature that the current device/firmware supports.",
+        "maxLength": 255
+      }
```
paulboocock, line 140: Is this useful information? Is it a huge list? Pondering if its worth tracking. Could also be a flag in the tracker if it a big ol' list that not everyone will want.

paulboocock (commented):
Looking good. Nice entity. Couple of comments.

schemas/com.roku/device_info/jsonschema/1-0-0
matus-tomlein: Actually, not always because the SDK was only introduced in Roku OS 9.2 🤷‍♂️  There are some deprecated calls that give this information in a slightly different format but I didn't think it was worth the effort to add implement that – it requires some string parsing to find the OS version, cumbersome but possible to implement. What do you suggest?

schemas/com.roku/device_info/jsonschema/1-0-0
matus-tomlein: I think that's a good idea. There shouldn't be much risk in Roku changing their versioning schema because we control how this version string looks – the system call gives me a dictionary with keys for major, minor, revision and build number and I construct this version string in tracker code. I added a pattern regex to check for it.

schemas/com.roku/device_info/jsonschema/1-0-0
matus-tomlein: This is exactly how I implemented it in the tracker :) I discussed this with @AlexBenny and we came to the same conclusion to make the RIDA tracking configurable. If it is disabled, these properties will be null.

schemas/com.roku/device_info/jsonschema/1-0-0
matus-tomlein: This is a bit of a dilemma for me, because so far I tried to maintain the same names for fields as are used in the Roku SDK. However, many times (also in the Video schema PR) they are not easily understandable. However, the advantage is that one can easily map the properties to the Roku SDK and better understand where they come from. This schema is very similar to the ifDeviceInfo object on Roku, that's why I also used com.roku as the vendor. What do you suggest? Should I stick to the original names or come up with more descriptive ones?

schemas/com.roku/device_info/jsonschema/1-0-0
matus-tomlein: For my Roku, it only contains 1 item. It can get up to 10 items if the device supports everything. In the Roku SDK, these are implemented as individual booleans, but I thought putting them all into a list would be more compact and easier to extend in the future. I actually don't know if they are worth tracking, it's difficult for me to assess, but these are the features that it can include: 5.1_surround_sound, can_output_5.1_surround_sound, sd_only_hardware, usb_hardware, sdcard_hardware, ethernet_hardware, gaming_hardware, energy_star_compliant, soundbar_hardware, handsfree_voice

I would be okay with it if we decide not to include them at this time and wait for feedback if they are useful to anyone. The same goes for the displayProperties list.

schemas/com.roku/device_info/jsonschema/1-0-0
paulboocock: No, I'm fine if this is missing for older OSs 👍

schemas/com.roku/device_info/jsonschema/1-0-0
paulboocock: Lets keep the original names then, we can always blame Roku for the terrible names! At least the description here makes it easy to understand.

schemas/com.roku/device_info/jsonschema/1-0-0
paulboocock: I guess the same thing applies here regarding the name. Fine to leave it as it is.

schemas/com.roku/device_info/jsonschema/1-0-0
paulboocock: Lets included them if there's only a selection of them. Likely interesting metadata for people building video applications for Roku about what features are available.

paulboocock (approved):

schemas/com.roku/device_info/jsonschema/1-0-0
paulboocock: The more I look at this, the more nervous it makes me. What if someone does this with a beta release or something, could that introduce non-numeric values? I'm now feeling like we should remove this regex, not sure it adds to much value vs the risk it introduces to cause all events to fail.

paulboocock (commented):
Looking good, I think one last comment where I've changed me mind 😓

schemas/com.roku/device_info/jsonschema/1-0-0
matus-tomlein: That's probably right, better to be cautious about this – I removed the regex pattern from the osVersion property.

paulboocock (approved):

chuwy:
Cherry-picked into https://github.com/snowplow/iglu-central/pull/1196. @AlexBenny, if you're going to review this schema you can do this in the release PR.

PR #1196, opened 2021-12-09
R131

chuwy:


schemas/com.roku/device_info/jsonschema/1-0-0
```diff
@@ -0,0 +1,159 @@
...
+      "type": "string",
+      "description": "Model name of the Roku device.",
+      "maxLength": 255
+    },
+    "modelDisplayName": {
+      "type": "string",
+      "description": "Model display name.",
+      "maxLength": 65535
+    },
+    "modelType": {
```
cksnp, line 22: Without having looked at the reflected spec, should this be an enum?

schemas/com.roku/device_info/jsonschema/1-0-0
```diff
@@ -0,0 +1,159 @@
...
+    },
+    "RIDA": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "description": "A persistent unique identifier (UUID) for the device.",
+      "maxLength": 255
+    },
+    "captionsMode": {
```
cksnp, line 58: Maybe this is also better served by an enum?

cksnp (approved):
A couple of small comments, but overall I think it looks good.

schemas/com.roku/device_info/jsonschema/1-0-0
matus-tomlein: I think you're right, ideally it should be an enum. However, Roku releases new device models quite often – won't it be a problem that the schema will need to be updated for each new model? We would have to release new schema versions and also update the tracker to use the new schema versions.

schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/2-0-0
```diff
@@ -0,0 +1,131 @@
...
+  "self": {
+    "vendor": "com.snowplowanalytics.monitoring.batch",
+    "name": "load_succeeded",
+    "format": "jsonschema",
+    "version": "2-0-0"
+  },
+
+  "type": "object",
+  "properties": {
+    "shredding": {
```
istreeter, line 13: @chuwy I guess you use the word `shredding`to match `shredding_complete`.  When is the right time to change these schemas to `transforming`?

schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/2-0-0
```diff
@@ -0,0 +1,131 @@
...
+              "format": {
+                "description": "File format in which the entities were shredded to",
+                "enum": ["TSV", "JSON"]
+              }
+            },
+            "required": ["schemaKey", "format"]
+          }
+        },
+        "timestamps": {
+          "description": "Set of timestampts associated with the batch",
```
istreeter, line 46: timestampts spelling

schemas/com.roku/device_info/jsonschema/1-0-0
matus-tomlein: Ideally yes, it should be an enum. But it's not actually under our control – the value is taken from the Roku SDK and they can add or change the values in future releases. Because of that, I am a bit wary of listing explicit values in the schema. What do you think?

schemas/com.roku/device_info/jsonschema/1-0-0
matus-tomlein: The same answer as before – is it okay to enumerate the values if we do not control how they are produced?

schemas/com.roku/device_info/jsonschema/1-0-0
```diff
@@ -0,0 +1,159 @@
...
+      "minimum": 0,
+      "maximum": 2147483647
+    },
+    "userCountryCode": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "description": "ISO 3166-1 (2-letter) country code associated with the user's Roku account.",
+      "maxLength": 255
```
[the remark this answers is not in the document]
matus-tomlein: The problem is that although normally the string should be 2 characters long, there are cases when it can be longer – according to the Roku documentation, for some specific channels (apps) that are published in some weird way that I don't really understand, the string can also be for example "Sky Germany", or "Econet". They promise to only return the 2-char long code in the future, but currently it is not guaranteed to be 2-char long.

schemas/com.roku/device_info/jsonschema/1-0-0
```diff
@@ -0,0 +1,159 @@
...
+    },
+    "countryCode": {
+      "type": "string",
+      "description": "Roku Channel Store associated with a user's Roku account.",
```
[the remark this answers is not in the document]
matus-tomlein: I agree that this is really confusing and doesn't match. But actually the name of the property is wrong and since I was trying to follow the same naming as the Roku SDK, I left it that way. Channel store is usually a country code but in some cases it can be something different (same as with `userCountryCode` above). My understanding is that the `userCountryCode` will in future Roku SDKs always be the 2-char country code and this `countryCode` will remain the channel store identifier (sometimes country code, sometimes something else). That's why I thought that both are useful. I admit that it's a mess, but don't know how to make it better.

schemas/com.roku/video/jsonschema/1-0-0
```diff
@@ -0,0 +1,143 @@
...
+      ],
+      "description": "Indicates whether the playback was resumed after trickplay."
+    },
+    "videoFormat": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "description": "Video codec of the currently playing video stream (e.g., hevc, mpeg2, mpeg4_15).",
+      "maxLength": 65535
```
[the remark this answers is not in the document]
matus-tomlein: I actually should have used 255, good catch! I will update this.

schemas/com.roku/device_info/jsonschema/1-0-0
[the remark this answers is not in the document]
matus-tomlein: Changed the description a bit, hopefully it's more understandable now.

schemas/com.roku/video/jsonschema/1-0-0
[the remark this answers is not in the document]
matus-tomlein: Fixed

schemas/com.roku/device_info/jsonschema/1-0-0
paulboocock: This equally makes me nervous. It's not even a particularly well defined and peer reviewed spec like a web standard might be, we could easily end up in a situation with failing events that are completely out of our control.

schemas/com.roku/device_info/jsonschema/1-0-0
cksnp: I go the impression this is a simple tri-state? If there are more options then you're right it's better as it is.

matus-tomlein: Yeah, but even that is not absolutely clear to me from the documentation – at one point they mention a 4th option "When mute". This is also something that they could change in the future, for example by adding a new feature in Roku to display captions under different conditions, so I still feel like we're not totally safe enumerating the options.

schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/2-0-0
chuwy: That's a good question. I _think_ it's when we start using "tarnsformer" out in the wild - right now it's just our plans and the change could be more confusing than helping. Perhaps it would be better to do in one big bang release: remove all references to "shredder" and use new schemas.

One good counter-point would be that if we change it now - data will remain in the same table for longer time, but to me it doesn't seem strong enough.

Anyways, if you think now is the time - I can change that.

chuwy:
Hey @paulboocock, @istreeter - I addressed some review items and it seems @matus-tomlein did too - could you please re-review it again.

schemas/com.roku/device_info/jsonschema/1-0-0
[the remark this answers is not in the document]
istreeter: Hey @matus-tomlein do you think the description should then be changed?  Currently it says `ISO 3166-1 (2-letter)`.  Something like `Country code associated with the user's Roku account, normally the ISO 3166-1 (2-letter) code.`

schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/2-0-0
istreeter: No strong opinion on this - was just giving you the opportunity to consider it.  I'll resolve this conversation.

istreeter (commented):
Happy to approve this without any further changes, but I will just let @matus-tomlein answer my final question first.

paulboocock (approved):
LGTM pending Ian's outstanding question, I've nothing else to add.

schemas/com.roku/device_info/jsonschema/1-0-0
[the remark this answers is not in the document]
matus-tomlein: Changed, thanks for the suggestion!

istreeter (approved):

jbeemster (approved):

PR #1199, opened 2022-01-24
R131

matus-tomlein:
This updated release patches `mobile_context` schema to increase the max limit for the appAvailableMemory property. It was set too low and some devices (new iPad Pros) already cross the 4GB threshold. Increasing it to match max values for other memory properties. Hoping that we can go for a patch release instead of new version of the schema to avoid having to publish a new version of the mobile trackers and fix issues for users implementing the current tracker version.


snowplowcla:

PR #1203, opened 2022-02-21
Add new output formats to shredding_complete schema

spenes:
Since RDB Shredder will be able to output in widerow and parquet format, these new formats are included in the necessary schemas. 

Additionally, `shredProperty` field is added under `types` item in `shredding_complete` schema because new Snowflake Loader needs this information while deciding new column name.


snowplowcla:

schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/2-0-1
```diff
@@ -0,0 +1,135 @@
...
+                "maxLength": 256
+              },
+              "format": {
+                "description": "File format in which the entities were shredded to",
+                "enum": ["TSV", "JSON", "WIDEROW", "PARQUET"]
+              },
+              "shredProperty": {
+                "description": "Field types of enriched event that can be shredded (self-describing JSONs)",
+                "enum": ["SELFDESCRIBING_EVENT", "CONTEXTS"]
+              }
```
istreeter, line 44: Please can you describe to me a bit more about what this `shredProperty` means?  I don't understand it at the moment, but I might have missed something obvious.

schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/2-0-1
```diff
@@ -0,0 +1,135 @@
...
+          "items": {
+            "properties": {
+              "schemaKey": {
+                "description": "Iglu URI",
+                "type": "string",
+                "maxLength": 256
+              },
+              "format": {
+                "description": "File format in which the entities were shredded to",
+                "enum": ["TSV", "JSON", "WIDEROW", "PARQUET"]
```
istreeter, line 39: If we had to explain what each of these means:

- TSV.  Until now, we have used this to mean _shredded_ TSV.  In theory you could have non-shredded (wide) TSV, but I doubt we will ever do that.
- JSON.  Until now, we have used this to mean _shredded_ JSON.
- WIDEROW.  I assume you mean the output the of the transformer when we load into Snowflake?  So this is also JSON, but it is wide not shredded.
- PARQUET.  This format is also "wide", meaning a single parquet file contains all unstruct events and contexts.  In theory we could have shredded PARQUET - a separate file for each type - but I doubt we will ever do that.

I wonder if it's better to name the enum fields like this:
- TSV
- JSON
- WIDE_JSON
- WIDE_PARQUET

or even to name the format like this:
- TSV
- JSON
- PARQUET

...and then have a separate boolean field called "shredded" which is true for the old formats and false for our new formats.

schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/2-0-1
```diff
@@ -0,0 +1,135 @@
...
+              },
+              "shredProperty": {
+                "description": "Field types of enriched event that can be shredded (self-describing JSONs)",
+                "enum": ["SELFDESCRIBING_EVENT", "CONTEXTS"]
+              }
+            },
+            "required": ["schemaKey", "format", "shredProperty"]
```
istreeter, line 46: I think if you add a property to the `required` list then you're not allowed to make it a `2-0-1` release because it's not backwards compatible.

istreeter (commented):
Thanks for putting this together @spenes.  I look forward to hearing comments from @voropaevp and @chuwy.

schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/2-0-1
spenes: It is direct mapping of [ShredProperty in Analytics SDK](https://github.com/snowplow/snowplow-scala-analytics-sdk/blob/master/src/main/scala/com.snowplowanalytics.snowplow.analytics.scalasdk/Data.scala#L45). 

Basically, it specifies the type of self-describing JSON field in the event (context or unstruct_event). I've used `SELFDESCRIBING_EVENT` instead of `UNSTRUCT_EVENT` because atomic schema is also part of the `types` list and we need to set `shredProperty` for it too. Setting `shredProperty` to `UNSTRUCT_EVENT` for atomic schema wasn't seem right therefore I named it `SELFDESCRIBING_EVENT`.  

Since it is named as `shredProperty` in Analytics SDK, I also named it like that in here. But it can be named better in here maybe. Maybe `shredType` ? But using `type` word might create some confusion since we are already in `types` list.

spenes: This new field is put here to create column names in new Snowflake Loader in same way with old Snowflake Loader. Old Snowflake Loader is using [SnowplowEvent.transformSchema from Analytics SDK](https://github.com/snowplow-incubator/snowplow-snowflake-loader/blob/master/transformer/src/main/scala/com/snowplowanalytics/snowflake/transformer/Transformer.scala#L43) to create new column names and `transformSchema` needs `shredProperty`. In order to create column names in same way with old Snowflake Loader, new  `shredProperty` field is introduced here.

schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/2-0-1
spenes: If we were planning to add new formats which outputs in `shredded` format as well, adding new "shredded" field would make sense completely. But since we are not planning to do that, I think it might create need for additional explanation for every new format in case we add `shredded` field. We need to explain new formats and `shredded = true` can not be together. Only `JSON` would be suitable for both `shredded = true` and `shredded = false` in that case.

IMO we might convert `WIDEROW` to `WIDE_JSON` but we can leave `PARQUET` as it is since we know new fields will only be in wide format. And specify that TSV and JSON are legacy formats.

schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/2-0-1
istreeter: Ah, I understand now.  I missed before that `shredProperty` is part of the `types` array; I thought it was a single shredProperty for the entire batch.

I understand your doubts about the name, and I understand it is made more messy because atomic schema is part of types.

Would it be a good idea to remove atomic schema completely from this types array?  I know that makes it a big breaking change, but it might be easier to reason about the semantic meaning of this array.

schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/2-0-1
istreeter: Now that I understand that `format` is a field within the `types` array, my opinion has changed.

I find it odd that we need to look _inside an array_ in order to discovery the format of the primary file.  It also permits nonsense JSONs, for example this is valid against the schema but it is nonsense:

```
types: [
  { schemaKey = abc, format = WIDE_JSON }
  { schemaKey = xyz, format = PARQUET
]
```

I wonder if we should say that a type's `format` is applicable only to shredded outputs.  If a batch was not shredded, then the items' formats are null.  (or set to `INTRINSIC`)

Then separately, we can have a `fileFormat` in the top level of the schema, to represent the format of the primary atomic event file.

schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/2-0-1
spenes: I can't see any downsides for removing atomic schema. 

@chuwy May I ask if you have any objections ?

schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/2-0-1
spenes: Even though it complicates the structure a bit, I guess it is better to do it like that. I didn't look at the what changes needs to be done in the code but I guess it won't be trivial. Just wanted to mention that.

Again pinging @chuwy in case if there is anything he wants to add.

spenes:
\cc @istreeter @chuwy @voropaevp 
After update, `typesInfo` part will look like this:

For shredded:
```
"typesInfo": {
  "transformation": "SHREDDED",
  "types": [
    {
      "schemaKey": "iglu:com.acme/event/jsonschema/1-0-0",
      "format": "TSV",
      "selfDescribingJsonType": "UNSTRUCT_EVENT"
    },
  ]
}
```

For widerow:
```
"typesInfo": {
  "transformation": "WIDEROW",
  "fileFormat": "PARQUET"
  "types": [
    {
      "schemaKey": "iglu:com.acme/event/jsonschema/1-0-0",
      "selfDescribingJsonType": "CONTEXTS"
    }
  ]
}
```

I am not sure about naming of `selfDescribingJsonType` though. I think it is better than `shredProperty` but it doesn't sound right exactly still. Let me know if you have any better idea.

spenes:
Another question is what to do with atomic schema in types array ? Should we remove it ? I guess it is okay to keep it. In this case, we can add `ATOMIC` to `selfDescribingJsonType` enum list.

istreeter:
I like it.  Did you consider something like this for the widerow type:
```
"typesInfo": {
  "transformation": "WIDEROW",
  "fileFormat": "PARQUET",
  "contexts": [
    "iglu:com.acme/context/jsonschema/1-0-0"
  ],
  "unstructEvents": [
    "iglu:com.acme/event/jsonschema/1-0-0"
  ]
}
```
It avoids the `selfDescribingJsonType` enum.  It conveys exactly the same information as your suggestion, so it's just personal preference which one we pick.

> Another question is what to do with atomic schema in types array

I lean towards removing it, because it does not convey any information.  It is invalid for atomic schema to be missing, so its presence should be implicit.

These are just opinions though.  Wait for the others' comments before changing anything.

chuwy:
Was about to propose something like `snowplowEntity` instead of `selfDescribingJsonType`, but I like Ian's proposal more. Also makes the payload smaller.

spenes:
I am in favor of keeping `types` array because it seems to me kinda arbitrary to split them according their shred type. Also, we might want to add some additional field to type object in the future and current format seems more extendable in that sense.

`snowplowEntity` sounds better to me.

istreeter:
> arbitrary to split them according their shred type

I don't know if it's arbitrary.  Contexts and self-describing events are very fundamental to how snowplow works.  And they are treated differently by the loader: contexts are loaded into an array field, whereas self-describing events are loaded into an object field.  I think that's an argument to communicate them as different things in the sqs message.

> we might want to add some additional field to type object in the future

I may regret saying this 😆..... but I cannot imagine us adding an additional field type.  I mean... snowplow SDJs are either contexts or events, and that's it.

I do not hold these opinions strongly though; I just wanted to put across to alternative view point.

spenes:
I see your points but I think it doesn't add much value split to them like this. We are not doing anything particular with contexts arrays or self describing arrays. We will still process them one by one. Also, it will require a bit more code change. I am still in favor of making shred type part of type object.

schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/3-0-0
```diff
@@ -0,0 +1,185 @@
...
+                  "type": "array",
+                  "items": {
+                    "properties": {
+                      "schemaKey": {
+                        "description": "Iglu URI",
+                        "type": "string",
+                        "maxLength": 256
+                      },
+                      "format": {
+                        "description": "File format in which the entities were shredded to",
```
istreeter, line 48: File format into which the entities were shredded

or

File format which the entities were shredded into

(Pedants like the former; plain English speakers say the latter)

istreeter (approved):

schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/3-0-0
spenes: Let's be pedant then 😄

schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/3-0-0
```diff
@@ -0,0 +1,185 @@
...
+                  "items": {
+                    "properties": {
+                      "schemaKey": {
+                        "description": "Iglu URI",
+                        "type": "string",
+                        "maxLength": 256
+                      },
+                      "snowplowEntity": {
+                        "description": "Type of the self-describing JSON in the event",
+                        "enum": ["SELF_DESCRIBING_EVENT", "CONTEXTS"]
```
chuwy, line 86: Also singular.

chuwy (changes requested):
Looks good! Let's just change `CONTEXTS` to `CONTEXT`

chuwy (approved):
👍

PR #1212, opened 2022-05-18
Add com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-2 (close #1211)

### LINKED ISSUE #1211: Add com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-2
Add `firstEventTimestamp` and `eventIndex` properties to the schema.

matus-tomlein:
This PR addresses issue #1211 and adds two new properties to the `client_session` schema:

* `eventIndex` is the sequence number of the event in the current session. I chose the name to make it consistent with `sessionIndex`. It is a required property which made it a breaking change from the previous version of the schema. That's why I made this a v2.
* `firstEventTimestamp` is the timestamp of the first event in the session. It's an optional property since also `firstEventId` is optional and I guess this is related. It's a "date-time" formatted string. I was also considering using a UNIX timestamp as in the raw events, but thought this is more common in schemas. But happy to change it to UNIX timestamp as well.

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-2
+++ b/schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-2
@@ -1,0 +1,81 @@
+{
+	"$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+	"description": "Schema for a client-generated user session",
+	"self": {
+		"vendor": "com.snowplowanalytics.snowplow",
+		"name": "client_session",
+		"format": "jsonschema",
+		"version": "1-0-2"
+	},
+	"type": "object",
+	"properties": {
+		"userId": {
+			"type": "string",
+			"format": "uuid",
+			"description": "An identifier for the user of the session"
+		},
+		"sessionId": {
+			"type": "string",
+			"format": "uuid",
+			"description": "An identifier for the session"
+		},
+		"sessionIndex": {
+			"type": "integer",
+			"minimum": 0,
+			"maximum": 2147483647,
+			"description": "The index of the current session for this user"
+		},
+		"eventIndex": {
+			"type": [
+				"null",
+				"integer"
+			],
+			"minimum": 0,
+			"maximum": 2147483647,
+			"description": "Optional index of the current event in the session"
+		},
+		"previousSessionId": {
+			"type": [
+				"null",
+				"string"
+			],
+			"format": "uuid",
+			"description": "The previous session identifier for this user"
+		},
+		"storageMechanism": {
+			"type": "string",
+			"enum": [
+				"SQLITE",
+				"COOKIE_1",
+				"COOKIE_3",
+				"LOCAL_STORAGE",
+				"FLASH_LSO"
+			],
+			"description": "The mechanism that the session information has been stored on the device"
+		},
+		"firstEventId": {
+			"type": [
+				"null",
+				"string"
+			],
+			"format": "uuid",
+			"description": "The optional identifier of the first event for this session"
+		},
+		"firstEventTimestamp": {
+			"description": "Optional date-time timestamp of when the first event in the session was tracked",
+			"type": [
+				"null",
+				"string"
+			],
+			"format": "date-time"
+		}
+	},
+	"required": [
+		"userId",
+		"sessionId",
+		"sessionIndex",
+		"previousSessionId",
+		"storageMechanism"
+	],
+	"additionalProperties": false
+}


snowplowcla:

mscwilson (commented):
`eventIndex` is a good name. Not sure about the timestamp being date-time. Is that format used elsewhere in the pipeline? I'm only familiar with the unix event timestamps.

paulboocock (commented):
I think `date-time` is the correct format, it's consistent with other schemas, the unix timestamp fields only exist in the tracker protocol, so we're fine here. Each to convert between the two.

My only concern here is the `major` bump. I guess we've done that because we want to make `eventIndex` required? That would break Redshift loading in a minor/revision (at least until https://github.com/snowplow/schema-ddl/issues/123 is out in the wild), but wouldn't break anywhere else 😞.

I wonder if we make it optional for the sake of keeping this at v1? Thats much less painful for consumers of this schema and it's not _such_ a big change that a major feels like the correct option. I'm also willing for us to be purists here, but it'll slow down the release as I think we'll need to align on a data model release too.

matus-tomlein:
I don't mind making the `eventIndex` optional that much. If it makes things easier, it's probably worth it. I changed it to be optional and made the version number 1-1-0.

schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-1-0
```diff
@@ -0,0 +1,70 @@
...
+			"enum": [ "SQLITE", "COOKIE_1", "COOKIE_3", "LOCAL_STORAGE", "FLASH_LSO" ],
+			"description": "The mechanism that the session information has been stored on the device"
+		},
+		"firstEventId": {
+			"type": [
+				"null",
+				"string"
+			],
+			"format": "uuid",
+			"description": "The optional identifier of the first event id for this session"
```
mscwilson, line 57: minor: should be "of the first event for"

mscwilson (approved):
lgtm

schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-1-0
```diff
@@ -0,0 +1,70 @@
...
+		"vendor": "com.snowplowanalytics.snowplow",
+		"name": "client_session",
+		"format": "jsonschema",
+		"version": "1-1-0"
+	},
+	"type": "object",
+	"properties": {
+		"userId": {
+			"type": "string",
+			"pattern": "^[0-9a-f]{8}-([0-9a-f]{4}-){3}[0-9a-f]{12}$|^[0-9a-f]{16}$",
```
AlexBenny, line 14: I wonder if there is any advantage using this pattern rather than the _"format": "uuid"_.
I would use _format uuid_ where possible. IMO it shouldn't be a breaking change for the loader.

AlexBenny (approved):
LGTM, I haven't much to add except a note about uuid and the regex pattern. Perhaps we can use uuid always?!

paulboocock:
@istreeter how do we feel about an addition bump `*-1-*` these days? This is an addition but I know we typically don't suggest to customers that they should ever bump in this way, since the loaders don't really care(?) - I'm unaware if there are any actual downsides though and if we have a preference here?

schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-1-0
```diff
@@ -0,0 +1,70 @@
...
+			"description": "The optional identifier of the first event id for this session"
+		},
+        "firstEventTimestamp": {
+          "description": "Optional date-time timestamp of when the first event in the session was tracked",
+          "type": [
+			  "null",
+			  "string"
+		  ],
+          "format": "date-time"
+        }
```
istreeter, line 66: You use some spaces here, whereas the rest of the file uses tabs.  It's why the formatting looks a bit odd in github.

schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-1-0
matus-tomlein: I don't see a reason why not, changed it to "format": "uuid"!

schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-1-0
matus-tomlein: Good catch, reformatted to all tabs to match the other schemas.

matus-tomlein:
You're right, @istreeter, I thought adding properties is considered like adding functionality in SemVer and makes it a minor release. But reading the SchemaVer blog post, it should be a patch. Fixed!

paulboocock (approved):
LGTM

matus-tomlein:
This change was integrated into release R135 with PR #1213

PR #1213, opened 2022-05-23
R135

matus-tomlein:
This release addresses a single issue #1211 reviewed in PR #1212.


snowplowcla:

schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-2
```diff
@@ -0,0 +1,81 @@
...
+				"string"
+			],
+			"format": "date-time"
+		}
+	},
+	"required": [
+		"userId",
+		"sessionId",
+		"sessionIndex",
+		"previousSessionId",
```
istreeter, line 77: Surprising that `previousSessionId` is required even though `null` is a valid value.  It's a completely legitimate schema, just surprising.  It means a tracker couldn't drop null fields to save space before sending the event.

istreeter (commented):
There are a couple of fields which I'm surprised are nullable - but I trust you know much more than I do about the trackers!  I would have guessed that if a tracker has a concept of a session then the tracker could also be written to know these fields:

- `eventIndex`
- `firstEventId`
- `firstEventTimestamp`

schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-2
matus-tomlein: Hmm, that's a good point. I think the intention could have been to make trackers implement tracking the `previousSessionId` even though it may be null in case there is no previous session. It doesn't make much difference in the end, but might be useful to indicate that tracking the ID is required.

matus-tomlein:
@istreeter We decided to make the fields optional in order to avoid making breaking changes to the schema and keep the major version at v1. The goal was to make the upgrade less painful for consumers of the schema and bumping the major seemed unnecessary for the changes here. However, if we were to be purists, we would probably make them required.

paulboocock (approved):
Whilst I appreciate that in an ideal world we'd probably move to v2 of this schema, I think staying on v1 gives a nice blend across the trackers whilst we add this feature to them all. If we jump to v2, we'll have half the trackers on v2 and half on v1 of this schema which would be quite annoying to work with, I guess there's always going to be some pain if we ever move to v2 but avoiding it where we can with minimal impact like this seems preferrable to me.

istreeter (approved):
Approved then.  Sorry, on my last review I had not realised that this builds on top of a preivous schema.  These look like good additions.

istreeter:
@paulboocock I accidentally removed your green tick.  You might need to re-approve it and then we can merge.

paulboocock (approved):

schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-2
```diff
@@ -0,0 +1,81 @@
...
+		"vendor": "com.snowplowanalytics.snowplow",
+		"name": "client_session",
+		"format": "jsonschema",
+		"version": "1-0-2"
+	},
+	"type": "object",
+	"properties": {
+		"userId": {
+			"type": "string",
+			"format": "uuid",
```
jbeemster, line 14: Hey @matus-tomlein - I am a bit worried about this format change as 1-0-2 uses regex instead.  According to @istreeter:

```
For Redshift we treat uuid as fixed-length char https://github.com/snowplow/schema-ddl/blob/master/modules/core/src/main/scala/com.snowplowanalytics/iglu.schemaddl/redshift/generators/TypeSuggestions.scala#L135-L136

whereas a string matching a pattern is a varchar.

[TypeSuggestions.scala](https://github.com/snowplow/schema-ddl/blob/master/modules/core/src/main/scala/com.snowplowanalytics/iglu.schemaddl/redshift/generators/TypeSuggestions.scala)
      case (Some(types), Some(Format.UuidFormat)) if types.possiblyWithNull(Type.String) =>
        Some(RedshiftChar(36))

The migration might be safe, but I certainly cannot say confidently.
```

Would it be possible to revert this format change to the previous type and save it for a v2 schema?

matus-tomlein: Didn't realize that could be a problem, fixed!

jbeemster (approved):

Issue #1216, opened 2022-08-03
Add nl.basjes/yauaa_context/jsonschema/1-0-4


dilyand:
The YAUAA enrichment can fail (produce bad rows) for certain valid useragent strings, where the `webviewAppNameVersionMajor` exceeds 50 characters.

Example: `Mozilla/5.0 EmbeddedBrowser (Linux; Android 10; SM-A202F Build/QP1A.190711.020; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/103.0.5060.71 Mobile Safari/537.36 DeviceUID: c3df64e9-f56e-4ede-b20e-f792196a8ee4 VendorUID: 7b16065e-78d9-d15a-8cb9-ae24c60cc55b`

This seems to be a case where the limit should have been longer to begin with, because `webviewAppNameVersionMajor` is a concatenation of `webviewAppName` and `webviewAppVersionMajor`. In the 1-0-3 webviewAppName is unlimited and webviewAppVersionMajor is also limited to 50. So the concatenation should probably already have accepted a higher limit.

But it also points to a larger issue with the schema, which we've seen before, eg in #1126. The useragent is a very dynamic string, which can vary dramatically between implementations. Occasionally, we run the risk falling afoul of one or another vendor's limitations.

For version 1-0-4 it might be a good idea to relax the schema more generally (not just this one offending field). There are a number of properties with limits of between 50 and 100 that all seem high-risk for a future breakage along the same lines. We could extend all of them to 1,024, which appears quite generous.

At the same time, there are many properties in the schema that have no limitations at all. Perhaps we should move to removing all max lengths to keep things consistent? Introducing limits to these fields would be a backward incompatible change, so unless there's a very good reason for it, we should probably avoid that.

Also, for context, none of the limitations are imposed by YAUAA or any official useragent specification. They are just best guesses about what is reasonable.

AlexBenny:
Sounds sensible. I tag @snowplow/com-snowplowanalytics-engineering-datavalue-integrations  for a feedback on this about data modelling.

agnessnowplow:
I don't see this issue anymore from a datamodeling point of view. There used to be a limitation within the Snowflake sql runner web model on the yauaa fields but it has already been removed: https://github.com/snowplow/data-models/issues/97 The rest of the supported warehouses and the dbt web model does not impose limits on these fields regarding the VARCHAR length.

dilyand:
[...]

PR #1217, opened 2022-08-04
Release R136

dilyand:


snowplowcla:

schemas/nl.basjes/yauaa_context/jsonschema/1-0-4
```diff
@@ -0,0 +1,234 @@
...
+    },
+    "type": "object",
+    "properties": {
+        "deviceClass": {
+            "description": "See https://yauaa.basjes.nl/README-Output.html",
+            "enum": ["Desktop", "Anonymized", "Unknown", "UNKNOWN", "Mobile", "Tablet", "Phone", "Watch", "Virtual Reality", "eReader", "Set-top box", "TV", "Game Console", "Home Appliance", "Handheld Game Console", "Voice", "Car", "Robot", "Robot Mobile", "Spy", "Hacker", "Augmented Reality", "Robot Imitator"]
+        },
+        "deviceName": {
+            "description": "Example: Google Nexus 6",
+            "type": ["string", "null"],
```
istreeter, line 18: Why the change from type string to type `[string, null]`?  I'm not against it if you have a good reason.  It struck me as odd because enrich doesn't output nulls.

schemas/nl.basjes/yauaa_context/jsonschema/1-0-4
```diff
@@ -0,0 +1,234 @@
...
+            "maxLength": 1000
+        },
+        "deviceVersion": {
+            "type": ["string", "null"],
+            "maxLength": 1000
+        },
+        "operatingSystemClass": {
+            "description": "See https://yauaa.basjes.nl/README-Output.html",
+            "type": ["string", "null"],
```
istreeter, line 44: You added "null" as a type... but this does not make any difference, unless you also add `null` to the enum array.  As currently written, null is still not an allowed value for `operatingSystemClass`.

istreeter (commented):
Hey @dilyand please can you explain more about what you are trying to achieve with this schema change, and therefore why you chose these values for `maxLength`?

We spoke earlier about this being "the schema change to end all schema changes" -- meaning we're all fed up of updating enrich each time the yauaa enrichment sees an unexpected user agent string.  So we talked about lengthening each field to protect it from future changes.

But with that in mind, I can see you have chosen a mixture of 256, 1000, and 1024 as max length for different fields.  How did you decide on these values?

schemas/nl.basjes/yauaa_context/jsonschema/1-0-4
dilyand: They are optional fields, so presumably at some point they may be sent with a `null` value. We even have it as a check in igluctl. 

I appreciate we currently do not do that in enrich, but if we change it in the future, we won't need to worry about the schema not allowing it. It's part of trying to make the schema more future-proof.

schemas/nl.basjes/yauaa_context/jsonschema/1-0-4
dilyand: Yes, I did not realise that. I think if we keep null as an allowed type, I'll add it to the enum.

dilyand:
> Hey @dilyand please can you explain more about what you are trying to achieve with this schema change, and therefore why you chose these values for `maxLength`?
> 
> We spoke earlier about this being "the schema change to end all schema changes" -- meaning we're all fed up of updating enrich each time the yauaa enrichment sees an unexpected user agent string. So we talked about lengthening each field to protect it from future changes.
> 
> But with that in mind, I can see you have chosen a mixture of 256, 1000, and 1024 as max length for different fields. How did you decide on these values?

I put most of the reasoning in the issue #1216, but very briefly:
- All fields that are extracted from the useragent string have a limit of 1000 because this is the max length for the useragent field in atomic.events. So, even if the whole string is interpreted as a single property, it still can't exceed that value.
- All fields that are deduced (rather than extracted) have a limit of 128 / 256. This somewhat arbitrary. I just wanted to increase the existing limits which seemed a bit too strict. My thinking here is that these values are controlled by the YAUAA library, not by whoever puts together the useragent strings, and so are less likely to fluctuate widely.
- The fields with length 1024 were already like that. I do not want to lower this limit because that is not a backward-compatible change. But in practice, we should never have anything higher than 1000 because of the limitation for the atomic useragent field.

schemas/nl.basjes/yauaa_context/jsonschema/1-0-4
istreeter: My gut feeling at the moment is like this:

If you're going to allow nulls for some fields, then you need to allow nulls for all fields, or else it's not consistent between fields.  Which means allow nulls for the enum fields too, like `operatingSystemClass`.  Which means add `null` to the enum array.

If you're adding `null` to the enum array then you need to be very very sure that RDB Loader will not choke on the schema migration.  Remember that RDB Loader *immediately* (and probably wrongly) tries to migrate the table as soon as the schema is published; it does not wait until it sees the first `1-0-4` event.  Remember also, we need to be sure that old versions of RDB Loader also don't choke on the schema migration, otherwise we might break oss pipelines just by publishing a schema.

Now RDB Loader _probably_ handles this OK.  I'm just saying we need to be 100% sure before merging this.

For that reason.... personally I would avoid adding the nulls, and put it back to be more like how it was in 1-0-3.  There is no reason whatsoever to add nulls to this schema.  I see no reason why enrich will ever change in future and to start emitting nulls.

Yes we want to make the schema future-proof, but I see it as future-proofing against things we are not in control of, like crazy user agent strings.  Whereas nulls are something we are completely in control of, so we don't need to future-proof against nulls.

dilyand: I think that is a sensible idea. I was trying to make igluctl happy, but I agree that in this instance, where we're in control of this, there is no need to worry.

schemas/nl.basjes/yauaa_context/jsonschema/1-0-4
```diff
@@ -0,0 +1,234 @@
+{
+    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "description": "Schema for an entity context generated by the YAUAA enrichment after parsing the user agent",
```
istreeter, line 3: Entity context or context entity?

istreeter (approved):
Looks good.

oguzhanunlu:
yauaa 7.4.0 has `WebviewAppNameVersion` as well, it is not documented [here](https://yauaa.basjes.nl/using/memoryusage/)

[ref](https://github.com/nielsbasjes/yauaa/blob/v7.4.0/analyzer/src/main/java/nl/basjes/parse/useragent/UserAgent.java#L120)

is there a specific reason we didn't include `webviewAppNameVersion` field in this new version of the schema?

oguzhanunlu:
I added a separate commit to add `webviewAppNameVersion` cc @istreeter @dilyand

adatzer:
Could we also merge https://github.com/snowplow/iglu-central/pull/1221 into `r136`? cc @oguzhanunlu @dilyand

CHANGELOG
```diff
@@ -1,3 +1,10 @@
+Release 136 (2022-08-23)
+------------------------
+Add nl.basjes/yauaa_context/jsonschema/1-0-4 (#1216)
+Extend copyright to 2022 (close #1218)
+Update links in Readme (close #1219)
+Add CONTRIBUTING.md (close #1220)
```
paulboocock, line 6: nit: I don't think we usually have the word `close` in these lines

schemas/nl.basjes/yauaa_context/jsonschema/1-0-4
paulboocock: "context entity" feels more right to me. I'd probably just go for "entity" though.

paulboocock (approved):
LGTM. Couple of nit pick comments but happy for this to go out once the changelog is tidied.

schemas/nl.basjes/yauaa_context/jsonschema/1-0-4
dilyand: I changed it back to just 'context'. I think the word entity confuses things in this case. This schema does not describe an entity, but rather just a grab bag of things that could be deduced from the useragent.

CHANGELOG
dilyand: Done.

paulboocock (approved):

CONTRIBUTING.md
```diff
@@ -0,0 +1,22 @@
...
+## How to find out more information
+
+Find out more information on the [contributing page][docs-contributing] in our docs.
+
+## Sign our Contributor License Agreement
+
+We need all contributors to sign a CLA - you can find out why and how to sign it [here][cla].
+
+[docs-contributing]: https://docs.snowplowanalytics.com/docs/contributing/
+[cla]: https://docs.snowplowanalytics.com/docs/contributing/contributor-license-agreement/
```
jbeemster, line 22:
```suggestion
[cla]: https://docs.snowplow.io/docs/contributing/contributor-license-agreement/
```

schemas/com.snowplowanalytics.mobile/remote_config/jsonschema/1-0-1
```diff
@@ -0,0 +1,225 @@
+{
```
jbeemster, line 1: @istreeter doing a diff with 1-0-0 a lot of these fields now have a `maximum` as well as a few now allowing `object, null` pairing over just `object`.  Just want to confirm that the addition of new limits and type groupings is not going to cause issues in any of our loaders / data-warehouses?

I am not entirely certain if this schema gets tracked into pipelines to be fair but if it was would these changes get automatically represented in their respective columns?

matus-tomlein: The question is still valid, just wanted to say that the schema is not being used in pipelines, it is just to help developers better define their configurations (some IDEs may use it for validating the JSONs).

README.md
```diff
@@ -60,14 +60,14 @@ limitations under the License.
...
 [setup-image]: https://d3i6fms1cm1j0i.cloudfront.net/github/images/setup.png
 [roadmap-image]: https://d3i6fms1cm1j0i.cloudfront.net/github/images/roadmap.png
 [contributing-image]: https://d3i6fms1cm1j0i.cloudfront.net/github/images/contributing.png
 
-[techdocs]: https://github.com/snowplow/iglu/wiki/Iglu-Central
-[setup]: https://github.com/snowplow/iglu/wiki/Iglu-Central-setup
-[roadmap]: https://github.com/snowplow/iglu/wiki/Product-roadmap
-[contributing]: https://github.com/snowplow/iglu/wiki/Contributing-to-Iglu-Central
+[techdocs]: https://docs.snowplowanalytics.com/docs/pipeline-components-and-applications/iglu/iglu-repositories/iglu-central/
+[setup]: https://docs.snowplowanalytics.com/docs/pipeline-components-and-applications/iglu/iglu-central-setup/
```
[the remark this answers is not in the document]
adatzer: All suggested changes applied to contributing and readme files:
 - Links to snowplow.io
 - "Snowplow Analytics" to "Snowplow"

schemas/com.snowplowanalytics.mobile/remote_config/jsonschema/1-0-1
istreeter: I tested this on a pipeline with a loader.  Even if it's not expected to be used in tracking, it was worth testing.

It is completely safe, because the field `configurationBundle` is a complex object, so RDB transformer simply stringifies the whole field, and RDB Loader loads it as a string.  Any change you make to a sub-field (e.g. `configurationBundle.namespace`) does not bother the transformer/loader at all.

schemas/nl.basjes/yauaa_context/jsonschema/1-0-4
```diff
@@ -0,0 +1,238 @@
+{
```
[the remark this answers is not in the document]
istreeter: **WARNING** -- this migration did not work when I tested in with Redshift RDB Loader.  The new field got added to the table, but the existing columns were not updated for the new lengths.

I opened [this issue in rdb loader](https://github.com/snowplow/snowplow-rdb-loader/issues/1051) to investigate it.  I spent quite a while trying to pinpoint the error, but no luck yet.

I'm not comfortable with merging this release until we have found why RDB Loader did not alter the lengths as it should.

schemas/com.snowplowanalytics.mobile/remote_config/jsonschema/1-0-1
```diff
@@ -0,0 +1,225 @@
...
+            "type": ["object", "null"],
+            "properties": {
+              "endpoint": {
+                "description": "URL of the collector that is going to receive the events tracked by the tracker. The URL can include the schema/protocol (e.g.: http://collector-url.com). In case the URL doesn’t include the schema/protocol, the HTTPS protocol is automatically selected.",
+                "type": "string",
+                "maxLength": 2048
+              },
+              "method": {
+                "description": "The method used to send the requests (GET or POST).",
+                "type": ["string", "null"],
```
istreeter, line 41: Hey @matus-tomlein just a warning, the "null" on this line does not do what you think it does!  If you want this field to be nullable then you must also add `null` to the enum:

```
"enum": ["get", "post", null]
```

Otherwise, a null will get rejected during validation.

I know this is slightly surprising.  You can test it out though [using an online validator](https://www.jsonschemavalidator.net/).

matus-tomlein: Thanks @istreeter, didn't know about this! Fixed.

schemas/io.snowplow.foundation/content/jsonschema/1-0-0
```diff
@@ -0,0 +1,37 @@
+{
```
paulboocock, line 1: If I'm being nitpicky, all these `io.snowplow.foundation` schemas are upsidedown compared to the rest of the Iglu Central schemas. I appreciate thats how the console renders them but would it be good to maintain Iglu Central patterns here?

schemas/io.snowplow.foundation/content/jsonschema/1-0-0
```diff
@@ -0,0 +1,37 @@
...
+      "maxLength": 255
+    }
+  },
+  "additionalProperties": false,
+  "type": "object",
+  "required": [
+    "name"
+  ],
+  "self": {
+    "vendor": "io.snowplow.foundation",
```
paulboocock, line 31: What does `foundation` represent/mean? Are these not the recipe schemas from Try Snowplow? Why not `io.snowplow.recipe` or a final part that represents the "usecase" we typically see this schema used in `io.snowplow.content`?

schemas/io.snowplow.foundation/object/jsonschema/1-0-0
```diff
@@ -0,0 +1,36 @@
...
+  "type": "object",
+  "required": [
+    "name"
+  ],
+  "self": {
+    "vendor": "io.snowplow.foundation",
+    "name": "object",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
```
paulboocock, line 34: This is a weird old schema. Not something we'd ever suggest someone uses in the real world as its so generic. Whats the use case for a schema like this?

paulboocock (commented):
Technically I have no issues with the schemas, but there's a few nitpicks and a few questions on if these are the schemas we want to be commiting to for whatever use case they are being used to build (especially `object` 🤮)

schemas/io.snowplow.foundation/content/jsonschema/1-0-0
istreeter: I was OK with the name "foundation" for the vendor, meaning like a collection of schemas that can be used when first starting off with Snowplow, before graduating to more advanced use cases when the user would want to design their own bespoke schemas.

But... I find the names of the schemas a bit strange.  The names "content" and "conversion" don't mean anything to me, but maybe that's because I don't come from the world of web analytics.

I don't have any problem with the schemas though, so I'd be OK to approve this release.

zephemeral: `foundation` was the group consensus for what Ian describes

we didn't want to go with use-case specific names in case we later want to add more advanced content / conversion / funnel schemas that would then clash with these

zephemeral: I'd be happy with `recipe` too, I don't who that decision is ultimately up to

schemas/io.snowplow.foundation/content/jsonschema/1-0-0
zephemeral: potentially a dumb question but I'm comparing a couple of schemas and I don't know what upsidedown means – could you clarify? 
If there's a pattern we should follow then sure we can rewrite the schemas for consistency's sake

schemas/io.snowplow.foundation/object/jsonschema/1-0-0
zephemeral: Good point, I just checked and it's not actually used in any of the tutorials, so I'm happy to ditch this one

schemas/io.snowplow.foundation/content/jsonschema/1-0-0
paulboocock: Sorry to be that guy...

Now that I understand what `foundation` means, as in foundational... I think I prefer it over receipes as its more generic and allows these schemas to be used in a more broader context than just the recipe context.

jbeemster (approved):
LGTM - can I get a green-light when you lot get online and ill go ahead and merge it!

jbeemster (approved):

PR #1224, opened 2022-08-26
Add com.snowplowanalytics.snowplow/web_page/jsonschema/1-0-1 (closes #1222)

### LINKED ISSUE #1222: Add com.snowplowanalytics.snowplow/web_page/jsonschema/1-1-0
Add the `tabId` property on the web_page schema. Based on https://github.com/snowplow/snowplow-javascript-tracker/pull/1088

igneel64:
Closes #1222 

- Both `id` and `tabId` are generated using the [uuid](https://www.npmjs.com/package/uuid) pacakge which produces valid UUIDs. 
- `tabId` is the identifier of the current tab the event was sent from. More info on the [JS repository](https://github.com/snowplow/snowplow-javascript-tracker/pull/1088).

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/web_page/jsonschema/1-0-1
+++ b/schemas/com.snowplowanalytics.snowplow/web_page/jsonschema/1-0-1
@@ -1,0 +1,25 @@
+{
+    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "description": "Schema for a web page context",
+    "self": {
+        "vendor": "com.snowplowanalytics.snowplow",
+        "name": "web_page",
+        "format": "jsonschema",
+        "version": "1-0-1"
+    },
+    "type": "object",
+    "properties": {
+        "id": {
+            "type": "string",
+            "format": "uuid",
+            "description": "An identifier for the current page view."
+        },
+        "tabId": {
+            "type": ["string", "null"],
+            "format": "uuid",
+            "description": "An identifier for the client browser tab the event is sent from."
+        }
+    },
+    "required": ["id"],
+    "additionalProperties": false
+}


snowplowcla:

matus-tomlein (approved):
LGTM! 👍

paulboocock (changes requested):
I have to admit, I'm a little nervous about changing the `web_page` schema. An addition shouldn't pose much trouble to those on Redshift, Snowflake and Databricks - I worry about a little pain for those on BigQuery that aren't using the new dbt datamodels.

@emielver can you comment on how painful it will be across the `sql-runner` models that don't combine the columns on BigQuery - I'm mainly worried that because page view id is so important its not a simple coalesce. 

If the above does feel like a major issue (and I really hope it isnt), then we might want a new schema for the tab uuid, which is a shame from a payload size point of view.

Lastly, this should be 1-0-1. We, perhaps confusingly, don't use SemVer here, we use something called SchemaVer which has a slightly different set of rules. An addition in this manner is a pure ADDITION and is backwards compatible. Infact we very rarely increment on the REVISION, since the circumstance is rare.

https://snowplowanalytics.com/blog/2014/05/13/introducing-schemaver-for-semantic-versioning-of-schemas/#schemaver

agnessnowplow:
@paulboocock As Emiel is off this week I had a look at this. The page_view context is taken during the creation of the events-this-run table which is then referred back to when building the page-view model. See related BigQuery sql for [web](https://github.com/snowplow/data-models/blob/master/web/v1/bigquery/sql-runner/sql/standard/01-base/01-main/06-events-this-run.sql ) 

The coalesce would need to happen here:
```sql
SELECT
        a.contexts_com_snowplowanalytics_snowplow_web_page_1_0_0[SAFE_OFFSET(0)].id AS page_view_id,
        a.* EXCEPT(contexts_com_snowplowanalytics_snowplow_web_page_1_0_0)

    FROM
      {{.input_schema}}.events a
```
To create the automatic coalescing there is already a procedure called `combine_context_versions` in the sql runner model that is used when referencing other context fields so the groundwork is there I think. I would also point out that the dbt web is also impacted as currently the 1_0_0 page_view_context schema is hard_coded but, as you mentioned we could leverage the `combine_column_versions` macro there. The `EXCEPT` bit needs further handling in case we want to limit superfluous columns and the integration tests would also need rework so that needs to be taken into consideration.

Not THAT quick to solve but is doable as much as I can see it from this.

emielver:
Just to elaborate on Agnes' answer, the `combine_context_versions` procedure for SQL runner creates a new table to include the context fields in a coalesced manner, so we will need to revamp the `events_this_run` logic as Agnes outlined. I think we'd need to also have an intermediary `events_this_run` table which doesn't yet include the `page_view_id`. We'd need to call it something else, however, so we'd have to first create a `events_this_run_temp` table, which we could define as follows:

```SQL
CREATE OR REPLACE TABLE {{.scratch_schema}}.events_this_run_temp{{.entropy}}
AS(
  -- Without downstream joins, it's safe to dedupe by picking the first event_id found.
  SELECT
    ARRAY_AGG(e ORDER BY e.collector_tstamp LIMIT 1)[OFFSET(0)].*
  FROM (
    SELECT
        a.* EXCEPT(contexts_com_snowplowanalytics_snowplow_web_page_1_0_0, contexts_com_snowplowanalytics_snowplow_web_page_1_0_1)

    FROM
      {{.input_schema}}.events a
    INNER JOIN
      {{.scratch_schema}}.base_sessions_to_include{{.entropy}} b
    ON a.domain_sessionid = b.session_id
    WHERE
      a.collector_tstamp >= LOWER_LIMIT
      AND a.collector_tstamp <= UPPER_LIMIT

      {{if eq (or .derived_tstamp_partitioned false) true}}

        AND a.derived_tstamp >= LOWER_LIMIT
        AND a.derived_tstamp <= UPPER_LIMIT

      {{end}}

  ) e
  GROUP BY
    e.event_id
);
```

Then we'd also need to introduce a different `combine_context_versions` procedure specifically for the web page context, to not take the `page_view_id` from the events_this_run table. It would look something like this:

```SQL
CREATE OR REPLACE PROCEDURE {{.output_schema}}.combine_context_versions_web_page (columns_prefix STRING)
BEGIN
  DECLARE COLUMN_COALESCE STRING;

  SET COLUMN_COALESCE = (
    -- Flatten results of nested query into comma separated list of coalesces
    SELECT ARRAY_TO_STRING(
      ARRAY_AGG(CONCAT('COALESCE(', ARRAY_TO_STRING(paths, ', '), ', NULL) AS ',  field_name)), ', ')
    FROM(
      -- Get field names and their corresponding paths
      SELECT
        SPLIT(field_path, '.')[SAFE_OFFSET(1)] AS field_name,
        ARRAY_AGG(CONCAT(column_name, '[SAFE_OFFSET(0)].', SPLIT(field_path, '.')[SAFE_OFFSET(1)]) ORDER BY column_name DESC) AS paths

      FROM {{.scratch_schema}}.INFORMATION_SCHEMA.COLUMN_FIELD_PATHS
      WHERE table_name = 'events_this_run_temp{{.entropy}}'
      AND column_name LIKE CONCAT(columns_prefix, '%')
      AND ARRAY_LENGTH(SPLIT(field_path, '.')) = 2 -- Only first-order fields
      AND data_type NOT LIKE 'STRUCT%' -- No structs
      AND data_type NOT LIKE 'ARRAY%' -- No arrays
      GROUP BY 1
    )
  );

  IF COLUMN_COALESCE IS NULL THEN
    RAISE USING MESSAGE = 'ERROR: Cannot combine context versions: No eligible top-level columns found.';

  END IF;

  -- Create scratch table with extracted data:
  EXECUTE IMMEDIATE CONCAT("""CREATE OR REPLACE TABLE {{.scratch_schema}}.""",
                            columns_prefix,
                            """{{.entropy}} AS( SELECT event_id, """,
                            COLUMN_COALESCE,
                            """ FROM {{.scratch_schema}}.events_this_run_temp{{.entropy}})"""); 
END;
```

.We would also need to introduce a new step before the `events_this_run` step but after the `events_this_run_temp` step to call the procedure to create the page view context table, but that should be a quick one, probably something like this:

```SQL
CALL {{.output_schema}}.combine_context_versions_web_page('contexts_com_snowplowanalytics_snowplow_web_page_1');
```

We'd need to change the `events_this_run` step to something like this:
```SQL
CREATE OR REPLACE TABLE {{.scratch_schema}}.events_this_run{{.entropy}}
AS(
    SELECT
        wpc.id as page_view_id,
        a.*

    FROM
      {{.scratch_schema}}.events_this_run_temp a
   INNER JOIN {{.scratch_schema}}.contexts_com_snowplowanalytics_snowplow_web_page_1{{.entropy}} wpc
    ON a.event_id = wpc.event_id
);
```

I think if we did that then we should be fine, although I might have missed something. Up to you whether you consider that a lot of work or not, but I think we're adding two steps and a new stored procedure to ensure things continue to run smoothly on SQL runner if we introduce this change.

paulboocock:
So your response makes me realise this is not the right time to introduce 1-0-1 until we've done the above to the models first. However, it's also clear this will have a significant impact on older data models which we won't be updating but know are out in the wild.

I think theres two choices:
1. Update dbt models to be ready for a 1-0-1 and block this feature being released until then - also knowing this will have consequence for people on older models when they update their tracker. Meaning it should perhaps be a tracker v4 feature.
2. Create a new schema/entity and store the tab id there. This is safe, the only downside I see is that it makes the payload bigger. Sort of a shame for one extra uuid.

paulboocock:
So to move this forward, lets go with option 2.

I think we dependencies on 1-0-0 of the webPage schema are too great at the moment to justify this optional schema.

matus-tomlein:
Wondering if this could be a good time to introduce a `browser_context` schema similar to the [`mobile_context`](http://iglucentral.com/schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-2) that we track on the mobile trackers? We were talking about deprecating some of the atomic event fields and maybe a browser context could also cover the resolution, viewport, browser language, cookies flag, document charset, or even the page offsets set by page ping (that probably fits the `web_page` context better)?

It's just an idea, might not be worth it, just thinking if we should go more general to avoid having to introduce other tiny entities in the future.

paulboocock:
Sounds like a good idea to me! Could maybe make it a toggle in v4, either track the atomic fields, or the browser_context?

igneel64:
Included in #1250

PR #1243, opened 2022-11-18
Release R138

### LINKED ISSUE #1241: Add com.snowplowanalytics.snowplow.ecommerce
Add schemas for the new Snowplow ecommerce plugin:
- Plugin PR: https://github.com/snowplow/snowplow-javascript-tracker/pull/1110

### LINKED ISSUE #1242: Add com.snowplowanalytics.snowplow/cmp_visible/jsonschema/1-0-0
Add cmp_visible schema.
Plugin PR - https://github.com/snowplow/snowplow-javascript-tracker/pull/1121

### LINKED ISSUE #1246: Add com.snowplowanalytics.snowplow/consent_preferences/jsonschema/1-0-0
Add consent_preferences schema.
Plugin PR - https://github.com/snowplow/snowplow-javascript-tracker/pull/1121

igneel64:
Schemas for the Snowplow Ecommerce plugin and the Enhanced Consent plugin.
(closes #1241)
(closes #1242)
(closes #1246 )

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow.ecommerce/cart/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow.ecommerce/cart/jsonschema/1-0-0
@@ -1,0 +1,35 @@
+{
+    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "description": "Schema for a cart entity in Ecommerce",
+    "self": {
+        "vendor": "com.snowplowanalytics.snowplow.ecommerce",
+        "name": "cart",
+        "format": "jsonschema",
+        "version": "1-0-0"
+    },
+    "type": "object",
+    "properties": {
+        "cart_id": {
+            "description": "The unique ID representing this cart.",
+            "type": ["string","null"],
+            "maxLength": 4096
+        },
+        "total_value": {
+            "description": "The total value of the cart after this interaction.",
+            "type": "number",
+            "minimum": 0,
+            "multipleOf": 0.01,
+            "maximum": 9999999
+        },
+        "currency": {
+            "description": "The currency used for this cart (ISO 4217).",
+            "type": "string",
+            "maxLength": 3,
+            "minLength": 3
+        }
+    },
+    "required": [
+        "total_value", "currency"
+    ],
+    "additionalProperties": false
+}
--- a/schemas/com.snowplowanalytics.snowplow.ecommerce/checkout_step/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow.ecommerce/checkout_step/jsonschema/1-0-0
@@ -1,0 +1,107 @@
+{
+    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "description": "Schema for a checkout step entity in Ecommerce",
+    "self": {
+        "vendor": "com.snowplowanalytics.snowplow.ecommerce",
+        "name": "checkout_step",
+        "format": "jsonschema",
+        "version": "1-0-0"
+    },
+    "type": "object",
+    "properties": {
+        "step" : {
+            "description" : "Checkout step index.",
+            "type": "integer",
+            "minimum": 1,
+            "maximum": 90
+        },
+        "shipping_postcode": {
+            "description": "Shipping address postcode.",
+            "type" : [
+                "string",
+                "null"
+            ],
+            "maxLength": 128
+        },
+        "billing_postcode": {
+            "description": "Billing address postcode.",
+            "type" : [
+                "string",
+                "null"
+            ],
+            "maxLength": 128
+        },
+        "shipping_full_address": {
+            "description": "Full shipping address.",
+            "type" : [
+                "string",
+                "null"
+            ],
+            "maxLength": 4096
+        },
+        "billing_full_address": {
+            "description": "Full billing address.",
+            "type" : [
+                "string",
+                "null"
+            ],
+            "maxLength": 4096
+        },
+        "delivery_provider": {
+            "description": "Can be used to discern delivery providers DHL, PostNL etc.",
+            "type" : [
+                "string",
+                "null"
+            ],
+            "maxLength": 128
+        },
+        "delivery_method": {
+            "description": "Can be used to discern delivery methods selected E.g. store pickup, standard delivery, express delivery, international.",
+            "type" : [
+                "string",
+                "null"
+            ],
+            "maxLength": 128
+        },
+        "coupon_code": {
+            "description": "Coupon applied at checkout.",
+            "type" : [
+                "string",
+                "null"
+            ],
+            "maxLength": 128
+        },
+        "account_type": {
+            "description": "Type of account used on checkout. E.g. existing user, guest.",
+            "type" : [
+                "string",
+                "null"
+            ],
+            "maxLength": 128
+        },
+        "payment_method": {
+            "description": "Any kind of payment method the user selected to proceed E.g. card, PayPal, Alipay etc.",
+            "type" : [
+                "string",
+                "null"
+            ],
+            "maxLength": 128
+        },
+        "proof_of_payment": {
+            "description": "Invoice or receipt.",
+            "type" : [
+                "string",
+                "null"
+            ],
+            "maxLength": 128
+        },
+        "marketing_opt_in": {
+            "type" : ["boolean", "null"],
+            "description": "If opted in to marketing campaigns."
+        }
+    },
+    "required": [
+        "step"
+    ],
+    "additionalProperties": false
+}
--- a/schemas/com.snowplowanalytics.snowplow.ecommerce/page/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow.ecommerce/page/jsonschema/1-0-0
@@ -1,0 +1,38 @@
+{
+    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "description": "Schema for a page entity in Ecommerce",
+    "self": {
+        "vendor": "com.snowplowanalytics.snowplow.ecommerce",
+        "name": "page",
+        "format": "jsonschema",
+        "version": "1-0-0"
+    },
+    "type": "object",
+    "properties": {
+        "type": {
+            "description": "The type of the page that was visited E.g. homepage, product page, checkout page.",
+            "type": "string",
+            "maxLength": 128
+        },
+        "language": {
+            "description": "The language that the web page is based in.",
+            "type" : [
+                "string",
+                "null"
+            ],
+            "maxLength": 128
+        },
+        "locale": {
+            "description": "The locale version of the site that is running.",
+            "type": [
+                "string",
+                "null"
+            ],
+            "maxLength": 128
+        }
+    },
+    "required": [
+        "type"
+    ],
+    "additionalProperties": false
+}
--- a/schemas/com.snowplowanalytics.snowplow.ecommerce/product/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow.ecommerce/product/jsonschema/1-0-0
@@ -1,0 +1,107 @@
+{
+    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "description": "Schema for a product entity in Ecommerce",
+    "self": {
+        "vendor": "com.snowplowanalytics.snowplow.ecommerce",
+        "name": "product",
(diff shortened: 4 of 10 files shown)


schemas/com.snowplowanalytics.snowplow.ecommerce/snowplow_ecommerce_action/jsonschema/1-0-0
```diff
@@ -0,0 +1,37 @@
...
+          "transaction"
+        ]
+      },
+      "name": {
+        "description": "The name of the list presented to the user E.g. product list, search results, shop the look, frequently bought with.",
+        "type": [
+          "string",
+          "null"
+        ],
+        "maxLength": 128
```
rlh1994, line 30: The description on the first one doesn't line up with the values - the second one might be fine I'm just not sure on that one :)

schemas/com.snowplowanalytics.snowplow.ecommerce/product/jsonschema/1-0-0
```diff
@@ -0,0 +1,105 @@
...
+            "description" : "The category the product belongs to. Use a consistent separator to express multiple levels. E.g. Woman/Shoes/Sneakers",
+            "type": "string",
+            "maxLength": 4096
+        },
+        "price": {
+            "description" : "The price of the product at the current time.",
+            "type": "number",
+            "minimum": 0,
+            "multipleOf": 0.01,
+            "maximum": 2147483647
```
istreeter, line 32: So `2147483647` is a sensible maximum to use when storing a 32-bit integer, but it is not the best choice for maximum when storing a Decimal type in the warehouse.

I suggest set maximum to `9999999` and then the warehouse will use a 32-bit column to store the full range of allowed values.  Do you think `9999999` is sufficiently large as a maximum?  Or do you think we need to raise the threshold and then use a 64-bit column in the warehouse?

The same argument goes for all other places in this PR where you set `multipleOf: 0.01` and `maximum: 2147483647`.

schemas/com.snowplowanalytics.snowplow.ecommerce/transaction/jsonschema/1-0-0
```diff
@@ -0,0 +1,78 @@
...
+        "currency": {
+            "type": "string",
+            "description": "The currency used for the transaction (ISO 4217).",
+            "maxLength": 3,
+            "minLength": 3
+        },
+        "payment_method": {
+            "type": "string",
+            "description": "The payment method used for the transaction.",
+            "maxLength": 99
```
istreeter, line 33: In the checkout schema, the max length of payment_method is 128, but here you set it to 99.  Is this deliberate to have a discrepancy?

schemas/com.snowplowanalytics.snowplow.ecommerce/transaction/jsonschema/1-0-0
```diff
@@ -0,0 +1,78 @@
...
+        },
+        "total_quantity": {
+            "type": "integer",
+            "description": "Total quantity of items in the transaction.",
+            "minimum": 0
```
istreeter, line 38: Should there be a maximum on this field?

schemas/com.snowplowanalytics.snowplow/consent_preferences/jsonschema/1-0-0
```diff
@@ -0,0 +1,57 @@
...
+    "description": "Schema for consent preferences selection event",
+    "self": {
+        "vendor": "com.snowplowanalytics.snowplow",
+        "name": "consent_preferences",
+        "format": "jsonschema",
+        "version": "1-0-0"
+    },
+    "type": "object",
+    "properties": {
+        "eventType": {
```
istreeter, line 12: Of all the schemas in this PR, this is the only one to use cameCase instead of snake_case for field names.  Is this deliberate?

schemas/com.snowplowanalytics.snowplow.ecommerce/page/jsonschema/1-0-0
```diff
@@ -0,0 +1,38 @@
...
+            "type": "string",
+            "maxLength": 128
+        },
+        "language": {
+            "description": "The language that the web page is based in.",
+            "type" : [
+                "string",
+                "null"
+            ],
+            "maxLength": 128
```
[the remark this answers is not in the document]
igneel64: On the `language`, we expect the value to come from a more 'text friendly' field e.g. the option that you selected in a shop language selector.

On the `locale` field we consider any value _(or key)_ the client selects to send to the database. Inspiration came from these lists:
- https://gist.github.com/jacobbubu/1836273
- https://stackoverflow.com/questions/3191664/list-of-all-locales-and-their-short-codes

schemas/com.snowplowanalytics.snowplow.ecommerce/product/jsonschema/1-0-0
```diff
@@ -0,0 +1,105 @@
...
+        "category": {
+            "description" : "The category the product belongs to. Use a consistent separator to express multiple levels. E.g. Woman/Shoes/Sneakers",
+            "type": "string",
+            "maxLength": 4096
+        },
+        "price": {
+            "description" : "The price of the product at the current time.",
+            "type": "number",
+            "minimum": 0,
+            "multipleOf": 0.01,
```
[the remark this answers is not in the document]
igneel64: Nice, believe so as well 👍

schemas/com.snowplowanalytics.snowplow.ecommerce/product/jsonschema/1-0-0
igneel64: I think it is a sufficient number, changing it.

schemas/com.snowplowanalytics.snowplow/consent_preferences/jsonschema/1-0-0
igneel64: These were two separate projects with different casing as it seems, not a deliberate decision.
Do we have a standard that every schema should follow moving forwards ?

schemas/com.snowplowanalytics.snowplow.ecommerce/transaction/jsonschema/1-0-0
igneel64: On similar `quantity` type fields, I have not seen a maximum in the past, I can set `9999999` which is valid for our case and stays in the 32-bit range.

schemas/com.snowplowanalytics.snowplow.ecommerce/transaction/jsonschema/1-0-0
igneel64: Nope, good catch. Fixing

schemas/com.snowplowanalytics.snowplow.ecommerce/checkout_step/jsonschema/1-0-0
```diff
@@ -0,0 +1,107 @@
...
+            ],
+            "maxLength": 128
+        },
+        "shipping_full_address": {
+            "description": "Full shipping address.",
+            "type" : [
+                "string",
+                "null"
+            ],
+            "maxLength": 128
```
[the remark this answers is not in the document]
igneel64: Applied :)

schemas/com.snowplowanalytics.snowplow/consent_preferences/jsonschema/1-0-0
istreeter: From Data Processing point of view, we do not have a standard and we allow anything.

But I wonder if your team wants to conform to a standard that you set for yourselves?

If you pick camel case, then events will look different between different warehouses.  For example, the field `basisForProcessing` -- in Snowflake this will look like `basisForProcessing` whereas in Redshift/Databricks/BigQuery the name will get converted to `basis_for_processing`.  Each warehouse will handle it just fine, but you end up with an inconsistency across different warehouses, which might be harder for you to manage.

If you pick `basis_for_processing` as the name of the original field then you don't have any inconsistency across warehouses.

I don't feel strongly about this, because it doesn't affect me or my team!!!  I'm just offering you this opportunity to think about whether your team needs a convention to stick to.

rlh1994: For what it's worth, the `normalize` package does translate from camel to snake but it pulls the original case from the schema in the first place, so it shouldn't break if they were set as snake in the schema. However from what I've seen we do tend to use camel in most schemas I looked at.

schemas/com.snowplowanalytics.snowplow.ecommerce/cart/jsonschema/1-0-0
```diff
@@ -0,0 +1,34 @@
...
+        "version": "1-0-0"
+    },
+    "type": "object",
+    "properties": {
+        "cart_id": {
+            "description": "The unique ID representing this cart.",
+            "type": ["string","null"],
+            "maxLength": 4096
+        },
+        "total_value": {
```
istreeter, line 17: Is this also a monetary field?  If yes, I think it would be great to put `multipleOf: 0.01` like you've done on other fields.  If you do this, then it will get stored in the warehouse as a decimal type, and that is the standard way to store money fields in databases.

schemas/com.snowplowanalytics.snowplow.ecommerce/product/jsonschema/1-0-0
```diff
@@ -0,0 +1,105 @@
...
+            "maximum": 9999999
+        },
+        "list_price": {
+            "description" : "The list or recommended retail price of a product.",
+            "type" : ["number", "null"],
+            "minimum": 0,
+            "multipleOf": 0.01,
+            "maximum": 9999999
+        },
+        "quantity" : {
```
istreeter, line 41: Currently, this will get stored in the warehouse as a 64-bit integer.  If you put a `maximum` on it, then you can make it get stored as a 32-bit integer -- if that's something you would prefer.

schemas/com.snowplowanalytics.snowplow.ecommerce/product/jsonschema/1-0-0
```diff
@@ -0,0 +1,105 @@
...
+        },
+        "inventory_status": {
+            "description" : "The inventory status of the product E.g. in stock, out of stock, preorder, backorder.",
+            "type" : [
+                "string",
+                "null"
+            ],
+            "maxLength": 256
+        },
+        "position": {
```
istreeter, line 78: Same comment about 32-bit vs 64-bit integer.

istreeter (commented):
I'm happy with the responses to my questions since last time I reviewed it.

As I reviewed it a second time I spotted a couple of other (small) things for you to consider.  Apart from these small things, I think the schemas look great and I see no problem with merging them.

schemas/com.snowplowanalytics.snowplow.ecommerce/cart/jsonschema/1-0-0
igneel64: Good catch, done!

istreeter (approved):

schemas/com.snowplowanalytics.snowplow.ecommerce/snowplow_ecommerce_action/jsonschema/1-0-0
```diff
@@ -0,0 +1,37 @@
...
+      }
+    },
+    "required": [
+      "type"
+    ],
+    "additionalProperties": false
+  }
```
jbeemster, line 37: Formatting seems off on this one as compared to the others?

igneel64: Yes seemed to have some places mis-indented.

schemas/com.snowplowanalytics.snowplow.ecommerce/snowplow_ecommerce_action/jsonschema/1-0-0
```diff
@@ -0,0 +1,37 @@
+{
+    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "description": "Schema for an Ecommerce action",
+    "self": {
+        "vendor": "com.snowplowanalytics.snowplow.ecommerce",
+        "name": "snowplow_ecommerce_action",
```
jbeemster, line 6: This is really nit-picky so feel free to ignore - the vendor is already `com.snowplowanalytics.snowplow.ecommerce` so the name here feels like it could be just `action`.  The fact that it is from Snowplow and related to ecommerce is already stated in the vendor.  Wdyr?

igneel64: It was explicitly set like that to assist in the modeling way :)

jbeemster (approved):

PR #1245, opened 2022-12-02
Add additional GA4 schemas

miike:
Adds some additional missing GA4 / measurement protocol schemas for these events. Some events (tutorials) have no properties.


miike:
cc @adatzer @zloff

adatzer (approved):

adatzer:
@miike closing in favor of #1387 , where the additional GA4 schemas are included as separate commits with minor/nit diffs:

- added permissive `maxLength` to string properties, matching similar lengths we use in other GTM-SS schemas
- removed `location_id` from `select_promotion` and `view_promotion` as it does not appear to be a property of those events
- for `add_shipping_info`, `remove_from_cart` and `view_cart`, reduced the required properties to just `items`, to match what we've done for existing similar events (e.g. `add_to_cart`)

PR #1248, opened 2023-01-05
Release R139

spenes:
It adds `count.bad` field to be able to implement [transformer bad row count metric](https://github.com/snowplow/snowplow-rdb-loader/issues/1171).


istreeter (approved):
This is good.  Happy to approve it.  It simply adds a new field compared to the previous version.

I think strictly strictly strictly speaking, this would be a `2-1-0` schema.  Because `additionalProperties` is implicitly `true` for `count` object.  And [if you read the original schema-ver blog post](https://snowplow.io/blog/introducing-schemaver-for-semantic-versioning-of-schemas/)..... (yawn)

jbeemster (approved):

PR #1250, opened 2023-01-06
Add com.snowplowanalytics.snowplow/browser_context/jsonschema/1-0-0

### LINKED ISSUE #1249: Add com.snowplowanalytics.snowplow/browser_context/jsonschema/1-0-0

igneel64:
Adding the first version of the browser context schema.

close #1249

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/browser_context/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow/browser_context/jsonschema/1-0-0
@@ -1,0 +1,81 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for browser contexts",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow",
+    "name": "browser_context",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "viewport": {
+      "type": "string",
+      "maxLength": 20,
+      "description": "Viewport dimensions of the browser. Arrives in the form of WidthxHeight e.g. 1200x900."
+    },
+    "documentSize": {
+      "type": "string",
+      "maxLength": 20,
+      "description": "Document dimensions. Arrives in the form of WidthxHeight e.g. 1200x900"
+    },
+    "resolution": {
+      "type": "string",
+      "maxLength": 20,
+      "description": "Device native resolution. Arrives in the form of WidthxHeight e.g. 1200x900"
+    },
+    "colorDepth": {
+      "type": "integer",
+      "minimum": 0,
+      "maximum": 1000,
+      "description": "The number of bits allocated to colors for a pixel in the output device, excluding the alpha channel."
+    },
+    "devicePixelRatio": {
+      "type": ["number", "null"],
+      "minimum": 0,
+      "maximum": 1000,
+      "description": "Ratio of the resolution in physical pixels to the resolution in CSS pixels for the current display device."
+    },
+    "cookiesEnabled": {
+      "type": "boolean",
+      "description": "Indicates whether cookies are enabled or not. More info and caveats at https://developer.mozilla.org/en-US/docs/Web/API/Navigator/cookieEnabled."
+    },
+    "online": {
+      "type": "boolean",
+      "description": "Returns the online status of the browser. Important caveats are described in https://developer.mozilla.org/en-US/docs/Web/API/Navigator/onLine."
+    },
+    "browserLanguage": {
+      "type": ["string", "null"],
+      "maxLength": 20,
+      "description": "The preferred language of the user, usually the language of the browser UI. RFC 5646 https://datatracker.ietf.org/doc/html/rfc5646."
+    },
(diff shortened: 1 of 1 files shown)


schemas/com.snowplowanalytics.snowplow/browser_context/jsonschema/1-0-0
```diff
@@ -0,0 +1,76 @@
...
+      "maximum": 1000,
+      "description": "Number of logical processors available to run threads on the user's computer."
+    },
+    "tabId": {
+      "type": ["string", "null"],
+      "format": "uuid",
+      "description": "An identifier for the client browser tab the event is sent from."
+    }
+  },
+  "required": ["viewport", "documentSize", "cookiesEnabled", "online", "colorDepth", "resolution"],
```
paulboocock, line 74: Think we could make `tabId` required so we can rely on it being there if the browser_context is enabled?

igneel64: We have discussed that in the thread https://github.com/snowplow/snowplow-javascript-tracker/pull/1088#discussion_r954815662

There are cases where we use `null` as the value of the tabId. Of course feel free to add your take!

schemas/com.snowplowanalytics.snowplow/browser_context/jsonschema/1-0-0
```diff
@@ -0,0 +1,76 @@
...
+      "description": "Device native resolution. Arrives in the form of WidthxHeight e.g. 1200x900"
+    },
+    "colorDepth": {
+      "type": "integer",
+      "minimum": 0,
+      "maximum": 1000,
+      "description": "The number of bits allocated to colors for a pixel in the output device, excluding the alpha channel."
+    },
+    "devicePixelRatio": {
+      "type": ["integer", "null"],
```
matus-tomlein, line 34: Could this be a non-integer value as well? In addition to the 1 and 2 values, the docs say that "Other values may be returned as well in the case of unusually low resolution displays or, more often, when a screen has a higher pixel depth than double the standard resolution of 96 or 76 DPI.". So theoretically it could also be 0.XXX, right?

schemas/com.snowplowanalytics.snowplow/browser_context/jsonschema/1-0-0
```diff
@@ -0,0 +1,76 @@
...
+      "minimum": 0,
+      "maximum": 1000,
```
matus-tomlein, line 36: Nit: the `maximum` and `maxLength` values in most of these seem quite loose. If we don't know the bounds or are not sure then lets keep it like this, but in case there is a known limit that we could set, I think there is a potential it could help filter out some fabricated or bot events in some cases if we validate the values.

igneel64: The limits are kind of arbitrary with some padding on future values since I thought we could be a bit more lenient.
 _(Probably a DPR cannot reach 1000 soon 😆 )_

>  I think there is a potential it could help filter out some fabricated or bot events in some cases if we validate the values

So in that one you mean by setting a standard limit and let them fall through to the bad rows ?

schemas/com.snowplowanalytics.snowplow/browser_context/jsonschema/1-0-0
igneel64: It could be theoretically since it is the result of a division, but I am not aware of a case where CSS pixels could/would be set as 'fractional' to the physical pixels. The classic _'There is no half pixel'_, but certainly a browser engineer could have not thought about the case as well.

- Do you think we should set this as a `number` value ? 
- For special cases which are out of scope of 'normal' values, I believe it would be ok to have the validation provide a hint on this. Meaning events going to the bad rows. WDYT ?

matus-tomlein: I don't know but the fact that the documentation for the `devicePixelRatio` mentions the weird cases like that makes me think that we should make it a `number` type. I guess I would opt for being more lenient here, there are all kinds of weird scaling setups which I don't know much about.

matus-tomlein: Also the [docs say](https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio) that it is "A double-precision floating-point value".

schemas/com.snowplowanalytics.snowplow/browser_context/jsonschema/1-0-0
matus-tomlein: Right, that's what I meant – for the events with obviously weird values to go to bad rows. But it's a nitpick that has pretty marginal value so if it's difficult to say for sure, just leave the lenient setting.

matus-tomlein (approved):
LGTM!

Btw, sorry for not going back to this after your changes, I didn't get a notification when you resolved the comments. Next time, please ask for re-review when I should look again, so that I get the email to remind me.

paulboocock (approved):

matus-tomlein:
Cherry picked into R140 #1256

PR #1253, opened 2023-01-12
Add com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-3 (close #1252)

### LINKED ISSUE #1252: Add com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-3
This is a new version of the `mobile_context` schema tracked by the mobile trackers which will add the following properties:

* isPortrait – A Boolean indicating whether the device orientation is portrait (either upright or upside down)
* resolution – Screen resolution in pixels. Arrives in the form of WIDTHxHEIGHT (e.g., 1200x900)
* scale – Scale factor used to convert logical coordinates to device coordinates of the screen (uses UIScreen.scale on iOS and DisplayMetrics.density on Android)
* language – System language currently used on the device (ISO 639)

The `resolution`, `scale`, and `language` properties are added to reflect the new `browser_context` on Web (#1249) which moves some of the canonical properties from the events to the context entity. This is another step on the path to deprecate the canonical properties and unify the Web and mobile data models.

matus-tomlein:
Issue #1252

This is a new version of the `mobile_context` schema tracked by the mobile trackers which will add the following properties:

* isPortrait – A Boolean indicating whether the device orientation is portrait (either upright or upside down)
* resolution – Screen resolution in pixels. Arrives in the form of WIDTHxHEIGHT (e.g., 1200x900)
* scale – Scale factor used to convert logical coordinates to device coordinates of the screen (uses UIScreen.scale on iOS and DisplayMetrics.density on Android)
* language – System language currently used on the device (ISO 639)

The `resolution`, `scale`, and `language` properties are added to reflect the new `browser_context` on Web (#1249) which moves some of the canonical properties from the events to the context entity. This is another step on the path to deprecate the canonical properties and unify the Web and mobile data models.

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-3
+++ b/schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-3
@@ -1,0 +1,136 @@
+{
+	"$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+	"description": "Schema for mobile contexts",
+	"self": {
+		"vendor": "com.snowplowanalytics.snowplow",
+		"name": "mobile_context",
+		"format": "jsonschema",
+		"version": "1-0-3"
+	},
+	"type": "object",
+	"properties": {
+		"osType": {
+			"type": "string",
+			"description": "Operating system type (e.g., ios, tvos, watchos, osx, android)"
+		},
+		"osVersion": {
+			"type": "string",
+			"description": "The current version of the operating system"
+		},
+		"deviceManufacturer": {
+			"type": "string",
+			"description": "The manufacturer of the product/hardware"
+		},
+		"deviceModel": {
+			"type": "string",
+			"description": "The end-user-visible name for the end product"
+		},
+		"carrier": {
+			"type": ["string", "null"],
+			"description": "The carrier of the SIM inserted in the device"
+		},
+		"networkType": {
+			"type": ["string", "null"],
+			"enum": ["mobile", "wifi", "offline", null],
+			"description": "Type of network the device is connected to"
+		},
+		"networkTechnology": {
+			"type": ["string", "null"],
+			"description": "Radio access technology that the device is using"
+		},
+		"openIdfa": {
+			"type": ["string", "null"],
+			"description": "Deprecated tracking identifier for iOS"
+		},
+		"appleIdfa": {
+			"type": ["string", "null"],
+			"description": "Advertising identifier on iOS"
+		},
+		"appleIdfv": {
+			"type": ["string", "null"],
+			"description": "UUID identifier for vendors on iOS"
+		},
+		"androidIdfa": {
+			"type": ["string", "null"],
+			"description": "Advertising identifier on Android"
+		},
+		"physicalMemory": {
+			"type": ["integer", "null"],
+			"minimum": 0,
+			"maximum": 9223372036854775807,
+			"description": "Total physical system memory in bytes"
+		},
+		"systemAvailableMemory": {
+			"type": ["integer", "null"],
+			"minimum": 0,
+			"maximum": 9223372036854775807,
+			"description": "Available memory on the system in bytes (Android only)"
+		},
+		"appAvailableMemory": {
+			"type": ["integer", "null"],
+			"minimum": 0,
+			"maximum": 9223372036854775807,
+			"description": "Amount of memory in bytes available to the current app (iOS only)"
+		},
+		"batteryLevel": {
+			"type": ["integer", "null"],
+			"minimum": 0,
+			"maximum": 100,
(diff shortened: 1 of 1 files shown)


igneel64 (approved):

schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-3
```diff
@@ -0,0 +1,126 @@
...
+		"totalStorage": {
+			"type": ["integer", "null"],
+			"minimum": 0,
+			"maximum": 9223372036854775807,
+			"description": "Total size of storage in bytes"
+		},
+		"isPortrait": {
+			"type": ["boolean", "null"],
+			"description": "A Boolean indicating whether the device orientation is portrait (either upright or upside down)"
+		},
```
paulboocock, line 106: Can you only get a boolean here, no way to understand the actual device orientation? (i.e. upsidedown)

schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-3
```diff
@@ -0,0 +1,126 @@
...
+		"resolution": {
+			"type": ["string", "null"],
+			"maxLength": 20,
+			"description": "Screen resolution in pixels. Arrives in the form of WIDTHxHEIGHT (e.g., 1200x900)"
+		},
```
paulboocock, line 111: Given this is device information, I think I'd imagine it to stay static, but I'm curious to know if you'd expect it to change as device orientation changes or stay the same? Might be worth adding the behvaiour to the description.

schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-3
matus-tomlein: | a | b |
| a | b |

matus-tomlein: It is possible to understand the orientation in more detail but the problem is that [iOS](https://developer.apple.com/documentation/uikit/uideviceorientation) and [Android](https://developer.android.com/reference/androidx/browser/trusted/ScreenOrientation) report it in a different way and I wasn't able to find a mapping between the states.

Here is a table with the values, I was only able to confidently map the portrait positions but not sure whether the natural landscape on Android is left or right on iOS and there are also extra positions on each platform not on the other (such as face up or down).

| iOS | Android | Description |
|---|---|---|
| portrait | portrait_primary | Home button at the bottom |
| portraitUpsideDown | portrait_secondary | Home button at the top |
| landscapeLeft | ? | Home button on the right side |
| landscapeRight | ? | Home button on the left side  |
| faceUp | ? | parallel to the ground with the screen facing upwards |
| faceDown | ? | parallel to the ground with the screen facing downwards |
| ? | landscape | can represent landscape-primary, landscape-secondary or both |
| ? | landscape_primary | natural orientation for landscape | 
| ? | landscape_secondary | rotated 180° from its natural orientation | 
| ? | natural | either portrait-primary or landscape-primary depending on the device's usual orientation |
| ? | portrait | portrait-primary, portrait-secondary or both |

schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-3
matus-tomlein: Good point! It doesn't change with the orientation, I added a note in description.

matus-tomlein:
@igneel64 @paulboocock I have added a couple of properties for the Android app set ID that we plan to add in [this issue](https://github.com/snowplow/snowplow-android-tracker/issues/565). Could you please take another look at the schema? Would be great to put together a release with `mobile_context`, `browser_context`, and `referrer_details` schemas.

schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-3
igneel64: I am ok with the boolean since this is the current state ^^

igneel64 (approved):
I am fine with the current state as well :) 
Should we cherry pick the browser_context addition in this branch to mark it as the release ?

matus-tomlein:
@igneel64 We usually make a new release PR where we cherry-pick the commits, let's wait for Paul regarding the comment he raised and we can go ahead with that 👍

schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-3
paulboocock: Oh Android.

I think a boolean is probably a good idea 😂

matus-tomlein:
Cherry picked into R140 #1256

PR #1255, opened 2023-01-13
Add com.android.installreferrer.api/referrer_details/jsonschema/1-0-0 (close #1254)

matus-tomlein:
Issue #1254

Add schema for the install referrer information provided by the [Google Play Install Referrer Library](https://developer.android.com/google/play/installreferrer/library) for Android apps. The library enables mobile app install attribution by surfacing the referrer URL which led to the app being installed from the Play Store. This URL may contain `umd` or other parameters to attribute the install to a campaign.

Unfortunately, this is an Android-only feature, there is no equivalent on iOS.

Issue #1254, opened 2023-01-13
Add com.android.installreferrer.api/referrer_details/jsonschema/1-0-0


matus-tomlein:
Add schema for the install referrer information provided by the [Google Play Install Referrer Library](https://developer.android.com/google/play/installreferrer/library) for Android apps. The library enables mobile app install attribution by surfacing the referrer URL which led to the app being installed from the Play Store. This URL may contain `umd` or other parameters to attribute the install to a campaign.

Unfortunately, this is an Android-only feature, there is no equivalent on iOS.

PR #1256, opened 2023-01-30
Release R140

matus-tomlein:
This PR was done with @igneel64 and adds a new schema for `browser_context` with information tracked by the JS tracker that used to be stored in canonical event properties and also some new info (e.g., `tabId`). It also adds a new schema for `install_referrer` to be tracked on Android.

Change | Issue | Issue PR
-- | -- | --
Add com.snowplowanalytics.snowplow/browser_context/jsonschema/1-0-0 | #1249 | #1250
Add com.android.installreferrer.api/referrer_details/jsonschema/1-0-0 | #1254 | #1255

### DIFF
--- a/schemas/com.android.installreferrer.api/referrer_details/jsonschema/1-0-0
+++ b/schemas/com.android.installreferrer.api/referrer_details/jsonschema/1-0-0
@@ -1,0 +1,34 @@
+{
+	"$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+	"description": "Represents an install referrer details for Android apps installed from the Play Store (see https://developer.android.com/reference/com/android/installreferrer/api/ReferrerDetails)",
+	"self": {
+		"vendor": "com.android.installreferrer.api",
+		"name": "referrer_details",
+		"format": "jsonschema",
+		"version": "1-0-0"
+	},
+	"type": "object",
+	"properties": {
+		"installReferrer": {
+			"type": ["string", "null"],
+			"maxLength": 4096,
+			"description": "The referrer URL of the installed package"
+		},
+		"referrerClickTimestamp": {
+			"type": ["string", "null"],
+			"format": "date-time",
+			"description": "The timestamp when referrer click happens"
+		},
+		"installBeginTimestamp": {
+			"type": ["string", "null"],
+			"format": "date-time",
+			"description": "The timestamp when installation begins"
+		},
+		"googlePlayInstantParam": {
+			"type": "boolean",
+			"description": "Boolean indicating if the user has interacted with the app's instant experience in the past 7 days"
+		}
+	},
+	"required": ["googlePlayInstantParam"],
+	"additionalProperties": false
+}
--- a/schemas/com.snowplowanalytics.snowplow.ecommerce/promotion/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow.ecommerce/promotion/jsonschema/1-0-0
@@ -1,0 +1,62 @@
+{
+    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "description": "Schema for a promotion entity in Ecommerce",
+    "self": {
+        "vendor": "com.snowplowanalytics.snowplow.ecommerce",
+        "name": "promotion",
+        "format": "jsonschema",
+        "version": "1-0-0"
+    },
+    "type": "object",
+    "properties": {
+        "id": {
+            "description" : "The ID of the promotion.",
+            "type": "string",
+            "maxLength": 4096
+        },
+        "name": {
+            "description" : "The name of the promotion.",
+            "type": [ "string", "null" ],
+            "maxLength": 256
+        },
+        "product_ids" : {
+            "description" : "Array of SKUs or product IDs showcased in the promotion.",
+            "type" : [ "array", "null" ],
+            "items" : {
+                "type" : "string",
+                "maxLength": 4096
+            }
+        },
+        "position": {
+            "description" : "The position the promotion was presented in a list of promotions E.g. banner, slider.",
+            "type" : [
+                "integer",
+                "null"
+            ],
+            "minimum": 0,
+            "maximum": 9999999
+        },
+        "creative_id": {
+            "description" : "Identifier/Name/Url for the creative presented on the promotion.",
+            "type" : [
+                "string",
+                "null"
+            ],
+            "maxLength": 4096
+        },
+        "type": {
+            "description": "Type of the promotion delivery mechanism. E.g. popup, banner, intra-content",
+            "type": [ "string", "null" ],
+            "maxLength": 256
+        },
+        "slot": {
+            "description": "The website slot in which the promotional content was added to. E.g. Identifier for slot sidebar-1, intra-content-2",
+            "type": [ "string", "null" ],
+            "maxLength": 256
+        }
+    },
+    "required": [
+        "id"
+    ],
+    "additionalProperties": false
+}
--- a/schemas/com.snowplowanalytics.snowplow/browser_context/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow/browser_context/jsonschema/1-0-0
@@ -1,0 +1,81 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for browser contexts",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow",
+    "name": "browser_context",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "viewport": {
+      "type": "string",
+      "maxLength": 20,
+      "description": "Viewport dimensions of the browser. Arrives in the form of WidthxHeight e.g. 1200x900."
+    },
+    "documentSize": {
+      "type": "string",
+      "maxLength": 20,
+      "description": "Document dimensions. Arrives in the form of WidthxHeight e.g. 1200x900"
+    },
+    "resolution": {
+      "type": "string",
+      "maxLength": 20,
+      "description": "Device native resolution. Arrives in the form of WidthxHeight e.g. 1200x900"
+    },
+    "colorDepth": {
+      "type": "integer",
+      "minimum": 0,
+      "maximum": 1000,
+      "description": "The number of bits allocated to colors for a pixel in the output device, excluding the alpha channel."
+    },
+    "devicePixelRatio": {
+      "type": ["number", "null"],
+      "minimum": 0,
+      "maximum": 1000,
+      "description": "Ratio of the resolution in physical pixels to the resolution in CSS pixels for the current display device."
+    },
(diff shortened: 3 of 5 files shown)


igneel64:
LGTM, just the changelog and it's great!

matus-tomlein:
@istreeter @jbeemster @paulboocock sorry for the tag, just wondering if we can move forward with the release or is there anything to address before?

istreeter (commented):
The schemas look great, I cannot find any fault in them.

But unfortunately there is a problem that is making me a bit nervous.  The RDB loader [has a bug](https://github.com/snowplow/snowplow-rdb-loader/issues/1065), in which loading to Redshift can break when new fields are added to a schema.  The bug is years old, but we are finally addressing it this quarter.

How urgent is it to get these extra fields added?  I'm wondering if it's worth waiting for a fixed RDB loader before trying to add the new fields?  Especially as `mobile_context` is such a commonly used schema.

I don't want to block this if it's urgent.  But if we go ahead and merge this, then we need to be prepared that some pipelines might need to be unstuck by re-shredding a failed batch.

matus-tomlein:
That's good to know @istreeter! I think we can wait for the RDB loader bug to be fixed. We also planned to release the mobile context changes in Q1, so hopefully we can do that right after the bug fix.

I separated the `mobile_context` schema update into release R141 that is now in a draft PR #1258. Kept here just the new schemas for `browser_context` and `install_referrer`.

istreeter (approved):

igneel64:
@istreeter Can you please take a look at three new schemas that have been included in the release ? _(the ecommerce ones)_

schemas/com.snowplowanalytics.snowplow.ecommerce/promotion/jsonschema/1-0-0
```diff
@@ -0,0 +1,62 @@
...
+            "maxLength": 4096
+        },
+        "name": {
+            "description" : "The name of the promotion.",
+            "type": [ "string", "null" ],
+            "maxLength": 256
+        },
+        "product_ids" : {
+            "description" : "Array of SKUs or product IDs showcased in the promotion.",
+            "type" : [ "array", "null" ],
```
istreeter, line 24: Just so you know, the way we handle arrays in redshift is non-ideal.  The `product_ids` column will be a VARCHAR column with max length 65535.  We serialize the array of strings as a JSON array and load it into the VARCHAR column.

If the serialized array exceeds 65535 characters, then the string gets truncated.  So it's possible the column contains a value like this (although longer):

```
["Lorem", "ipsum", "d
```

Having said that, load of Iglu Central schemas contain array fields already.  So I don't suggest you change anything.  Just be aware of it when it comes to data modelling.

istreeter (commented):
The new ecommerce schemas look good too.

schemas/com.snowplowanalytics.snowplow.ecommerce/promotion/jsonschema/1-0-0
igneel64: @emielver Are we ok with that ?

rlh1994: I think that should be fine, if anyone does every encounter this issue we could do some specific string processing on it for redshift, but we don't do anything for other columns at the moment so we should be okay.

schemas/com.snowplowanalytics.snowplow.ecommerce/refund/jsonschema/1-0-0
```diff
@@ -0,0 +1,43 @@
...
+        "currency": {
+            "description" : "The currency in which the product is being priced (ISO 4217).",
+            "type" : "string",
+            "maxLength": 3,
+            "minLength": 3
+        },
+        "refund_amount" : {
+            "type": "number",
+            "description": "The monetary amount refunded.",
+            "multipleOf": 0.01,
```
miike, line 26: Nit: There are currencies (the dinar) with three decimal places. I think we are safe to remove `multipleOf` here.

schemas/com.snowplowanalytics.snowplow/browser_context/jsonschema/1-0-0
```diff
@@ -0,0 +1,81 @@
...
+    },
+    "resolution": {
+      "type": "string",
+      "maxLength": 20,
+      "description": "Device native resolution. Arrives in the form of WidthxHeight e.g. 1200x900"
+    },
+    "colorDepth": {
+      "type": "integer",
+      "minimum": 0,
+      "maximum": 1000,
```
miike, line 30: If we bump this down to 128 we can take advantage of SMALLINT (2 bytes) in Redshift.

schemas/com.snowplowanalytics.snowplow/browser_context/jsonschema/1-0-0
```diff
@@ -0,0 +1,81 @@
...
+  "description": "Schema for browser contexts",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow",
+    "name": "browser_context",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "viewport": {
```
miike, line 12: Any risk that some of these might be nullable in a headless browser?

schemas/com.android.installreferrer.api/referrer_details/jsonschema/1-0-0
```diff
@@ -0,0 +1,36 @@
...
+		"installReferrer": {
+			"type": ["string", "null"],
+			"maxLength": 4096,
+			"description": "The referrer URL of the installed package"
+		},
+		"referrerClickTimestampSeconds": {
+			"type": "integer",
+			"minimum": 0,
+			"maximum": 9223372036854775807,
+			"description": "The timestamp in seconds when referrer click happens"
```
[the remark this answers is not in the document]
matus-tomlein: That's a good point, I originally wanted to keep it in the same format as the Android APIs provides the timestamps, but actually why not make them easier to consume. Changed the timestamps to date-time formatted strings.

schemas/com.snowplowanalytics.snowplow/browser_context/jsonschema/1-0-0
igneel64: I cannot think of a screen for a few years that will have 128-bit color depth, except if it is a browser bug. Considering this will be a highly used context, I think the tradeoff could be worth it.
Will change.

schemas/com.snowplowanalytics.snowplow.ecommerce/refund/jsonschema/1-0-0
igneel64: Really good observation. AFAIK we use this in other places which has some efficiencies (a [past comment](https://github.com/snowplow/iglu-central/pull/1243/files#r1048228802)), but does that mean we cannot take advantage of multipleOf for monetary values ? _There are some other places that I saw this pattern used in the past._

I will remove it.

schemas/com.snowplowanalytics.snowplow/browser_context/jsonschema/1-0-0
igneel64: On headless browsers there are some default values e.g. for pptr it is 800x600. I am not sure it could be made null since it is part of the CDP and can only be initiated with a number _(in normal circumstances)_.

We can be more lenient and make it nullable but not sure if we are being too cautious and allowing for exceptional cases to slip in the data and not the bad rows. WDYT ?

miike: Probably too cautious, let's leave as is I think.

schemas/com.snowplowanalytics.snowplow/browser_context/jsonschema/1-0-0
miike: Yeah, and if it's a bug in a browser I think we probably want to generate a failed event anyway.

schemas/com.snowplowanalytics.snowplow.ecommerce/refund/jsonschema/1-0-0
miike: We could maybe go to 3 decimal places but it is a fair bit of an edge case, and limiting the scale of a decimal is useful in terms of storage, and we do want to make sure that for currencies that schemaddl results in a decimal, rather than a float.

schemas/com.snowplowanalytics.snowplow/browser_context/jsonschema/1-0-0
igneel64: Re:
Checked our database column translation docs and seems to be fine since the value is less than `32767`.

image

schemas/com.snowplowanalytics.snowplow.ecommerce/refund/jsonschema/1-0-0
igneel64: @istreeter We need your input here on if it will play well to set the multipleOf attribute to `0.001` for our monetary values.

Could we also for the future plan to change the multipleOf attributes from `0.01` to `0.001` ?

istreeter: I definitely recommend multiple of _something_ so that it gets stored as a decimal type.  It is not safe to change the multipleOf in a patch release of the schema, so it is important to think about this carefully for the 1-0-0 schema.

**For Databricks/parquet**: either of these combinations will be stored as a 32 bit decimal value:

```
multipleOf: 0.01
maximum: 9999999

-- or --

multipleOf: 0.001
maximum: 999999
```

You can of course use a larger maximum, and it will get stored as a 64 bit value.  Which is probably OK.

**For BigQuery**: all decimals are stored as 64 bit decimals, so everything from above still applies, but it is less important to limit the maximum.

**For Redshift**: This is disappointing and you won't like it:

- For `multipleOf: 0.01` we store it as a 64 bit decimal value ([link](https://github.com/snowplow/schema-ddl/blob/0.17.1/modules/core/src/main/scala/com.snowplowanalytics/iglu.schemaddl/redshift/generators/TypeSuggestions.scala#L92-L93))
- For multiple of _any other value_ (e.g. of 0.001) then it gets stored as a Double :(

So it seems you need to choose... either multipleOf 0.01 and then you get nice behaviour in all warehouses.  Or multipleOf 0.01, and then you get nice behaviour in most warehouses except for Redshift.

istreeter: I was thinking a bit more... I think we could safely update schema-ddl so it nicely handles `multipleOf: 0.001` for Redshift and BigQuery.  To get the benefit of the change, pipelines will need to be updated to a new version of RDB loader or BQ loader, before starting to send events with this schema.

Let me know if you want us to work on this small change.

matus-tomlein: I have moved the `refund` schema into R141 (PR #1258) to be released after this issue is resolved.

matus-tomlein:
@miike can you please take another look? We have moved the `refund` schema into R141 so there shouldn't be any more blockers for releasing this.

miike:
I'm happy with this.

cc'ing @jbeemster to help out with getting this released.

miike (approved):

Issue #1269, opened 2023-03-24
Add com.snowplowanalytics.snowplow/media_player_quality/jsonschema/1-0-0


matus-tomlein:
Add schema for a context entity that is tracked with media player events on data quality change.

Issue #1268, opened 2023-03-24
Add com.snowplowanalytics.snowplow.media/ad_break/jsonschema/1-0-0


matus-tomlein:
Add schema for a context entity that is shared for all media player ad events that belong to the ad break.

Issue #1267, opened 2023-03-24
Add com.snowplowanalytics.snowplow.media/ad/jsonschema/1-0-0


matus-tomlein:
Add schema for a context entity with information about the currently played ad. The schema will contain metadata about the ad such as it's ID and name.

Issue #1266, opened 2023-03-24
Add com.snowplowanalytics.snowplow.media/session/jsonschema/1-0-0


matus-tomlein:
Add schema for media player session context entity that is used to identify all events belonging to the same video player. Additionally, the schema will provide a few properties for edge analytics (such as the play time).

PR #1270, opened 2023-03-24
Add media player entity schemas

matus-tomlein:
Schemas for new context entities that will be attached to media player events to enable:

* tracking a session that groups all events coming from the same media player,
* tracking ads within media,
* tracking media player quality.

## `media_player_session` (#1266)

A session groups events from the same media player using a `sessionId` identifier (which may be provided by the user).

A secondary role of the session is a sort of edge analytics since it provides statistics about the playback (such as the time spent playing) in order to make data modelling more efficient.

## `media_player_ad_break` (#1268)

An ad break is a grouping of ads played back to back. The schema will identify all ad events belonging to the same ad break.

## `media_player_ad` (#1267)

Describes an ad being played at the time of the event. Provide metadata about the ad.

## `media_player_quality` (#1269)

Describes quality of experience for the media playback at a specific time. It will be tracked with `updateQuality` media player events so it will give a snapshot of the data quality rather than aggregated metrics for the whole playback.

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow.media/ad/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow.media/ad/jsonschema/1-0-0
@@ -1,0 +1,63 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for a context entity with information about the currently played ad",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow.media",
+    "name": "ad",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "name": {
+      "description": "Friendly name of the ad.",
+      "type": [
+        "null",
+        "string"
+      ],
+      "maxLength": 4096
+    },
+    "adId": {
+      "type": "string",
+      "maxLength": 256,
+      "description": "Unique identifier for the ad."
+    },
+    "creativeId": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "maxLength": 4096,
+      "description": "The ID of the ad creative."
+    },
+    "podPosition": {
+      "type": [
+        "integer",
+        "null"
+      ],
+      "description": "The position of the ad within the ad break, starting with 1.",
+      "minimum": 1,
+      "maximum": 65535
+    },
+    "duration": {
+      "type": [
+        "number",
+        "null"
+      ],
+      "description": "Length of the video ad in seconds.",
+      "minimum": 0,
+      "maximum": 2147483647
+    },
+    "skippable": {
+      "type": [
+        "boolean",
+        "null"
+      ],
+      "description": "Indicating whether skip controls are made available to the end user."
+    }
+  },
+  "additionalProperties": false,
+  "required": [
+    "adId"
+  ]
+}
--- a/schemas/com.snowplowanalytics.snowplow.media/ad_break/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow.media/ad_break/jsonschema/1-0-0
@@ -1,0 +1,59 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for a context entity, shared with all ad events belonging to the ad break.",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow.media",
+    "name": "ad_break",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "name": {
+      "description": "Ad break name (e.g., pre-roll, mid-roll, and post-roll).",
+      "type": [
+        "null",
+        "string"
+      ],
+      "maxLength": 4096
+    },
+    "breakId": {
+      "type": "string",
+      "maxLength": 256,
+      "description": "An identifier for the ad break."
+    },
+    "startTime": {
+      "type": "number",
+      "description": "Playback time in seconds at the start of the ad break.",
+      "minimum": 0,
+      "maximum": 2147483647
+    },
+    "breakType": {
+      "description": "Type of ads within the break: linear (take full control of the video for a period of time), nonlinear (run concurrently to the video), companion (accompany the video but placed outside the player).",
+      "enum": [
+        "linear",
+        "nonlinear",
+        "companion",
+        null
+      ],
+      "type": [
+        "string",
+        "null"
+      ]
+    },
+    "podSize": {
+      "type": [
+        "integer",
+        "null"
+      ],
+      "description": "The number of ads to be played within the ad break.",
+      "minimum": 0,
+      "maximum": 65535
+    }
+  },
+  "additionalProperties": false,
+  "required": [
+    "breakId",
+    "startTime"
+  ]
+}
--- a/schemas/com.snowplowanalytics.snowplow.media/player/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow.media/player/jsonschema/1-0-0
@@ -1,0 +1,131 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for a context entity for media events that describes the media player and playback state",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow.media",
+    "name": "player",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "currentTime": {
+      "type": "number",
+      "description": "The current playback time position within the media in seconds.",
+      "minimum": 0,
+      "maximum": 2147483647
+    },
+    "duration": {
+      "type": [
+        "number",
+        "null"
+      ],
+      "description": "A floating-point value indicating the duration of the media in seconds.",
+      "minimum": 0,
+      "maximum": 2147483647
+    },
+    "ended": {
+      "type": "boolean",
+      "description": "Whether playback of the media has ended."
+    },
+    "fullscreen": {
+      "type": [
+        "boolean",
+        "null"
+      ],
+      "description": "Whether the video element is fullscreen."
+    },
+    "livestream": {
+      "type": [
+        "boolean",
+        "null"
+      ],
+      "description": "Whether the media is a live stream."
+    },
+    "label": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "description": "Human readable name given to tracked media content.",
+      "maxLength": 4096
+    },
+    "loop": {
+      "type": [
+        "boolean",
+        "null"
+      ],
+      "description": "Whether the video should restart after ending."
+    },
+    "mediaType": {
+      "description": "Type of media content.",
+      "enum": [
+        "video",
+        "audio",
+        null
+      ],
+      "type": [
+        "string",
+        "null"
+      ]
+    },
+    "muted": {
+      "type": [
+        "boolean",
+        "null"
+      ],
+      "description": "Whether the media element is muted."
+    },
+    "paused": {
+      "type": "boolean",
+      "description": "Whether the media element is paused"
+    },
+    "pictureInPicture": {
+      "type": [
+        "boolean",
+        "null"
+      ],
+      "description": "Whether the video element is showing picture-in-picture."
+    },
+    "playbackRate": {
+      "type": [
+        "number",
+        "null"
+      ],
+      "description": "Playback rate (1 is normal).",
+      "minimum": 0,
+      "maximum": 16
+    },
+    "playerType": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "description": "Type of the media player (e.g., com.youtube-youtube, com.vimeo-vimeo, org.whatwg-media_element).",
+      "maxLength": 4096
+    },
+    "quality": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "description": "Quality level of the playback (e.g., 1080p).",
+      "maxLength": 4096
+    },
+    "volume": {
+      "type": [
+        "integer",
+        "null"
+      ],
+      "description": "Volume percentage at which the media will be played.",
+      "minimum": 0,
+      "maximum": 100
+    }
+  },
+  "additionalProperties": false,
+  "required": [
+    "currentTime",
+    "ended",
+    "paused"
+  ]
+}
--- a/schemas/com.snowplowanalytics.snowplow.media/session/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow.media/session/jsonschema/1-0-0
@@ -1,0 +1,138 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for a context entity for media player events that tracks a session of a single media player usage",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow.media",
+    "name": "session",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "mediaSessionId": {
+      "type": "string",
+      "maxLength": 256,
+      "description": "An identifier for the media session that is kept while the media content is played in the media player."
+    },
+    "startedAt": {
+      "type": [
+        "null",
+        "string"
+      ],
+      "description": "Local date-time timestamp of when the session started.",
+      "format": "date-time"
+    },
+    "pingInterval": {
+      "type": [
+        "integer",
+        "null"
+      ],
+      "description": "Interval (seconds) in which the ping events will be sent. Default (specified in the tracker media docs) is assumed if not specified.",
+      "minimum": 0,
+      "maximum": 65535
+    },
+    "timePlayed": {
+      "type": [
+        "number",
+        "null"
+      ],
+      "description": "Total seconds user spent playing content (excluding linear ads).",
+      "minimum": 0,
+      "maximum": 2147483647
+    },
+    "timePlayedMuted": {
+      "type": [
+        "number",
+        "null"
+      ],
+      "description": "Total seconds user spent playing content on mute (excluding linear ads).",
+      "minimum": 0,
+      "maximum": 2147483647
+    },
+    "timePaused": {
+      "type": [
+        "number",
+        "null"
+      ],
+      "description": "Total seconds user spent with paused content (excluding linear ads).",
+      "minimum": 0,
+      "maximum": 2147483647
+    },
+    "contentWatched": {
+      "type": [
+        "number",
+        "null"
+      ],
+      "description": "Total seconds of the content played. Each part of the content played is counted once (i.e., counts rewinding or rewatching the same content only once). Playback rate does not affect this value.",
+      "minimum": 0,
+      "maximum": 2147483647
+    },
+    "timeBuffering": {
+      "type": [
+        "number",
+        "null"
+      ],
+      "description": "Total seconds that playback was buffering during the session.",
+      "minimum": 0,
+      "maximum": 2147483647
+    },
+    "timeSpentAds": {
+      "type": [
+        "number",
+        "null"
+      ],
+      "description": "Total seconds that ads played during the session.",
+      "minimum": 0,
+      "maximum": 2147483647
+    },
+    "ads": {
+      "type": [
+        "integer",
+        "null"
+      ],
+      "description": "Number of ads played.",
+      "minimum": 0,
+      "maximum": 65535
+    },
+    "adsClicked": {
+      "type": [
+        "integer",
+        "null"
+      ],
+      "description": "Number of ads that the user clicked on.",
+      "minimum": 0,
+      "maximum": 65535
+    },
+    "adsSkipped": {
+      "type": [
+        "integer",
+        "null"
+      ],
+      "description": "Number of ads that the user skipped.",
+      "minimum": 0,
+      "maximum": 65535
+    },
+    "adBreaks": {
+      "type": [
+        "integer",
+        "null"
+      ],
+      "description": "Number of ad breaks played.",
+      "minimum": 0,
+      "maximum": 65535
+    },
+    "avgPlaybackRate": {
+      "type": [
+        "number",
+        "null"
+      ],
+      "description": "Average playback rate (1 is normal speed).",
+      "minimum": 0,
+      "maximum": 16
+    }
+  },
+  "additionalProperties": false,
+  "required": [
+    "mediaSessionId"
+  ]
+}


schemas/com.snowplowanalytics.snowplow/media_player_quality/jsonschema/1-0-0
```diff
@@ -0,0 +1,58 @@
...
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "bitrate": {
+      "type": [
+        "number",
+        "null"
+      ],
+      "description": "The current bitrate in bps.",
```
greg-el, line 17: Might be nice to use `Bits Per Second` rather than `bps`.

greg-el (commented):
A couple of things I picked up on from an initial skim over the schemas.

schemas/com.snowplowanalytics.snowplow/media_player_ad/jsonschema/1-0-0
```diff
@@ -0,0 +1,56 @@
...
+      "maximum": 65535
+    },
+    "duration": {
+      "type": [
+        "number",
+        "null"
+      ],
+      "description": "Length of the video ad in seconds.",
+      "minimum": 0,
+      "maximum": 9007199254740991
```
rlh1994, line 49: Could we be more explicit and make these integers?

schemas/com.snowplowanalytics.snowplow/media_player_ad_break/jsonschema/1-0-0
```diff
@@ -0,0 +1,57 @@
...
+    },
+    "breakId": {
+      "type": "string",
+      "format": "uuid",
+      "description": "An identifier for the ad break."
+    },
+    "position": {
+      "type": "number",
+      "description": "The position of the ad break within the content, starting with 1.",
+      "minimum": 0,
```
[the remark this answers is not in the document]
rlh1994: Also, what does this actually mean? As in this is the nth ad break? I'd probably rephrase the description in that case as it wasn't immediatly clear to me that's what this meant.

schemas/com.snowplowanalytics.snowplow/media_player_quality/jsonschema/1-0-0
```diff
@@ -0,0 +1,58 @@
...
+      "minimum": 0,
+      "maximum": 9007199254740991
+    },
+    "droppedFrames": {
+      "type": [
+        "number",
+        "null"
+      ],
+      "description": "The number of dropped frames in the stream.",
```
rlh1994, line 26: Maybe just specify if this is cumulative or since the last event?

schemas/com.snowplowanalytics.snowplow/media_player_session/jsonschema/1-0-0
```diff
@@ -0,0 +1,84 @@
...
+      "format": "uuid",
+      "description": "An identifier for the session."
+    },
+    "startedAt": {
+      "type": [
+        "null",
+        "string"
+      ],
+      "description": "Date-time timestamp of when the session started.",
+      "format": "date-time"
```
rlh1994, line 23: Should this not be a timestamp type?

schemas/com.snowplowanalytics.snowplow/media_player_session/jsonschema/1-0-0
```diff
@@ -0,0 +1,84 @@
...
+    },
+    "pingIterval": {
+      "type": [
+        "number",
+        "null"
+      ],
+      "description": "Interval (seconds) in which the ping events will be sent. Default (specified in the tracker media docs) is assumed if not specified.",
+      "minimum": 0,
+      "maximum": 65535
```
rlh1994, line 32: integer?

schemas/com.snowplowanalytics.snowplow/media_player_session/jsonschema/1-0-0
```diff
@@ -0,0 +1,84 @@
...
+      "maximum": 2147483647
+    },
+    "totalAds": {
+      "type": [
+        "number",
+        "null"
+      ],
+      "description": "Number of ads played.",
+      "minimum": 0,
+      "maximum": 65535
```
rlh1994, line 77: Do we want to add the number of ad breaks as well? Also, integer?

schemas/com.snowplowanalytics.snowplow/media_player_ad/jsonschema/1-0-0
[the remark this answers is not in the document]
matus-tomlein: That's a very good point. The VAST schema (a standard for representing ads) has a `skipoffset` property which specifies the position when the ad skip controls appear, but I think that's too granular for us. So I opted for the `skippable` boolean.

schemas/com.snowplowanalytics.snowplow/media_player_ad/jsonschema/1-0-0
matus-tomlein: I wanted to make this have the same format as the `duration` property in the `media_player` schema with info about the main media content. That one allows for double precision values, so I thought we should do the same here.

schemas/com.snowplowanalytics.snowplow/media_player_ad_break/jsonschema/1-0-0
```diff
@@ -0,0 +1,57 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for a context entity that is shared for all media player ad events that belong to the ad break",
```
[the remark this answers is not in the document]
matus-tomlein: Sounds better, changed!

schemas/com.snowplowanalytics.snowplow/media_player_ad_break/jsonschema/1-0-0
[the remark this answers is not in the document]
matus-tomlein: Thinking more about this, I don't really have a good use case for the ad break position property... I took from another schema as it seemed useful to me but now I can't find a good reason for having it. We already have the `startTime` of the ad break so we should be able to order them. Also I added a `totalAdBreaks` property to the `media_player_session` so we can easily know how many ad breaks there were. I removed it from the schema.

schemas/com.snowplowanalytics.snowplow/media_player_quality/jsonschema/1-0-0
matus-tomlein: Replaced `bps` with `bits per second`.

schemas/com.snowplowanalytics.snowplow/media_player_quality/jsonschema/1-0-0
matus-tomlein: Specified since last report as that would be my preference (happy to change if there are other opinions).

schemas/com.snowplowanalytics.snowplow/media_player_session/jsonschema/1-0-0
matus-tomlein: I think a date-time is more user friendly. This info is the actual time of the start of the session, it doesn't relate to the playback time so I think it's better to represent it in a readable format.

schemas/com.snowplowanalytics.snowplow/media_player_session/jsonschema/1-0-0
matus-tomlein: Right, fixed.

schemas/com.snowplowanalytics.snowplow/media_player_session/jsonschema/1-0-0
matus-tomlein: I think that makes sense, added `totalAdBreaks` and fixed integer type.

schemas/com.snowplowanalytics.snowplow/media_player_quality/jsonschema/1-0-0
```diff
@@ -0,0 +1,58 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for a context entity that is tracked with media player events on data quality change",
```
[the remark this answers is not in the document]
matus-tomlein: I think both cases – triggered by manual selection of quality by user but also by an automatic change in the player. We should have different event types for the two cases – maybe `updateQuality` and `userUpdateQuality`?

schemas/com.snowplowanalytics.snowplow/media_player_ad/jsonschema/1-0-0
rlh1994: Sorry I meant the `podPosition` instead?

schemas/com.snowplowanalytics.snowplow/media_player_quality/jsonschema/1-0-0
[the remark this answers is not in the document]
miike: Agreed on separating. Automatic updates might get quite noisy whereas user ones are more likely to be useful.

schemas/com.snowplowanalytics.snowplow/media_player_ping_event/jsonschema/1-0-0
```diff
@@ -0,0 +1,23 @@
...
+  "description": "Schema for a media ping event sent in regular intervals during playback.",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow",
+    "name": "media_player_ping_event",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "label": {
```
miike, line 12: Might be nice if we can re-use this for percentage based progress as well. Do we want to have units somewhere?

schemas/com.snowplowanalytics.snowplow/media_player_ad_event/jsonschema/1-0-0
```diff
@@ -0,0 +1,31 @@
...
+  "description": "Schema for a media event related to ads during the media playback",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow",
+    "name": "media_player_ad_event",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "type": {
```
miike, line 12: I would go so far as to separate this out into different schemas entirely (e.g., ad_start, ad_end etc) so that we can capture the properties separately.

miike:
Adding @jethron for review as well.

schemas/com.snowplowanalytics.snowplow/media_player_ad_event/jsonschema/1-0-0
matus-tomlein: Right, I separated out the individual event schemas into another PR – #1283
There are quite a few of them, let me know if there is something I can do to make it easier to review!

schemas/com.snowplowanalytics.snowplow/media_player_ping_event/jsonschema/1-0-0
matus-tomlein: In PR #1283, I added a schema for the `percent_progress` events as well.

We store the current progress using the `currentTime` and `percentProgress` properties in the [`media_player` entity schema](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/media_player/jsonschema/1-0-0#L47).

Is that what you mean by units or did I misunderstand?

schemas/com.snowplowanalytics.snowplow/media_player_quality/jsonschema/1-0-0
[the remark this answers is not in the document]
matus-tomlein: In PR #1283, I added two event schemas for the two types of events:

* `media_player_event_quality_change` – Media player event tracked when the video playback quality changes automatically.
* `media_player_event_user_update_quality` – Media player event tracked when the video playback quality changes as a result of user interaction (choosing a different quality setting).

matus-tomlein:
Update: I created a new PR #1283 where I added schemas for media player events.

schemas/com.snowplowanalytics.snowplow/media_player_ad_break/jsonschema/1-0-0
```diff
@@ -0,0 +1,50 @@
...
+      "maxLength": 4096
+    },
+    "breakId": {
+      "type": "string",
+      "format": "uuid",
+      "description": "An identifier for the ad break."
+    },
+    "startTime": {
+      "type": "number",
+      "description": "Playback time at the start of the ad break.",
```
jethron, line 27: Can we specify a unit here? In context it's obviously seconds, but we also have `timeToStart` below that's milliseconds so probably good to be explicit.

schemas/com.snowplowanalytics.snowplow/media_player_session/jsonschema/1-0-0
```diff
@@ -0,0 +1,102 @@
...
+      "description": "Total seconds user spent playing content on mute (excluding ads).",
+      "minimum": 0,
+      "maximum": 2147483647
+    },
+    "contentDurationPlayed": {
+      "type": [
+        "number",
+        "null"
+      ],
+      "description": "Total seconds of the content played. Each part of the content played is counted once (i.e., counts rewinding or rewatching the same content only once). Playback rate does not affect this value.",
```
jethron, line 57: This is the only one that mentions playback rate, does that mean the other fields implicitly _are_ affected by playback rate?

schemas/com.snowplowanalytics.snowplow/media_player_ad_break/jsonschema/1-0-0
matus-tomlein: Good point, added units!

schemas/com.snowplowanalytics.snowplow/media_player_ad/jsonschema/1-0-0
```diff
@@ -0,0 +1,63 @@
...
+      "description": "Friendly name of the ad.",
+      "type": [
+        "null",
+        "string"
+      ],
+      "maxLength": 4096
+    },
+    "adId": {
+      "type": "string",
+      "format": "uuid",
```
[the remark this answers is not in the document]
matus-tomlein: That's a great point, UUID was too restrictive as the ID will be passed from users. Thinking about this, I also changed the `mediaSessionId` and `breakId` in the session and ad-break schemas, because those might also be passed by users and with different formats.

schemas/com.snowplowanalytics.snowplow/media_player_session/jsonschema/1-0-0
matus-tomlein: None of them are actually, but the other ones talk about the time user spent which I thought makes it more clear that playback rate can't affect it. On the other hand, "seconds of content played" sounds more unclear to me whether it can be affected by the playback rate. Happy to add the note to the other field descriptions if you think that would help.

matus-tomlein:
Update about what has changed in the last force push:

### `media_player_session` 

* rename timePlayedContent to timePlayed
* rename timePlayedContentMuted to timePlayedMuted
* rename contentDurationPlayed to contentWatched
* add timePaused
* rename totalAds to ads
* rename totalAdBreaks to adBreaks
* add adsClicked
* add adsSkipped
* add timeBuffering

### `media_player_quality`

Decided not to create this schema because the data quality information provided by the players is different in each one of them and it’s very likely that users will need to define a custom entity anyway to fit the information provided by their player and the way that the player provides the information. Happy to reconsider this if we figure out the common properties, but I don't think that I have (every time I look at a different player, I can see that other info needs to be tracked).

Additionally, I added a `timeBuffering` property to the session entity which calculates the amount of time that the player was buffering – this is a universal property that can be tracked anywhere and also the tracker can track it automatically (based on the buffer start and end events).

schemas/com.snowplowanalytics.snowplow/media_player_ad_break/jsonschema/1-0-0
```diff
@@ -0,0 +1,50 @@
...
+    "breakId": {
+      "type": "string",
+      "maxLength": 256,
+      "description": "An identifier for the ad break."
+    },
+    "startTime": {
+      "type": "number",
+      "description": "Playback time in seconds at the start of the ad break.",
+      "minimum": 0,
+      "maximum": 9007199254740991
```
miike, line 29: Why is this max so high?

schemas/com.snowplowanalytics.snowplow/media_player_session/jsonschema/1-0-0
```diff
@@ -0,0 +1,138 @@
...
+      "minimum": 0,
+      "maximum": 65535
+    },
+    "avgPlaybackRate": {
+      "type": [
+        "number",
+        "null"
+      ],
+      "description": "Average playback rate (1 is normal speed).",
+      "minimum": -9007199254740991,
```
miike, line 130: I'm not sure if we need a negative playback rate (technically that be playing the video backwards I think?)

schemas/com.snowplowanalytics.snowplow/media_player_ping_event/jsonschema/1-0-0
matus-tomlein: We addressed this in the PR for events.

schemas/com.snowplowanalytics.snowplow/media_player_ad_break/jsonschema/1-0-0
matus-tomlein: This was because it was based on the previous `media_player` schema which had these limits for playback time. With the new `player` schema, I have changed this now to 2147483647.

schemas/com.snowplowanalytics.snowplow/media_player_session/jsonschema/1-0-0
matus-tomlein: This was also based on the older `media_player` schema that had this range. I changed it to the range 0 to 16 now (I found that 16 is usually the max rate on browsers).

matus-tomlein:
**Update:**

1. changed the vendor for all media entities to `com.snowplowanalytics.snowplow.media` and renamed `media_player_session` to `session` and similar for the ad schemas.
2. added a new `player` schema that should be an updated version of the [`media_player` schema](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/media_player/jsonschema/1-0-0).
    * added properties to it: `label` (used to be in events before), `playerType` and `mediaType` (used in the media dbt model but haven't had a single place before), `fullscreen`, `pictureInPicture`
    * made some properties optional: `muted`, `volume`, `playbackRate`, `loop`
    * removed the `percentProgress` property which is now reported in the percent progress event and I don't think it's that useful because the schema has `currentTime` and `duration` properties anyway
    * changed max values for `currentTime` and `duration` to 2147483647 (was too large before)
    * changed min and max for `playbackRate` to 0 to 16
3. `ad_break` schema – decreased max value for `startTime`
4. `ad` schema
    * removed `percentProgress` which is now reported in `quartile`, `skip` and `click` events
    * decreased max value for `duration`

matus-tomlein:
Update: added a `quality` attribute to the `player` schema.

schemas/com.snowplowanalytics.snowplow.media/ad_break/jsonschema/1-0-0
```diff
@@ -0,0 +1,50 @@
...
+    },
+    "startTime": {
+      "type": "number",
+      "description": "Playback time in seconds at the start of the ad break.",
+      "minimum": 0,
+      "maximum": 2147483647
+    },
+    "breakType": {
+      "description": "Type of ads within the break: linear (take full control of the video for a period of time), nonlinear (run concurrently to the video), companion (accompany the video but placed outside the player).",
+      "enum": [
```
miike, line 33: @jethron Thoughts on removing the enum here? There may be other things we otherwise need to include (overlays - complete or partial, outstream (not sure if we should include these at all?).

schemas/com.snowplowanalytics.snowplow.media/session/jsonschema/1-0-0
```diff
@@ -0,0 +1,138 @@
...
+    "name": "session",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "mediaSessionId": {
+      "type": "string",
+      "maxLength": 256,
+      "description": "An identifier for the media session."
```
miike, line 15: We should describe if / when this is reset (or does that description belong in the tracker?)

schemas/com.snowplowanalytics.snowplow.media/session/jsonschema/1-0-0
```diff
@@ -0,0 +1,138 @@
...
+    },
+    "startedAt": {
+      "type": [
+        "null",
+        "string"
+      ],
+      "description": "Date-time timestamp of when the session started.",
```
miike, line 22: utc or local time?

schemas/com.snowplowanalytics.snowplow.media/session/jsonschema/1-0-0
```diff
@@ -0,0 +1,138 @@
...
+        "integer",
+        "null"
+      ],
+      "description": "Number of ad breaks played.",
+      "minimum": 0,
+      "maximum": 65535
+    },
+    "avgPlaybackRate": {
+      "type": [
+        "number",
```
miike, line 126: Do we need to / plan to round this to a certain precision in the trackers?

schemas/com.snowplowanalytics.snowplow.media/session/jsonschema/1-0-0
matus-tomlein: I have used local time in the trackers – clarified it here as well.

schemas/com.snowplowanalytics.snowplow.media/session/jsonschema/1-0-0
matus-tomlein: It is not automatically reset, but provided by the user for the media content played in the media player. I have clarified it here and also in the trackers.

schemas/com.snowplowanalytics.snowplow.media/ad_break/jsonschema/1-0-0
jethron: IMO overlays is a special case of companion and outstream is still linear; the context of the player's location in the page isn't captured in this property so I don't think should be factored in here. It _might_ be useful for users to populate this freeform, but probably only worth the flexibility if we don't intend specific handling of the value in the modeling stage.

schemas/com.snowplowanalytics.snowplow.media/ad_break/jsonschema/1-0-0
```diff
@@ -0,0 +1,50 @@
...
+        "linear",
+        "nonlinear",
+        "companion",
+        null
+      ],
+      "type": [
+        "string",
+        "null"
+      ]
```
jethron, line 42: Should there be like an optional `podSize` here to complement `podPosition` in `ad`?

schemas/com.snowplowanalytics.snowplow.media/ad/jsonschema/1-0-0
```diff
@@ -0,0 +1,63 @@
...
+      "maxLength": 4096,
+      "description": "The ID of the ad creative."
+    },
+    "podPosition": {
+      "type": [
+        "integer",
+        "null"
+      ],
+      "description": "The position of the ad within the ad break, starting with 1.",
+      "minimum": 0,
```
jethron, line 39: This was probably covered, did we intend `minimum: 1`/`exclusiveMinimum: true` here to match the description?

schemas/com.snowplowanalytics.snowplow.media/ad_break/jsonschema/1-0-0
```diff
@@ -0,0 +1,50 @@
...
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow.media",
+    "name": "ad_break",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "name": {
+      "description": "Ad break name (e.g., pre-roll, mid-roll, and post-roll).",
```
jethron, line 13: Nit: Trailing period in descriptions seems inconsistent between the various schemas.

schemas/com.snowplowanalytics.snowplow.media/player/jsonschema/1-0-0
```diff
@@ -0,0 +1,131 @@
...
+    "vendor": "com.snowplowanalytics.snowplow.media",
+    "name": "player",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "currentTime": {
+      "type": "number",
+      "description": "The current playback time",
```
jethron, line 14: `in seconds`? Should this mention "position" at all or is that more confusing (thinking `player.currentTime` sounds a lot like `sum(session.timePlayed)`)?

schemas/com.snowplowanalytics.snowplow.media/player/jsonschema/1-0-0
```diff
@@ -0,0 +1,131 @@
...
+      "minimum": 0,
+      "maximum": 2147483647
+    },
+    "duration": {
+      "type": [
+        "number",
+        "null"
+      ],
+      "description": "A double-precision floating-point value indicating the duration of the media in seconds",
```
jethron, line 23: `double-precision` seems like an unnecessary implementation detail.

schemas/com.snowplowanalytics.snowplow.media/player/jsonschema/1-0-0
```diff
@@ -0,0 +1,131 @@
...
+      "description": "Whether playback of the media has ended"
+    },
+    "fullscreen": {
+      "type": [
+        "boolean",
+        "null"
+      ],
+      "description": "Whether the video element is fullscreen"
+    },
+    "isLive": {
```
jethron, line 38: Maybe drop `is` to be consistent with the other flags? Maybe `livestream` or something?

schemas/com.snowplowanalytics.snowplow.media/ad_break/jsonschema/1-0-0
```diff
@@ -0,0 +1,50 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for a context entity, shared with all media_player_ad events belonging to the ad break.",
```
jethron, line 3: Not sure how to reference the new name unambiguously. :sweat_smile: `com.snowplowanalytics.snowplow.media/ad` is a bit more of a mouthful.

schemas/com.snowplowanalytics.snowplow.media/ad_break/jsonschema/1-0-0
matus-tomlein: I think that is useful! Have added the `podSize` property to ad break schema.

schemas/com.snowplowanalytics.snowplow.media/ad/jsonschema/1-0-0
matus-tomlein: I do remember changing this already based on some comment, must have been in a different place... Have changed to minimum 1.

schemas/com.snowplowanalytics.snowplow.media/ad_break/jsonschema/1-0-0
matus-tomlein: Unified on adding the period to all descriptions.

schemas/com.snowplowanalytics.snowplow.media/player/jsonschema/1-0-0
matus-tomlein: Good point, the property is based on the [HTML5 player one](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/currentTime) but I agree that is a bit confusing considering the stats in session. Have changed the description to clarify it more and mention position.

schemas/com.snowplowanalytics.snowplow.media/player/jsonschema/1-0-0
matus-tomlein: Removed the double-precision part.

schemas/com.snowplowanalytics.snowplow.media/player/jsonschema/1-0-0
matus-tomlein: True, have changed to `livestream` to be consistent with other boolean properties.

schemas/com.snowplowanalytics.snowplow.media/ad_break/jsonschema/1-0-0
matus-tomlein: The `media_player_ad` was a name of what I originally planned to be a single schema for all ad events but we have diverged from that so just changed it to "ad events".

jethron (approved):
LGTM!

matus-tomlein:
Integrated into R142 PR #1302

PR #1297, opened 2023-04-28
Add com.snowplowanalytics.snowplow.ecommerce/transaction_error/jsonschema/1-0-0

### LINKED ISSUE #1296: Add com.snowplowanalytics.snowplow.ecommerce/transaction_error/jsonschema/1-0-0
Add a transaction error schema to track transaction failures in first or third party systems.

igneel64:
Introduce the `transaction_error` schema. This schema will be sent together with a `trns_error` ecommerce action and will have a `transaction` entity attached to it as well.
(close #1296)

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow.ecommerce/transaction_error/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow.ecommerce/transaction_error/jsonschema/1-0-0
@@ -1,0 +1,58 @@
+{
+    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "description": "Schema for a transaction error or rejection entity in ecommerce.",
+    "self": {
+        "vendor": "com.snowplowanalytics.snowplow.ecommerce",
+        "name": "transaction_error",
+        "format": "jsonschema",
+        "version": "1-0-0"
+    },
+    "type": "object",
+    "properties": {
+        "error_code": {
+            "type": [
+                "string",
+                "null"
+            ],
+            "description": "Error-identifying code for the transaction issue. E.g. E522",
+            "maxLength": 256
+        },
+        "error_shortcode" : {
+            "type": [
+                "string",
+                "null"
+            ],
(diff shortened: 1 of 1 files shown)


schemas/com.snowplowanalytics.snowplow.ecommerce/transaction_error/jsonschema/1-0-0
```diff
@@ -0,0 +1,58 @@
...
+    "properties": {
+        "error_code": {
+            "type": [
+                "string",
+                "null"
+            ],
+            "description": "Error-identifying code for the transaction issue. E.g. E522",
+            "maxLength": 256
+        },
+        "error_shortcode" : {
```
matus-tomlein, line 20: Off-topic: it would be really useful to have some style guide about how we define our schemas (or do we already have that?). Currently we follow different guidelines based on the use case of the schemas, for instance the media and mobile schemas name properties with camel case, not snake case. It's better to follow the local conventions of other schemas within the use case to make the experience consistent, but it'd be useful to know how to go about completely new schemas...

matus-tomlein (approved):
LGTM!

schemas/com.snowplowanalytics.snowplow.ecommerce/transaction_error/jsonschema/1-0-0
igneel64: Yeap that is true, we don't even have a formatting guide for the schemas afaik.
For the property casing/naming I would leave it up to the use-case, I don't see any particular issue TBH _(considering consistency not being an issue between different groups of schemas)_. e.g. the `snowplow.ecommerce` schemas are all snake_cased.

igneel64: Thanks Matus!

rlh1994: I think it's handled somewhere in the pipeline anyway, in that all warehouses except snowflake have snake_case fields but from tracking it would be good to be consistent I guess. I though we went with camelCase but that might just be the ones I used

rlh1994 (approved):

igneel64:
Will try to add into #1258

PR #1300, opened 2023-05-19
Update com.sendgrid schemas to v2.0.1

johannac:
This PR updates the com.sendgrid schemas to conform to the latest data types being sent by the [sendgrid webhook](https://docs.snowplow.io/docs/collecting-data/collecting-data-from-third-parties/sendgrid/).

We've recently started collecting events from the sendgrid webhook with our snowplow pipeline and are experiencing a large amount of schema violations using v2.0.0 of the schemas. The majority of events we're collecting are failing validation. After inspecting the errors I've made updates to the schemas including:

 - removing the maximum length setting on the `asm_group_id` property: this property is an integer and the maximum value was set to 32767. I can't think of any reason why this would be. Our ASM group ID in sendgrid is larger than this number as I'm sure are many other user's.
 - remove ipv4 format on `ip` property: unfortunately sendgrid will send the string `unknown` in this field when there is no IP address, so requiring ipv4 format isn't appropriate for this field.
 - remove email format on `email` property: there are a few events which will contain invalid email addresses in this field - for example a bounced event, so in those cases we're actually expecting an invalid email address and the email format validator isn't appropriate. 

Since these are all new files it's hard to see in the diff exactly what changes so I compiled a list of what I changed in each schema:

- bounce
  - remove maximum length on asm_group_id
  - remove ipv4 format on IP address field
  - remove email format on email property
- click
  - remove maximum length on asm_group_id
  - remove ipv4 format on IP address field
- deferred
  - remove maximum length on asm_group_id
  - remove ipv4 format on IP address field
  - remove email format on email property
- delivered
  - remove maximum length on asm_group_id 
- dropped
  - remove email format on email property
- group_resubscribe
  - remove maximum length on asm_group_id
  - remove ipv4 format on IP address field
- group_unsubscribe
  - remove maximum length on asm_group_id
  - remove ipv4 format on IP address field
- open
  - remove maximum length on asm_group_id, integer
  - remove ipv4 format on IP address field
- processed
  - remove maximum length on asm_group_id
- spamreport
  - remove maximum length on asm_group_id 
- unsubscribe
  - remove maximum length on asm_group_id

### DIFF
--- a/schemas/com.sendgrid/bounce/jsonschema/2-0-1
+++ b/schemas/com.sendgrid/bounce/jsonschema/2-0-1
@@ -1,0 +1,105 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for a SendGrid bounce event. Property descriptions derived from the SendGrid documentation: https://sendgrid.com/docs/for-developers/tracking-events/event/",
+  "self": {
+    "vendor": "com.sendgrid",
+    "name": "bounce",
+    "version": "2-0-1",
+    "format": "jsonschema"
+  },
+  "type": "object",
+  "properties": {
+    "ip": {
+      "description": "The IP address used to send the email. For open and click events, it is the IP address of the recipient who engaged with the email.",
+      "type": "string"
+    },
+    "timestamp": {
+      "description": "The timestamp of when the message was sent",
+      "type": "string",
+      "format": "date-time"
+    },
+    "email": {
+      "description": "The email address of the recipient",
+      "type": "string"
+    },
+    "newsletter": {
+      "description": "Legacy Marketing Email tool fields",
+      "type": "object",
+      "properties": {
+        "newsletter_user_list_id": {
+          "type": "string"
+        },
+        "newsletter_id": {
+          "type": "string"
+        },
+        "newsletter_send_id": {
+          "type": "string"
+        }
+      },
+      "additionalProperties": true
+    },
+    "sg_event_id": {
+      "description": "A unique ID to this event that you can use for deduplication purposes. These IDs are up to 100 characters long and are URL safe.",
+      "type": "string",
+      "minLength": 22,
+      "maxLength": 4096
+    },
+    "smtp-id": {
+      "description": "A unique ID attached to the message by the originating system",
+      "type": "string"
+    },
+    "reason": {
+      "description": "Any sort of error response returned by the receiving server that describes the reason this event type was triggered",
+      "type": "string"
+    },
+    "tls": {
+      "description": "Indicates whether TLS encription was used in sending this message. For more information about TLS, see the SendGrid TLS Glossary page.",
+      "type": ["string", "integer"]
+    },
+    "status": {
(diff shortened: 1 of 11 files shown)


schemas/com.sendgrid/bounce/jsonschema/2-0-1
```diff
@@ -0,0 +1,105 @@
...
+    "category": {
+      "description": "Categories are custom tags that you set for the purpose of organizing your emails. Categories can be set as an array or string, and they will be returned as such when posted in your event endpoint.",
+      "type": ["array", "string"],
+      "items": {
+        "type": "string"
+      }
+    },
+    "asm_group_id": {
+      "description": "The ID of the unsubscribe group the recipient’s email address is included in. ASM IDs correspond to the ID that is returned when you create an unsubscribe group.",
+      "type": "integer",
```
matus-tomlein, line 76: I think removing or changing the maximum might make this a breaking change that requires us to make it a `3-0-0` version instead of `2-0-1`.

Also, we could consider keeping maximum, just making it higher – perhaps 2147483647 make it a 32-bit integer or even 9223372036854775807 if we think we need it.

schemas/com.sendgrid/bounce/jsonschema/2-0-1
```diff
@@ -0,0 +1,105 @@
...
+      "description": "The IP address used to send the email. For open and click events, it is the IP address of the recipient who engaged with the email.",
+      "type": "string"
+    },
+    "timestamp": {
+      "description": "The timestamp of when the message was sent",
+      "type": "string",
+      "format": "date-time"
+    },
+    "email": {
+      "description": "The email address of the recipient",
```
matus-tomlein, line 22: Having removed the `email` format, could we add a regular expression to make sure that the emails follow some expected format?

```suggestion
      "pattern": "^\\S+@\\S+\\.\\S+$",
      "description": "The email address of the recipient",
```

matus-tomlein: Or is Sendgrid even breaking that expectation and sends invalid email adresses?

schemas/com.sendgrid/click/jsonschema/2-0-1
```diff
@@ -0,0 +1,95 @@
...
+    },
+    "timestamp": {
+      "description": "The timestamp of when the message was sent",
+      "type": "string",
+      "format": "date-time"
+    },
+    "email": {
+      "description": "The email address of the recipient",
+      "type": "string",
+      "format": "email"
```
matus-tomlein, line 24: This still has the `email` format, is this intentional? I think it'd be better to be consistent about the validation in all sendgrid schemas.

matus-tomlein (commented):
Thank you for putting this together!

I had just three types of comments that show up in multiple schemas:

* Can we use a regular expression to validate the email property or does Sendgrid break even that expectation?
* Can we change the validation for the email property in all of the Sendgrid schemas to make it consistent?
* By changing the maximum for the `asm_group_id` property, I think we'll need to make this a version 3-0-0 instead of 2-0-1 as that change leads to a different type of the property.

matus-tomlein:
Thanks again for the PR! We have taken the changes with some modifications into PR #1310 and will work towards releasing them.

Issue #1301, opened 2023-05-29
Add com.snowplowanalytics.snowplow.media/player/jsonschema/1-0-0


matus-tomlein:
Add a new version of the `iglu:/com.snowplowanalytics.snowplow/media_player/jsonschema/1-0-0` schema with the new media vendor `com.snowplowanalytics.snowplow.media` and new properties for `fullscreen`, `pictureInPicture` and more.

PR #1283, opened 2023-04-03
Add media player event schemas

matus-tomlein:
This PR separates out the different event types that we previously represented in the [media_player_event](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/media_player_event/jsonschema/1-0-0) into separate event schemas. The reasons for separating out the schemas are:

1. Make it more efficient to filter events in modelling using their schemas instead of having to dive into the `type` property.
2. Enable real-time use cases where one would subscribe to the particular event schemas that they are interested in.
3. Improve validation (previously the `type` property was a string without any validation).

Each of the schemas contain the `label` property as the `media_player_event`:

Request Key | Required | Type/Format | Description
-- | -- | -- | --
label | N | string | A custom identifier for the media player

In addition to the existing event types, there are several new ones for ads and data quality tracking.

### Event schemas for controlling the playback

Event schema | Description
-- | --
media_player_event_ready | Media player event fired when the media tracking is successfully attached to the player and can track events.
media_player_event_play | Media player event sent when the player changes state to playing from previously being paused.
media_player_event_pause | Media player event sent when the user pauses the playback.
media_player_event_end | Media player event sent when playback stops when end of the media is reached or because no further data is available.
media_player_event_seek_start | Media player event sent when a seek operation begins.
media_player_event_seek_end | Media player event sent when a seek operation completes.

### Event schemas for changes in playback settings

Event schema | Description
-- | --
media_player_event_playbackrate_change | Media player event sent when the playback rate has changed.
media_player_event_volume_change | Media player event sent when the volume has changed.
media_player_event_fullscreen_change | Media player event fired immediately after the browser switches into or out of full-screen mode.
media_player_event_pictureinpicture_change | Media player event fired immediately after the browser switches into or out of picture-in-picture mode.

### Event schemas for tracking playback progress

Event schema | Description
-- | --
media_player_event_ping | Media player event fired periodicaly during main content playback, regardless of other API events that have been sent.
media_player_event_percent_progress | Media player event fired when a percentage boundary set in options.boundaries is reached

### Event schemas for ad events

Event schema | Description
-- | --
media_player_event_ad_break_start | Media player event that signals the start of an ad break.
media_player_event_ad_break_end | Media player event that signals the end of an ad break.
media_player_event_ad_start | Media player event that signals the start of an ad.
media_player_event_ad_percent_progress | Media player event fired when a percentage boundary is reached after continuous ad playback at normal speed.
[...]

Issue #1276, opened 2023-03-29
Add com.snowplowanalytics.snowplow.media/xxx_event/jsonschema/1-0-0


matus-tomlein:
Add schemas for media events:

com.snowplowanalytics.snowplow.media/volume_change_event/jsonschema/1-0-0
com.snowplowanalytics.snowplow.media/seek_start_event/jsonschema/1-0-0
com.snowplowanalytics.snowplow.media/seek_end_event/jsonschema/1-0-0
com.snowplowanalytics.snowplow.media/ready_event/jsonschema/1-0-0
com.snowplowanalytics.snowplow.media/quality_change_event/jsonschema/1-0-0
com.snowplowanalytics.snowplow.media/playback_rate_change_event/jsonschema/1-0-0
com.snowplowanalytics.snowplow.media/play_event/jsonschema/1-0-0
com.snowplowanalytics.snowplow.media/picture_in_picture_change_event/jsonschema/1-0-0
com.snowplowanalytics.snowplow.media/percent_progress_event/jsonschema/1-0-0
com.snowplowanalytics.snowplow.media/pause_event/jsonschema/1-0-0
com.snowplowanalytics.snowplow.media/fullscreen_change_event/jsonschema/1-0-0
com.snowplowanalytics.snowplow.media/error_event/jsonschema/1-0-0
com.snowplowanalytics.snowplow.media/end_event/jsonschema/1-0-0
com.snowplowanalytics.snowplow.media/buffer_start_event/jsonschema/1-0-0
com.snowplowanalytics.snowplow.media/buffer_end_event/jsonschema/1-0-0
com.snowplowanalytics.snowplow.media/ad_start_event/jsonschema/1-0-0
com.snowplowanalytics.snowplow.media/ad_skip_event/jsonschema/1-0-0
com.snowplowanalytics.snowplow.media/ad_resume_event/jsonschema/1-0-0
com.snowplowanalytics.snowplow.media/ad_quartile_event/jsonschema/1-0-0
com.snowplowanalytics.snowplow.media/ad_pause_event/jsonschema/1-0-0
com.snowplowanalytics.snowplow.media/ad_complete_event/jsonschema/1-0-0
com.snowplowanalytics.snowplow.media/ad_click_event/jsonschema/1-0-0
com.snowplowanalytics.snowplow.media/ad_break_start_event/jsonschema/1-0-0
com.snowplowanalytics.snowplow.media/ad_break_end_event/jsonschema/1-0-0
[...]

PR #1302, opened 2023-05-29
Release R142

matus-tomlein:
The scope of this release is limited to new schemas for a new media tracking plugin for Web and mobile trackers.
It contains schemas for both new events as well as new entities.

They all have the same vendor: `com.snowplowanalytics.snowplow.media`
They all have the same version: 1-0-0

### Relation to the previous media schemas

Previously, there was only one media event schema: [`iglu:com.snowplowanalytics.snowplow/media_player_event/jsonschema/1-0-0`](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/media_player_event/jsonschema/1-0-0)
It contained a `type` property which denoted the type of the media event.
Instead, we will now have separate schemas for each event type.
The reason is that we want to make it easier to subscribe to certain event types in real-time scenarios, improve validation and add various properties to individual event types.

The schemas for context entities are all new except for the `player` schema.
The player schema previously had a different vendor and name: [`iglu:/com.snowplowanalytics.snowplow/media_player/jsonschema/1-0-0`](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/media_player/jsonschema/1-0-0)
The new player schema adds new properties such as `fullscreen` or `pictureInPicture`.

### Media event schemas

**Issue** (I created one issue for all of the event schemas, but happy to change that if you think it's better to have separate issues): #1276
**PR**: #1283

Schema names:

* `volume_change_event`
* `seek_start_event`
* `seek_end_event`
* `ready_event`
* `quality_change_event`
* `playback_rate_change_event`
* `play_event`
* `picture_in_picture_change_event`
* `percent_progress_event`
* `pause_event`
* `fullscreen_change_event`
* `error_event`
* `end_event`
* `buffer_start_event`
* `buffer_end_event`
* `ad_start_event`
* `ad_skip_event`
* `ad_resume_event`
* `ad_quartile_event`
* `ad_pause_event`
* `ad_complete_event`
* `ad_click_event`
* `ad_break_start_event`
* `ad_break_end_event`
* `ping_event`

### Media entity schemas

**PR**: #1270.

Schema name | Issue
--|--
`player` | #1301
`ad_break` | #1268
`ad` | #1267
`session` | #1266


schemas/com.snowplowanalytics.snowplow.media/playback_rate_change_event/jsonschema/1-0-0
```diff
@@ -0,0 +1,29 @@
...
+      "type": [
+        "number",
+        "null"
+      ],
+      "description": "Playback rate before the change (1 is normal)",
+      "minimum": 0,
+      "maximum": 16
+    },
+    "newRate": {
+      "type": "number",
```
istreeter, line 22: Should `newRate` be a required field?  The schema is fine -- I'm only asking because I noticed you don't allow null for this field.

schemas/com.snowplowanalytics.snowplow.media/volume_change_event/jsonschema/1-0-0
```diff
@@ -0,0 +1,29 @@
...
+      "type": [
+        "integer",
+        "null"
+      ],
+      "description": "Volume percentage before the change.",
+      "minimum": 0,
+      "maximum": 100
+    },
+    "newVolume": {
+      "type": "integer",
```
istreeter, line 22: Same for `newVolume` -- I noticed you don't allow null, so I wonder if you also intended to make it required?

schemas/com.snowplowanalytics.snowplow.media/playback_rate_change_event/jsonschema/1-0-0
matus-tomlein: Yes, it should be a required field, good catch! I have made it a required field now.

istreeter (approved):

miike (approved):

jbeemster (approved):

PR #1304, opened 2023-05-30
Add com.snowplowanalytics.snowplow/web_vitals/jsonschema/1-0-0 

### LINKED ISSUE #1303: Add com.snowplowanalytics.snowplow/web_vitals/jsonschema/1-0-0
Add the first version of the schema for the [Web Vitals](https://web.dev/vitals/) metrics

igneel64:
Add the Web Vitals schema to be paired with the Snowplow Web Vitals JavaScript plugin

close #1303

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/web_vitals/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow/web_vitals/jsonschema/1-0-0
@@ -1,0 +1,61 @@
+{
+    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "description": "Schema for a web vitals tracking event. For more information on web vitals you can visit https://web.dev/vitals/.",
+    "self": {
+        "vendor": "com.snowplowanalytics.snowplow",
+        "name": "web_vitals",
+        "format": "jsonschema",
+        "version": "1-0-0"
+    },
+    "type": "object",
+    "properties": {
+        "cls": {
+            "title": "Cumulative Layout Shift",
+            "type": [ "number", "null" ],
+            "maximum": 2147483647,
+            "minimum": 0,
+            "description": "A unitless metric for measuring visual stability because it helps quantify how often users experience unexpected layout shifts. For more information https://web.dev/cls/."
+        },
+        "fid": {
+            "title": "First Input Delay",
+            "type": [ "number", "null" ],
+            "maximum": 2147483647,
+            "minimum": 0,
+            "description": "A metric for measuring load responsiveness because it quantifies the experience users feel when trying to interact with unresponsive pages. Measured in milliseconds. For more information https://web.dev/fid/."
+        },
+        "lcp": {
+            "title": "Largest Contentful Paint",
+            "type": [ "number", "null" ],
+            "maximum": 2147483647,
+            "minimum": 0,
+            "description": "A metric for measuring perceived load speed because it marks the point in the page load timeline when the page's main content has likely loaded. Measured in milliseconds. For more information https://web.dev/lcp/."
+        },
+        "fcp": {
+            "title": "First Contentful Paint",
+            "type": ["number", "null"],
+            "maximum": 2147483647,
+            "minimum": 0,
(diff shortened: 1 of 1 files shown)


jethron (approved):
LGTM

schemas/com.snowplowanalytics.snowplow/web_vitals/jsonschema/1-0-0
```diff
@@ -0,0 +1,55 @@
...
+        "fid": {
+            "type": [ "number", "null" ],
+            "maximum": 2147483647,
+            "minimum": -2147483647,
+            "description": "A metric for measuring load responsiveness because it quantifies the experience users feel when trying to interact with unresponsive pages. Measured in milliseconds. For more information https://web.dev/fid/."
+        },
+        "lcp": {
+            "type": [ "number", "null" ],
+            "maximum": 2147483647,
+            "minimum": -2147483647,
```
matus-tomlein, line 27: Should we keep the minimum for these properties below 0? As I understand they measure the speed of the page load, so negative numbers wouldn't make sense?

schemas/com.snowplowanalytics.snowplow/web_vitals/jsonschema/1-0-0
```diff
@@ -0,0 +1,55 @@
...
+            "description": "A metric that assesses responsiveness. INP observes the latency of all interactions a user has made with the page, and reports a single value which all (or nearly all) interactions were below that value. For more information https://web.dev/inp/."
+        },
+        "ttfb": {
+            "type": [ "number", "null" ],
+            "maximum": 2147483647,
+            "minimum": -2147483647,
+            "description": "A DOMHighResTimeStamp referring to the time in milliseconds between the browser requesting a page and when it receives the first byte of information from the server. For more information https://web.dev/ttfb/."
+        },
+        "navigationType": {
```
matus-tomlein, line 48: Nit: Is it intentional that there is camel-case here? I think we informally agreed on the underscore case for new schemas if there is no previous context where other case is used. (btw, this is a tiny nitpick)

matus-tomlein (approved):
Agreed with Jethro on expanding the abbreviations and it seems to me that below 0 values should be rejected, but otherwise LGTM!

schemas/com.snowplowanalytics.snowplow/web_vitals/jsonschema/1-0-0
igneel64: There are some cases where properties like this might get reported as less than 0. This is to keep it on the safe side.

schemas/com.snowplowanalytics.snowplow/web_vitals/jsonschema/1-0-0
```diff
@@ -0,0 +1,55 @@
...
+        "format": "jsonschema",
+        "version": "1-0-0"
+    },
+    "type": "object",
+    "properties": {
+        "cls": {
+            "type": [ "number", "null" ],
+            "maximum": 2147483647,
+            "minimum": -2147483647,
+            "description": "A unitless metric for measuring visual stability because it helps quantify how often users experience unexpected layout shifts. For more information https://web.dev/cls/."
```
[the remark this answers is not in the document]
igneel64: Done! First schema having titles 😂

schemas/com.snowplowanalytics.snowplow/web_vitals/jsonschema/1-0-0
```diff
@@ -0,0 +1,55 @@
+{
+    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "description": "Schema for a web vitals tracking event. For more information on web vitals you can visit https://web.dev/vitals/.",
+    "self": {
+        "vendor": "com.snowplowanalytics.snowplow",
```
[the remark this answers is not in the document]
igneel64: I think we should own it here, in the worst case scenario we just change the description.

schemas/com.snowplowanalytics.snowplow/web_vitals/jsonschema/1-0-0
igneel64: Thinking about this. SInce this is a specific event and will not mess with any other events being sent from the browers, I added a minimum of 0. If some values come as negative it is a browser bug and should not be included in the calculations.

igneel64:
Added in #1305

PR #1313, opened 2023-07-10
Release R145

matus-tomlein:
~~This release adds 3-0-0 versions of the Sengrid schemas, originally reviewed in PR #1310.~~

This release adds the Privacy Sandbox Topics schema, reviewed at https://github.com/snowplow/iglu-central/pull/1319


miike:
These schemas are used quite frequently so it would be good to fuzz these against some real data. @johannac is this something you'd be able to help out with?

johannac:
@miike Sure, happy to help test, if we can. We are a Snowplow BDP customer, so we may need to coordinate this through the support team. I can put the v3 schemas into our iglu, but I believe someone from support would need to upgrade the SendGrid enrichment adapter to use the v3-0-0 schemas. Alternatively if you'd just like a big chunk of test data, we could pull some from our production pipeline which has a good volume of SendGrid events coming in daily. Again support could probably coordinate that for us.

miike:
Thanks - let me link up with the support team and get them to coordinate so we can get this tested!

igneel64:
All previous commits moved to #1320 since we are blocked on sendgrid testing.

igneel64 (approved):

matus-tomlein:
Sorry for adding new commits to this release, the reason is because of an issue that we discovered in the recently published media schemas that could potentially cause pipeline failures.

We currently have 2 schemas on Iglu Central that may load into the same column in raw events (`contexts_com_snowplowanalytics_snowplow_media_player_1`) or the same table on some warehouses:

* This one has been published for a long time: `iglu:com.snowplowanalytics.snowplow/media_player/jsonschema/1-0-0`
* This one we have published relatively recently: `iglu:com.snowplowanalytics.snowplow.media/player/jsonschema/1-0-0`

The second one adds new properties as well as makes some other properties optional. We need to remove the recently published v1 schema and publish it as v2 instead in order to make sure that loading events doesn't break in case events are tracked for both of the above schemas.

istreeter (approved):

miike (approved):

jbeemster (approved):

PR #1327, opened 2023-08-22
Add com.snowplowanalytics.snowplow/cross_navigation/jsonschema/1-0-0

### LINKED ISSUE #1326: Add com.snowplowanalytics.snowplow/cross_navigation/jsonschema/1-0-0

igneel64:
Add schema for cross-domain, cross-platform, app-web/web-app navigations. This schema is going to be used in both web and mobile implementations. Also expecting enrich changes.

(close #1326)

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/cross_navigation/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow/cross_navigation/jsonschema/1-0-0
@@ -1,0 +1,68 @@
+{
+    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "description": "Schema for a user and session identifiers entity included in cross platform or cross domain navigation.",
+    "self": {
+        "vendor": "com.snowplowanalytics.snowplow",
+        "name": "cross_navigation",
+        "format": "jsonschema",
+        "version": "1-0-0"
+    },
+    "type": "object",
+    "properties": {
+        "user_id": {
+            "description": "The business user ID, the user was identified in the source app.",
+            "type": [
+                "null",
+                "string"
+            ],
+            "maxLength": 4096
+        },
+        "domain_user_id": {
+            "type": "string",
+            "maxLength": 128,
+            "description": "Domain user ID of the source app from which a link was followed. The corresponding atomic property is named refr_domain_userid."
+        },
+        "session_id": {
+            "type": [
+                "null",
+                "string"
+            ],
+            "format": "uuid",
+            "description": "Session ID of the source app from which a link was followed."
+        },
+        "timestamp": {
+            "description": "Timestamp of the link click in the source app which triggered this navigation. The corresponding atomic property is named refr_dvce_tstamp.",
+            "type": "string",
+            "format": "date-time"
+        },
+        "source_id": {
+            "type": [
+                "null",
+                "string"
+            ],
+            "maxLength": 4096,
+            "description": "ID of the source app where the link that started this navigation originated from."
+        },
+        "source_platform": {
+            "type": [
+                "null",
+                "string"
+            ],
+            "enum": ["web", "mob", "pc", "srv", "app" , "tv", "cnsl" ,"iot", null],
+            "description": "Platform of the source app (e.g., mobile, web, tablet app)."
+        },
+        "reason": {
+            "description": "Cross navigation linking information/identifier.",
+            "type": [
+                "null",
+                "string"
+            ],
+            "maxLength": 4096
+        }
+    },
+    "required": [
+        "domain_user_id",
+        "timestamp"
+    ],
+    "additionalProperties": false
+}


schemas/com.snowplowanalytics.snowplow/cross_navigation/jsonschema/1-0-0
```diff
@@ -0,0 +1,70 @@
...
+                "string"
+            ],
+            "maxLength": 1024,
+            "description": "ID of the source app where the link that started this navigation originated from."
+        },
+        "source_platform": {
+            "type": [
+                "null",
+                "string"
+            ],            "enum": ["web", "mob", "pc", "srv", "app" , "tv", "cnsl" ,"iot", null],
```
matus-tomlein, line 53: Nit: this needs a line break.

matus-tomlein (approved):
LGTM!

schemas/com.snowplowanalytics.snowplow/cross_navigation/jsonschema/1-0-0
igneel64: Stupid VSCode auto format-spacing!

schemas/com.snowplowanalytics.snowplow/cross_navigation/jsonschema/1-0-0
```diff
@@ -0,0 +1,71 @@
...
+            "type": [
+                "null",
+                "string"
+            ],
+            "maxLength": 4096
+        }
+    },
+    "required": [
+        "domain_user_id",
+        "timestamp"
```
adatzer, line 68: The `timestamp` can be of type `null` so i thought it would be optional. Is there a reason why it is required?

schemas/com.snowplowanalytics.snowplow/cross_navigation/jsonschema/1-0-0
```diff
@@ -0,0 +1,71 @@
...
+        },
+        "session_id": {
+            "type": [
+                "null",
+                "string"
+            ],
+            "format": "uuid",
+            "description": "Session ID of the source app from which a link was followed."
+        },
+        "timestamp": {
```
adatzer, line 33: The corresponding atomic property is named `refr_dvce_tstamp`. We don't have to use the same name, but wondering whether being consistent has some advantage. In addition, for this case in particular the atomic name seems a bit more descriptive than just `timestamp`.

schemas/com.snowplowanalytics.snowplow/cross_navigation/jsonschema/1-0-0
```diff
@@ -0,0 +1,71 @@
...
+    "properties": {
+        "user_id": {
+            "description": "The business user ID, the user was identified in the source app.",
+            "type": [
+                "null",
+                "string"
+            ],
+            "maxLength": 4096
+        },
+        "domain_user_id": {
```
adatzer, line 20: Here as well, the corresponding atomic property is named `refr_domain_userid` in case we'd want to use that.

schemas/com.snowplowanalytics.snowplow/cross_navigation/jsonschema/1-0-0
```diff
@@ -0,0 +1,71 @@
...
+            "type": "string",
+            "maxLength": 36,
```
adatzer, line 22: The maxLength needs to be at least 128 ([as is the `refr_domain_userid`](https://github.com/snowplow/iglu-central/blob/0d08c92ebde14b77c2f191b64622ac088c70f230/schemas/com.snowplowanalytics.snowplow/atomic/jsonschema/1-0-0#L449)).

schemas/com.snowplowanalytics.snowplow/cross_navigation/jsonschema/1-0-0
```diff
@@ -0,0 +1,71 @@
...
+                "string"
+            ],
+            "format": "date-time"
+        },
+        "source_id": {
+            "type": [
+                "null",
+                "string"
+            ],
+            "maxLength": 1024,
```
adatzer, line 46: why not 4096 ? :rocket:

schemas/com.snowplowanalytics.snowplow/cross_navigation/jsonschema/1-0-0
```diff
@@ -0,0 +1,71 @@
...
+            "description": "Cross navigation linking information/identifier.",
+            "type": [
+                "null",
+                "string"
+            ],
+            "maxLength": 4096
+        }
+    },
+    "required": [
+        "domain_user_id",
```
adatzer, line 67: Do we really need `domain_user_id` as required? For example, what if only `user_id` is available? Should that be considered a reason to invalidate the event?

schemas/com.snowplowanalytics.snowplow/cross_navigation/jsonschema/1-0-0
matus-tomlein: Have missed this before, but we should not have any nullable required properties (we have a few schemas like this and it's causing us problems). If we go based on the current tracking implementation, I think the `timestamp` should be required but should not be nullable – it is a required property for cross-domain tracking along with the domain user id.

But I am also happy if we decide to remove the required properties and make everything nullable and optional in the schema. I can imagine that in the future we might add an option to track the session ID in cross navigation without tracking the domain user ID. Not sure under which case we would not track the timestamp, but can't rule it out completely.

matus-tomlein: Decision: make it not nullable and required

adatzer (approved):
Looks good to me!

spenes (approved):
LGTM!

matus-tomlein:
Cherry picked to R152

Issue #1330, opened 2023-09-08
Change integer to number for deviceMemory in com.snowplowanalytics.snowplow/browser_context/jsonschema/1-0-0


dumkydewilde:
Some devices have less than 1GB of memory and will show e.g. 0.5 or 0.25 as [defined here](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/deviceMemory). These hits are determined invalid by:
```   
"deviceMemory": {
      "type": ["integer", "null"],
      "minimum": 0,
      "maximum": 1000,
      "description": "Approximate amount of device memory in gigabytes."
   }
```

Should be a non-breaking fix to change to
```   
"deviceMemory": {
      "type": ["number", "null"],
      "minimum": 0,
      "maximum": 1000,
      "description": "Approximate amount of device memory in gigabytes."
   }
```

PR #1331, opened 2023-09-08
1330 fix devicememory to comply with specs

dumkydewilde:
As per #1330, deviceMemory should be a number, not an int.


dumkydewilde:
Sorry, I see I messed up my commit history here a bit (☕️) with a change that I needed locally previously that should not be in this PR. The actual diff is correct in only showing the change from int to number.

PR #1342, opened 2023-09-15
Add updated schemas for the Mandrill webhook

matus-tomlein:
These new schemas account for the changes in the Mandrill webhook that caused failed events with the previous schema versions. All the changes are non-breaking. The following changes were made:

* set additional properties to true in the root object, in `msg` object, `location` and `metadata` objects
* allow null in `user_agent_parsed` (non-required) properties
* add `country_long` property to location object
* add descriptions for properties based on the Mandrill webhook docs

There is also one new schema for an event that we didn't capture before – `message_delivered`.

### Testing

The schemas were tested using ~1500 Mandrill events that previously failed validation with the previous schemas. The events were extracted from failed events storage in S3.

### DIFF
--- a/schemas/com.mandrill/message_clicked/jsonschema/1-0-2
+++ b/schemas/com.mandrill/message_clicked/jsonschema/1-0-2
@@ -1,0 +1,290 @@
+{
+	"$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+	"description": "Schema for a Mandrill message clicked event",
+	"self": {
+		"vendor": "com.mandrill",
+		"name": "message_clicked",
+		"format": "jsonschema",
+		"version": "1-0-2"
+	},
+
+	"type": "object",
+	"properties": {
+		"_id": {
+			"type": "string",
+			"description": "The unique identifier of the message that generated the event. This is not an event identifier, but rather a reference to the message ID for the email that was sent, opened, clicked, etc."
+		},
+		"ip": {
+			"type": "string",
+			"description": "IP address where the event originated"
+		},
+		"location": {
+			"type": ["object", "null"],
+			"description": "The geolocation where the event occurred. Value will be null if no location can be determined from the IP address of the event.",
+			"properties": {
+				"city": {
+					"type": "string"
+				},
+				"country_short": {
+					"type": "string",
+					"description": "Abbreviated country"
+				},
+				"country": {
+					"type": "string"
+				},
+				"country_long": {
+					"type": ["string", "null"],
+					"description": "Full country name"
+				},
+				"latitude": {
+					"type": "number"
+				},
+				"longitude": {
+					"type": "number"
+				},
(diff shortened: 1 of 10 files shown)


schemas/com.mandrill/message_clicked/jsonschema/1-0-2
```diff
@@ -0,0 +1,290 @@
...
+		},
+		"location": {
+			"type": ["object", "null"],
+			"description": "The geolocation where the event occurred. Value will be null if no location can be determined from the IP address of the event.",
+			"properties": {
+				"city": {
+					"type": "string"
+				},
+				"country_short": {
+					"type": "string",
```
miike, line 29: If `country_long` is nullable should country short be nullable as well?

schemas/com.mandrill/message_clicked/jsonschema/1-0-2
```diff
@@ -0,0 +1,290 @@
...
+						"type": "string"
+					}
+				},
+				"ts": {
+					"type": "string",
+					"description": "The timestamp when the message was sent",
+					"format": "date-time"
+				},
+				"resends": {
+					"type": "array"
```
miike, line 166: Are we able to type this array?

miike:
Thanks Matus this looks good. @istreeter anything we need to be aware of with toggling `additonalProperties: true` for any of the loaders?

schemas/com.mandrill/message_clicked/jsonschema/1-0-2
matus-tomlein: Unfortunately, this is a property that does not even appear in the Mandrill docs and I don't have any data for it. I only kept it in because it was in the previous versions of the schemas.

schemas/com.mandrill/message_clicked/jsonschema/1-0-2
matus-tomlein: This is a question I was pondering a lot when updating the schemas – almost all of the properties in the previous schema version were non-required but non-nullable. I was thinking of making them all non-required but nullable. In the end, I went for minimal changes to the schema since these are just patch updates. I'm happy to make it nullable, but in that case, should we change everything to nullable to be consistent?

matus-tomlein:
@istreeter Could you please take a look at the Mandrill schemas here and Mike's comment above?

istreeter:
> anything we need to be aware of with toggling `additonalProperties: true` for any of the loaders?

Some of our loaders load the data exactly as it gets received, without using the Iglu schema for guidance.  For example, the Snowflake loader, elasticsearch loader.  For these loaders, any additional property will get loaded to the warehouse.

Other loaders use the Iglu schema to manipulate the data into stricter types.  For example, Redshift, Databricks, Bigquery loaders.  For these loaders, any additional property will get ignored.

In both cases, it is completely OK to toggle it from false to true, or from true to false, within a series of schemas.

miike (approved):

PR #1343, opened 2023-09-18
Add com.snowplowanalytics.snowplow/browser_context/jsonschema/2-0-0

### LINKED ISSUE #1330: Change integer to number for deviceMemory in com.snowplowanalytics.snowplow/browser_context/jsonschema/1-0-0

igneel64:
Fix the `deviceMemory` type for lower end devices. 
Thanks @dumkydewilde for raising the issue and initial contribution.

close #1330

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/browser_context/jsonschema/2-0-0
+++ b/schemas/com.snowplowanalytics.snowplow/browser_context/jsonschema/2-0-0
@@ -1,0 +1,81 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for browser contexts",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow",
+    "name": "browser_context",
+    "format": "jsonschema",
+    "version": "2-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "viewport": {
+      "type": "string",
+      "maxLength": 20,
+      "description": "Viewport dimensions of the browser. Arrives in the form of WidthxHeight e.g. 1200x900."
+    },
+    "documentSize": {
+      "type": "string",
+      "maxLength": 20,
+      "description": "Document dimensions. Arrives in the form of WidthxHeight e.g. 1200x900"
+    },
+    "resolution": {
+      "type": "string",
+      "maxLength": 20,
+      "description": "Device native resolution. Arrives in the form of WidthxHeight e.g. 1200x900"
+    },
+    "colorDepth": {
+      "type": "integer",
+      "minimum": 0,
+      "maximum": 1000,
+      "description": "The number of bits allocated to colors for a pixel in the output device, excluding the alpha channel."
+    },
+    "devicePixelRatio": {
+      "type": ["number", "null"],
+      "minimum": 0,
+      "maximum": 1000,
+      "description": "Ratio of the resolution in physical pixels to the resolution in CSS pixels for the current display device."
+    },
+    "cookiesEnabled": {
+      "type": "boolean",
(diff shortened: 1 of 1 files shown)


matus-tomlein (commented):
It looks good to me but I'm not very confident whether it's a non-breaking change – the [docs say](https://docs.snowplow.io/docs/understanding-tracking-design/versioning-your-data-structures/enterprise/) that changing a type of a field is a breaking change, not sure if changing `integer` to `number` would be different. So just hoping for @istreeter to confirm.

miike:
This will be a breaking change due to the data type from integer => float / double. Technically in Snowflake / BigQuery this is less of an issue (as this creates new columns0 but in Redshift the loader won't mutate the type of the column so this will need to be bumped to 2-0-0.

schemas/com.snowplowanalytics.snowplow/browser_context/jsonschema/2-0-0
```diff
@@ -0,0 +1,81 @@
...
+      "description": "The language of the HTML document. RFC 5646 https://datatracker.ietf.org/doc/html/rfc5646."
+    },
+    "webdriver": {
+      "type": ["boolean", "null"],
+      "description": "Indicates whether the user agent is controlled by automation."
+    },
+    "deviceMemory": {
+      "type": ["number", "null"],
+      "minimum": 0,
+      "maximum": 1000,
```
matus-tomlein, line 64: Nit: shouldn't we raise the maximum as well while we are here? 1TB is probably safe but there already are computers out there with more memory (e.g., Mac Pro has 1.5TB). I don't know what would be a good limit, but maybe we can set it to something in the order of 1 million GB to avoid problems in the future?

matus-tomlein (approved):
LGTM!

schemas/com.snowplowanalytics.snowplow/browser_context/jsonschema/2-0-0
adatzer: [Based on this document](https://www.w3.org/TR/device-memory/#sec-device-memory-js-api) this deviceMemory is the amount of RAM, which is calculated [as described here](https://www.w3.org/TR/device-memory/#computing-device-memory-value). This last link recommends an upper bound of 8GB, which i do also think is outdated. It looks we are for now safe with `1000`, but there is always the risk of user-agents not conforming, so we could indeed raise.

matus-tomlein: My understanding is that they recommend setting an upper bound when computing the device memory. I think what we are doing is different – we are saying that the max memory won't be over 1TB. I just think that if there are PCs on the market that go over this limit (which there are), we should probably raise it as there can be valid users with higher memory on their PCs.

matus-tomlein: Ahh checking the first link I see that they are saying that the `self.navigator.deviceMemory` follows the recommendation of the setting the upper bound. In that case we should be fine (unless they change the recommendation in the future). So nevermind, we can keep this for now.

adatzer:
Rebased and merged into `release/r148`. Closing.

PR #1350, opened 2023-11-06
Add load_succeeded/4-0-0

oguzhanunlu:


istreeter:
Did you consider making this a 3-0-1 schema instead of 4-0-0?  To do this, the schema would be basically the same, but you would make `recoverTableNames` a non-required field.

A major change is disruptive for downstream consumers of the data, because they need to consume from a different table.  In this case, you have not radically re-designed the shape of the data; you have only added a new field.  So I think downstream consumers would prefer to have new events land in the same table as the old events.

schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/4-0-0
```diff
@@ -0,0 +1,192 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for a data load succeeding",
+  "self": {
+    "vendor": "com.snowplowanalytics.monitoring.batch",
+    "name": "load_succeeded",
+    "format": "jsonschema",
+    "version": "3-0-0"
```
istreeter, line 8: This line should be 3-0-1

oguzhanunlu: thanks!

schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/3-0-1
```diff
@@ -0,0 +1,192 @@
...
+      "type": "string",
+      "format": "date-time"
+    },
+    "loadingCompleted": {
+      "description": "Time when loader finished loading (should match the ingestion timestamp in manifest)",
+      "type": "string",
+      "format": "date-time"
+    },
+    "recoveryTableNames": {
+      "description": "List of recovery table names created",
```
istreeter, line 179: > List of recovery table names created

Is it for tables _created_ by this batch, or tables _used_ by this batch?  In other words.... if a pipeline continually uses an invalid schema over several weeks, should it appear only once in the `load_succeeded_3` table, or should it appear for every batch that used the invalid schema?

oguzhanunlu: It is `created`, not `used`. If a pipeline keep using a broken schema, we expect to see the corresponding recovery table name once.

oguzhanunlu: > should it

Our discussions have been about reporting the creation if I remember right. On the other hand, stdout logs have both creation and usage separately.

If we want to change the semantic to `usage`, or add it, a batch could be loaded to the main table or a set of recovery table(s). In that case, we should have a `loadedTables` key.

istreeter (approved):

PR #1352, opened 2023-11-07
Add com.snowplowanalytics.snowplow/button_click/jsonschema/1-0-0 (close #1351)

### LINKED ISSUE #1351: Add com.snowplowanalytics.snowplow/button_click/jsonschema/1-0-0

greg-el:
This PR adds a schema for button clicks, in the same vein as the existing [link_click](https://github.com/snowplow/iglu-central/blob/8ff48b2485b3c95447e38a9bb925ef3f5266112c/schemas/com.snowplowanalytics.snowplow/link_click/jsonschema/1-0-1) schema. This will be used in the upcoming OTB Button Click event in the JS tracker.

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/button_click/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow/button_click/jsonschema/1-0-0
@@ -1,0 +1,50 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for a button click event",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow",
+    "name": "button_click",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "label": {
+      "type": "string",
+      "description": "The text on the button clicked, or a user-provided override",
+      "minLength": 1,
+      "maxLength": 4096
+    },
+    "id": {
+      "description": "The ID of the button clicked, if present",
+      "type": [
+        "string",
+        "null"
+      ],
+      "maxLength": 4096
+    },
+    "classes": {
+      "description": "An array of class names from the button clicked",
+      "type": [
+        "array",
+        "null"
+      ],
+      "items": {
+        "type": "string",
+        "maxLength": 4096
+      }
+    },
+    "name": {
+      "description": "The name of the button, if present",
+      "type": [
+        "string",
+        "null"
+      ],
+      "maxLength": 4096
+    }
+  },
+  "required": [
+    "label"
+  ],
+  "additionalProperties": false
+}


schemas/com.snowplowanalytics.snowplow/button_click/jsonschema/1-0-0
```diff
@@ -0,0 +1,60 @@
...
+    "vendor": "com.snowplowanalytics.snowplow",
+    "name": "button_click",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "action": {
+      "type": "string",
+      "description": "The action associated with the button click",
```
matus-tomlein, line 14: How would we know the action? Is it something that needs to be passed from the user or can be identified automatically?

schemas/com.snowplowanalytics.snowplow/button_click/jsonschema/1-0-0
```diff
@@ -0,0 +1,60 @@
...
+      "type": [
+        "array",
+        "null"
+      ],
+      "items": {
+        "type": "string"
+      }
+    },
+    "elementName": {
+      "description": "The name of the element if present",
```
matus-tomlein, line 38: Would this be the value of a `name` property? I don't think it's very common to set the `name` for buttons, but might be wrong.

schemas/com.snowplowanalytics.snowplow/button_click/jsonschema/1-0-0
```diff
@@ -0,0 +1,60 @@
...
+      "type": [
+        "string",
+        "null"
+      ],
+      "minLength": 1,
+      "maxLength": 4096
+    },
+    "elementContent": {
+      "description": "The content of the element if present and enabled",
```
matus-tomlein, line 47: This is basically the label of the button, right? Why not just have a `label` property?

schemas/com.snowplowanalytics.snowplow/button_click/jsonschema/1-0-0
greg-el: I got `name` this from the [button spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#name), but yeah, I'm not sure how common it would actually be to set it (and we want to avoid overlapping this feature with form tracking). Perhaps it can be removed?

schemas/com.snowplowanalytics.snowplow/button_click/jsonschema/1-0-0
greg-el: I was trying to keep this consistent with `link_click` (and gtm to an extent), but `label` works fine for me too.

schemas/com.snowplowanalytics.snowplow/button_click/jsonschema/1-0-0
greg-el: It would have to be identifier by the user. I was thinking initially that we could just take the text on the button as an "action", but there are enough icon buttons out there that that's probably not do-able, even with something like aria-labels, since they're often missing. 

I was thinking was that this could be added via a data attribute (something like `data-sp-action` on a button?), or with the usual `<button onClick="() => trackButtonClick({action: "foo"})"></button>`.

matus-tomlein: If it's intended to be the text on the button, then maybe it's better to call it the `label`? But then I'm not sure how would this be different from the `elementContent`.

I think it would make sense if, by default, the text of the button was used here. That is the most common way to describe what a button does (on all platforms) so probably should be the only required property in the schema.

It's true that in some cases, buttons may not have text, but given that we limit the scope of this schema to `button_click` as opposed to `click`, that should be rare enough. And for those cases we could let users provide the label manually in some way.

schemas/com.snowplowanalytics.snowplow/button_click/jsonschema/1-0-0
matus-tomlein: Nevermind, I have seen this in other button click events as well, happy to leave as is.

schemas/com.snowplowanalytics.snowplow/button_click/jsonschema/1-0-0
greg-el: `elementContent` has been removed in favour of the `label` property.

schemas/com.snowplowanalytics.snowplow/button_click/jsonschema/1-0-0
greg-el: Sounds good, `action` has been renamed `label` and we can take the text on the button if present by default, and provide a way for the user to provide a label manually.

schemas/com.snowplowanalytics.snowplow/button_click/jsonschema/1-0-0
```diff
@@ -0,0 +1,53 @@
...
+        "maxLength": 4096
+      }
+    },
+    "name": {
+      "description": "The name of the button, if present",
+      "type": [
+        "string",
+        "null"
+      ],
+      "minLength": 1,
```
matus-tomlein, line 45: Nit: this seems a bit strict to require non-empty strings for optional fields. I am happy to go with it, but something about invalidating events for this (it can be quite a common problem that developers don't handle) makes me nervous.

matus-tomlein (approved):
LGTM!

schemas/com.snowplowanalytics.snowplow/button_click/jsonschema/1-0-0
greg-el: Makes sense, I've removed the minLengths.

igneel64 (approved):

PR #1348, opened 2023-10-27
Release R148

matus-tomlein:
This release add schemas for:

* updated schemas for the Mandrill adapter (reviewed in PR #1342)
* new major version of the browser context schema (reviewed in PR #1343)
* new schema for GA4 cookies (reviewed in #1325)
* new schema for AMP tracker consent (reviewed in #1329)

## Changelog

Add com.google.ga4/cookies/jsonschema/1-0-0 (#1324)
Add schemas/dev.amp.snowplow/amp_consent/jsonschema/1-0-0 (#1328)
Add com.snowplowanalytics.snowplow/browser_context/jsonschema/2-0-0 (#1330)
Add com.mandrill/recipient_unsubscribed/jsonschema/1-0-2 (#1341)
Add com.mandrill/message_soft_bounced/jsonschema/1-0-2 (#1340)
Add com.mandrill/message_sent/jsonschema/1-0-1 (#1339)
Add com.mandrill/message_rejected/jsonschema/1-0-1 (#1338)
Add com.mandrill/message_opened/jsonschema/1-0-2 (#1337)
Add com.mandrill/message_marked_as_spam/jsonschema/1-0-2 (#1336)
Add com.mandrill/message_delivered/jsonschema/1-0-0 (#1335)
Add com.mandrill/message_delayed/jsonschema/1-0-2 (#1334)
Add com.mandrill/message_clicked/jsonschema/1-0-2 (#1333)
Add com.mandrill/message_bounced/jsonschema/1-0-2 (#1332)

PR #1353, opened 2023-11-08
Release R148

jbeemster:
Splits from https://github.com/snowplow/iglu-central/pull/1348 as I want extra time to validate the Mandrill changes in there - specifically the change in field type to allow nullability.

### DIFF
--- a/schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/3-0-1
+++ b/schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/3-0-1
@@ -1,0 +1,192 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for a data load succeeding",
+  "self": {
+    "vendor": "com.snowplowanalytics.monitoring.batch",
+    "name": "load_succeeded",
+    "format": "jsonschema",
+    "version": "3-0-1"
+  },
+
+  "type": "object",
+  "properties": {
+    "shredding": {
+      "description": "Information about the batch taken from transformation step, isomorphic to shredding_complete schema",
+      "type": "object",
+      "properties": {
+        "base": {
+          "description": "Blob storage path to the root of the batch",
+          "type": "string",
+          "format": "uri",
+          "maxLength": 1024
+        },
+        "compression": {
+          "description": "File compression type",
+          "enum": ["GZIP", "NONE"]
+        },
+        "typesInfo": {
+          "description": "Info about schemas used in events and output formats in the respective batch",
+          "type": "object",
+          "oneOf": [
+            {
+              "properties": {
+                "transformation": {
+                  "description": "Type of the transformation",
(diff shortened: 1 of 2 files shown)


schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/3-0-1
```diff
@@ -0,0 +1,192 @@
...
+      "format": "date-time"
+    },
+    "loadingCompleted": {
+      "description": "Time when loader finished loading (should match the ingestion timestamp in manifest)",
+      "type": "string",
+      "format": "date-time"
+    },
+    "recoveryTableNames": {
+      "description": "List of recovery table names loaded in this batch",
+      "type": "array",
```
AlexBenny, line 180: I think we should put null as extra type (at least decide a policy). In this same schema we have a precedent for count field line 145.

AlexBenny (commented):
I've just raised the point about null type for the new field. Open to any decision but it would be good to agree on a unique policy to follow for everyone.

istreeter (approved):

schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/3-0-1
jbeemster: @istreeter has stated its not needed so I think we should move to not requiring this and trusting that we do not need to define "null" in optional fields anymore.  This likely also means removing that advice from `iglu` itself given that its not strictly needed and doesn't add anything meaningful?

AlexBenny (approved):

schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/3-0-1
AlexBenny: I’m ok for this schema, but as I pointed up we would have the schema adopting two different approaches (line 145). 
> its not needed so I think we should move to not requiring this and trusting that we do not need to define "null" in optional fields anymore.
IMO it's not totally correct because a not required property is a property that doesn't compare in the json payload. If the property compare but it's null it would be invalidated. If we add the null type to the property we are sure that even a null property would be considered accepted.
I don't know if it's ok in this case but I know this was the extra safety adopted for all the schemas developed by Activation to prevent those kind of issues, and probably the reason why it's in `igluctl` as a warning (?!).

jbeemster: @istreeter @AlexBenny should I just add the null type for the sake of history and we can have the semantic debate on how this should be ideally going forward at another time?

jbeemster:
Diff:

```
8c8
<     "version": "3-0-0"
---
>     "version": "3-0-1"
177a178,184
>     "recoveryTableNames": {
>       "description": "List of recovery table names loaded in this batch",
>       "type": ["array", "null"],
>       "items": {
>         "type": "string"
>       }
>     },
```

AlexBenny (approved):

istreeter (approved):

PR #1358, opened 2023-11-25
Add schemas for screen engagement tracking on mobile

matus-tomlein:
This PR adds 3 schemas for screen engagement tracking on mobile (`screen_time` and `screen_end`) and both on mobile and Web (`view_segment`).

These schema support the proposal presented [in this demo](https://f001.backblazeb2.com/file/referred/snowplow-screen-engagement-demo/index.html).

* com.snowplowanalytics.snowplow/screen_summary/jsonschema/1-0-0
   * schema for entity tracked with mobile events (foreground/background/scren_end) with duration of the screen in foreground and background
* com.snowplowanalytics.snowplow/screen_end/jsonschema/1-0-0
   * schema for event tracked before transitioning to the next screen on mobile. it doesn't have any properties but will be tracked with the `screen_time` and `view_segment` entities

### DIFF
--- a/schemas/com.snowplowanalytics.mobile/list_item_view/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.mobile/list_item_view/jsonschema/1-0-0
@@ -1,0 +1,32 @@
+{
+	"$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+	"description": "Schema for an event tracked when an item is displayed in a list",
+	"self": {
+		"vendor": "com.snowplowanalytics.mobile",
+		"name": "list_item_view",
+		"format": "jsonschema",
+		"version": "1-0-0"
+	},
+	"type": "object",
+	"properties": {
+		"index": {
+			"type": "integer",
+			"description": "Index of the item in a list on the screen",
+			"minimum": 0,
+			"maximum": 65535
+		},
+		"items_count": {
+			"type": [
+				"integer",
+				"null"
+			],
+			"description": "Total number of items in a list on the screen",
+			"minimum": 0,
+			"maximum": 65535
+		}
+	},
+	"required": [
+		"index"
+	],
+	"additionalProperties": false
+}
--- a/schemas/com.snowplowanalytics.mobile/screen_summary/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.mobile/screen_summary/jsonschema/1-0-0
@@ -1,0 +1,104 @@
+{
+	"$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+	"description": "Schema for an entity tracked with foreground/background/screen_end events with summary statistics about the screen view",
+	"self": {
+		"vendor": "com.snowplowanalytics.mobile",
+		"name": "screen_summary",
+		"format": "jsonschema",
+		"version": "1-0-0"
+	},
+	"type": "object",
+	"properties": {
+		"foreground_sec": {
+			"type": "number",
+			"description": "Time in seconds spent on the current screen while the app was in foreground",
+			"minimum": 0,
+			"maximum": 2147483647
+		},
+		"background_sec": {
+			"type": [
+				"number",
+				"null"
+			],
+			"description": "Time in seconds spent on the current screen while the app was in background",
+			"minimum": 0,
+			"maximum": 2147483647
+		},
+		"last_item_index": {
+			"type": [
+				"integer",
+				"null"
+			],
+			"description": "Index of the last viewed item in the list on the screen",
+			"minimum": 0,
+			"maximum": 65535
+		},
+		"items_count": {
+			"type": [
+				"integer",
+				"null"
+			],
+			"description": "Total number of items in the list on the screen",
+			"minimum": 0,
+			"maximum": 65535
+		},
+		"min_x_offset": {
+			"type": [
+				"integer",
+				"null"
+			],
+			"description": "Minimum horizontal scroll offset on the scroll view in pixels",
+			"minimum": -2147483647,
+			"maximum": 2147483647
+		},
+		"max_x_offset": {
+			"type": [
+				"integer",
+				"null"
+			],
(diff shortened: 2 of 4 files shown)


rlh1994 (commented):
A few Qs

schemas/com.snowplowanalytics.snowplow/screen_time/jsonschema/1-0-0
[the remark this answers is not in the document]
matus-tomlein: My thought was to send this with background/foreground events and also with the new `screen_end` event (below). So basically every time the screen changes visibility or we move to another screen. We should also be able to get the screen view ID from the screen context entity.

schemas/com.snowplowanalytics.snowplow/screen_end/jsonschema/1-0-0
```diff
@@ -0,0 +1,13 @@
...
+	"self": {
+		"vendor": "com.snowplowanalytics.snowplow",
+		"name": "screen_end",
+		"format": "jsonschema",
+		"version": "1-0-0"
+	},
+	"type": "object",
+	"properties": {},
+	"additionalProperties": false
+}
```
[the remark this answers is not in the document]
matus-tomlein: Well it's included by default. It is possible to disable, but it's a similar case like the web_page context entity, we kind of assume that it is enabled for everything else to work properly...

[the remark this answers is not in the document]
rlh1994: If it's on by default then that's fine!

schemas/com.snowplowanalytics.snowplow/screen_end/jsonschema/1-0-0
rlh1994, line 1: Last question, do we have any other events that have no properties (that's aren't the page ping). Just wondering how this will work with the loader in terms of creating a column... @jbeemster ?

matus-tomlein: We do have a few events like that – the [`application_install` event](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.mobile/application_install/jsonschema/1-0-0) is quite a common one for example. Checking our dev1 Snowflake atomic events, I don't see a column for it, so probably the column is not created if the event schema has no properties?

matus-tomlein:
As I was implementing this in the iOS tracker, I iterated a bit on the schemas. I recorded a short video to demo how the schemas will be used in tracking, it [can be found here](https://snplow.atlassian.net/browse/PE-5762?focusedCommentId=70520).

Overall the changes are:

1. Renamed the vendor for all schemas to `com.snowplowanalytics.mobile`. All of these schemas are for mobile trackers and this is also the vendor used for the screen view event which they are used with.
2. In the `screen_summary` entity schema, I renamed the list item properties to be: `last_item_index` and `items_count`. So they will contain the index of the last viewed item and the number of all items in the list.
3. I added a third event schema for `list_item_view` event. If screen engagement is enabled on mobile, this event will be aggregated into the `screen_summary` context entity. It is just used as a way for users to update the currently seen list items.

schemas/com.snowplowanalytics.mobile/list_item_view/jsonschema/1-0-0
```diff
@@ -0,0 +1,32 @@
...
+		},
+		"items_count": {
+			"type": [
+				"integer",
+				"null"
+			],
+			"description": "Total number of list items on the screen",
+			"minimum": 0,
+			"maximum": 65535
+		}
```
rlh1994, line 26: Dumb question, more just for my knowledge - would these be 0 indexed so there might be 5 items but the index of the last item will be 4?

schemas/com.snowplowanalytics.mobile/screen_summary/jsonschema/1-0-0
```diff
@@ -0,0 +1,50 @@
...
+			"description": "Index of the last viewed item in a list on the screen",
+			"minimum": 0,
+			"maximum": 65535
+		},
+		"items_count": {
+			"type": [
+				"integer",
+				"null"
+			],
+			"description": "Total number of list items on the screen",
```
rlh1994, line 41: These descriptions confuse me a little, total number of list items, but the index of the last viewed item _in_ a list? Maybe something like 'Total number of items in the list on the screen'? Or can you have multiple lists on one screen?

rlh1994 (approved):
Just a few questions because I don't develop mobile apps so don't really know what this means

schemas/com.snowplowanalytics.mobile/list_item_view/jsonschema/1-0-0
rlh1994: No I think this is the right way to do it!

schemas/com.snowplowanalytics.mobile/screen_summary/jsonschema/1-0-0
matus-tomlein: Good point! Have unified the wording, it is expected that there is only one list on the screen.

matus-tomlein:
Final update: after some discussion, we also added scroll view tracking in pixels (in addition to the list view tracking). We can't provide automatic tracking for this, but users can listen for scroll changed events in their scroll views and track a `scroll_changed` event. This is handled the same as the `list_item_view` event and aggregated in the `screen_summary` entity – there are two new properties: `max_y_offset` and `content_height`.

rlh1994:
>there are two new properties: max_y_offset and content_height

Can mobile applications have an x offset and width greater than the screen?

matus-tomlein:
It'd be quite rare, but actually why not add it since we also have the horizontal scroll depth on Web too... Have updated the schemas to make it possible to track too.

matus-tomlein:
One more update: have also added `min_x_offset` and `min_y_offset` to match with scroll depth tracking on the Web.

rlh1994 (approved):
LGTM!

mscwilson (approved):
LGTM

matus-tomlein:
Cherry picked to R152

PR #1362, opened 2023-12-18
Add schemas for visionOS

mscwilson:
This PR adds schemas based on various SwiftUI objects. They will allow users to track activity in visionOS apps. For issue #1367.

### DIFF
--- a/schemas/com.apple.swiftui/dismiss_immersive_space/jsonschema/1-0-0
+++ b/schemas/com.apple.swiftui/dismiss_immersive_space/jsonschema/1-0-0
@@ -1,0 +1,13 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "self": {
+    "vendor": "com.apple.swiftui",
+    "name": "dismiss_immersive_space",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "description": "Schema for an event for dismissing a visionOS immersive space.",
+  "type": "object",
+  "properties": {},
+  "additionalProperties": false
+}
--- a/schemas/com.apple.swiftui/dismiss_window/jsonschema/1-0-0
+++ b/schemas/com.apple.swiftui/dismiss_window/jsonschema/1-0-0
@@ -1,0 +1,13 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "self": {
+    "vendor": "com.apple.swiftui",
+    "name": "dismiss_window",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "description": "Schema for an event for dismissing a SwiftUI window or window group.",
+  "type": "object",
+  "properties": {},
+  "additionalProperties": false
+}
--- a/schemas/com.apple.swiftui/immersive_space/jsonschema/1-0-0
+++ b/schemas/com.apple.swiftui/immersive_space/jsonschema/1-0-0
@@ -1,0 +1,48 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "self": {
+    "vendor": "com.apple.swiftui",
+    "name": "immersive_space",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "description": "Schema for an immersive space entity, representing the VisionOS immersive space that the event occurs in.",
+  "type": "object",
+  "properties": {
+    "view_id": {
+      "type": ["string", "null"],
+      "format": "uuid",
+      "description": "UUID for the view of the immersive space."
+    },
+    "id": {
+      "type": "string",
+      "description": "The identifier of the immersive space to present.",
+      "maxLength": 255
+    },
+    "immersion_style": {
+      "type": ["string", "null"],
+      "enum": [
+        "automatic",
+        "full",
+        "mixed",
+        "progressive",
+        null
+      ],
+      "description": "The style of an immersive space."
+    },
+    "upper_limb_visibility": {
+      "type": ["string", "null"],
+      "enum": [
+        "automatic",
+        "visible",
+        "hidden",
+        null
+      ],
+      "description": "Preferred visibility of the user's upper limbs, while an immersive space scene is presented."
+    }
+  },
+  "required": [
+    "id"
+  ],
+  "additionalProperties": false
+}
--- a/schemas/com.apple.swiftui/open_immersive_space/jsonschema/1-0-0
+++ b/schemas/com.apple.swiftui/open_immersive_space/jsonschema/1-0-0
@@ -1,0 +1,13 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "self": {
+    "vendor": "com.apple.swiftui",
+    "name": "open_immersive_space",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "description": "Schema for an event for opening a visionOS immersive space.",
+  "type": "object",
+  "properties": {},
+  "additionalProperties": false
+}
--- a/schemas/com.apple.swiftui/open_window/jsonschema/1-0-0
+++ b/schemas/com.apple.swiftui/open_window/jsonschema/1-0-0
@@ -1,0 +1,13 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "self": {
+    "vendor": "com.apple.swiftui",
+    "name": "open_window",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "description": "Schema for an event for opening a SwiftUI window or window group.",
+  "type": "object",
+  "properties": {},
+  "additionalProperties": false
+}
--- a/schemas/com.apple.swiftui/window_group/jsonschema/1-0-0
+++ b/schemas/com.apple.swiftui/window_group/jsonschema/1-0-0
@@ -1,0 +1,47 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "self": {
+    "vendor": "com.apple.swiftui",
+    "name": "window_group",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "description": "Schema for a window group entity, representing the SwiftUI window group that the event occurs in.",
+  "type": "object",
+  "properties": {
+    "window_id": {
+      "type": ["string", "null"],
+      "format": "uuid",
+      "description": "UUID for the current window within the group."
+    },
+    "id": {
+      "type": "string",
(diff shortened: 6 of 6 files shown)


schemas/com.apple.arkit/trackable_anchor/jsonschema/1-0-0
```diff
@@ -0,0 +1,66 @@
...
+      "enum": [
+        "device",
+        "world",
+        "hand",
+        "image"
+      ],
+      "description": "Type of the anchor."
+    },
+    "description": {
+      "type": "string",
```
miike, line 28: Is description nullable?

schemas/com.apple.swiftui/dismiss_window/jsonschema/1-0-0
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "self": {
+    "vendor": "com.apple.swiftui",
+    "name": "dismiss_window",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "description": "Schema for a window being dismissed (VisionOS)",
```
matus-tomlein, line 9: This is not strictly limited to visionOS.

```suggestion
  "description": "Schema for a SwiftUI window being dismissed",
```

schemas/com.apple.swiftui/open_window/jsonschema/1-0-0
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "self": {
+    "vendor": "com.apple.swiftui",
+    "name": "open_window",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "description": "Schema for a window being opened (VisionOS)",
```
matus-tomlein, line 9:
```suggestion
  "description": "Schema for a SwiftUI window being opened",
```

schemas/com.apple.swiftui/window_group/jsonschema/1-0-0
```diff
@@ -0,0 +1,47 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "self": {
+    "vendor": "com.apple.swiftui",
+    "name": "window_group",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "description": "Schema for a window group scene entity (VisionOS)",
```
matus-tomlein, line 9:
```suggestion
  "description": "Schema for the SwiftUI window group that the event belongs to",
```

schemas/com.apple.swiftui/window_group/jsonschema/1-0-0
```diff
@@ -0,0 +1,47 @@
...
+  "type": "object",
+  "properties": {
+    "id": {
+      "type": "string",
+      "format": "uuid",
+      "description": "Randomly generated ID for the current window within the group. Generated by the tracker if not provided by the user."
```
matus-tomlein, line 15: This can only be generated by the tracker if the tracker then remembers the ID and can automatically attach to all events from the window. But I think this is not currently implemented in the tracker and it's also quite difficult to implemnet because the tracker is not aware which window the event is coming from – there may be multiple open at the same time. So I think we need t o leave up to the users to manage.

```suggestion
      "description": "Randomly generated ID for the current window within the group."
```

schemas/com.apple.swiftui/immersive_space/jsonschema/1-0-0
```diff
@@ -0,0 +1,48 @@
...
+    },
+    "immersive_space_id": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "description": "The identifier of the immersive space to present."
+    },
+    "immersion_style": {
+      "type": "string",
```
matus-tomlein, line 25: This should be nullable since it is not required. To do that, you'll need to add null to both the type and the enum values, [see here](https://github.com/snowplow/iglu-central/blob/f89b6aea3231abb49fb86e87f999b66f9131db3e/schemas/com.snowplowanalytics.snowplow/mobile_context/jsonschema/1-0-2#L72-L73)

schemas/com.apple.swiftui/immersive_space/jsonschema/1-0-0
```diff
@@ -0,0 +1,48 @@
...
+      "enum": [
+        "automatic",
+        "full",
+        "mixed",
+        "progressive"
+      ],
+      "description": "The style of an immersive space."
+    },
+    "upper_limb_visibility": {
```
matus-tomlein, line 34: This should also be nullable.

schemas/com.apple.swiftui/window_group/jsonschema/1-0-0
```diff
@@ -0,0 +1,47 @@
...
+    },
+    "title_key": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "description": "A localized string key to use for the window's title in system menus and in the window's title bar. Provide a title that describes the purpose of the window."
+    },
+    "window_style": {
+      "type": "string",
```
matus-tomlein, line 32: Should also be nullable.

schemas/com.apple.swiftui/open_immersive_space/jsonschema/1-0-0
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "self": {
+    "vendor": "com.apple.swiftui",
+    "name": "open_immersive_space",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "description": "Schema for an immersive space being opened (VisionOS)",
```
matus-tomlein, line 9: Just to clarify a bit that it's an event.

```suggestion
  "description": "Schema for an event for opening an immersive space (visionOS)",
```

schemas/com.apple.swiftui/dismiss_immersive_space/jsonschema/1-0-0
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "self": {
+    "vendor": "com.apple.swiftui",
+    "name": "dismiss_immersive_space",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "description": "Schema for an immersive space being dismissed (VisionOS)",
```
matus-tomlein, line 9:
```suggestion
  "description": "Schema for an event for dismissing an immersive space (visionOS)",
```

schemas/com.apple.swiftui/immersive_space/jsonschema/1-0-0
```diff
@@ -0,0 +1,48 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "self": {
+    "vendor": "com.apple.swiftui",
+    "name": "immersive_space",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "description": "Schema for an immersive space entity (VisionOS)",
```
matus-tomlein, line 9:
```suggestion
  "description": "Schema for an immersive space entity (visionOS)",
```

matus-tomlein (commented):
Just a few small comments. Could you also rearrange the commits to have one commit per schema? I'll then be able to cherry pick them into the release branch.

schemas/com.apple.swiftui/immersive_space/jsonschema/1-0-0
mscwilson: I see that some nullable enums have null in their values list, but I don't think it is required? Change_form and focus_form don't

matus-tomlein: I think in the newer schemas, we tend to add `null` in the values list (see [the media player](https://github.com/snowplow/iglu-central/commit/25d59d3e39bb54b8c9b431a3c435f8ea4a15b5f8#diff-1b5edede2341bc080382bc706d0f53d6d37514754c8059496a2fd118bb4c7e1eR60-R70)). I remember Ian telling me that it was required for some reason, but can't find the conversation, you can double check with him to be sure.

schemas/com.apple.swiftui/window_group/jsonschema/1-0-0
```diff
@@ -0,0 +1,48 @@
...
+    "name": "window_group",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "description": "Schema for a window group entity, representing the SwiftUI window group that the event occurs in.",
+  "type": "object",
+  "properties": {
+    "id": {
+      "type": "string",
+      "format": "uuid",
```
matus-tomlein, line 14: I am wondering since this ID will come from the user (it's not generated by the tracker), whether we should be so strict and require it to be in the UUID format? Might be easier for users in case they already have some other identifier of the window, to just pass that. What do you think?

mscwilson: I think it's good to have the UUID option, but it would probably work better to make the string ID the required one yes (since string ID is used to make the WindowGroup)

matus-tomlein: Should we name it something like `windowId` (instead of `uuid`) to make it clear what it identifies?

schemas/com.apple.swiftui/immersive_space/jsonschema/1-0-0
```diff
@@ -0,0 +1,47 @@
...
+  "description": "Schema for an immersive space entity, representing the VisionOS immersive space that the event occurs in.",
+  "type": "object",
+  "properties": {
+    "view_id": {
+      "type": ["string", "null"],
+      "format": "uuid",
+      "description": "UUID for the view of the immersive space."
+    },
+    "id": {
+      "type": "string",
```
matus-tomlein, line 18: One more thing before merging this into the release branch – could you add a max length here? We usually go with something like 255, 1024, 4096 – I think 255 would be good.

schemas/com.apple.swiftui/window_group/jsonschema/1-0-0
```diff
@@ -0,0 +1,45 @@
...
+  },
+  "description": "Schema for a window group entity, representing the SwiftUI window group that the event occurs in.",
+  "type": "object",
+  "properties": {
+    "window_id": {
+      "type": ["string", "null"],
+      "format": "uuid",
+      "description": "UUID for the current window within the group."
+    },
+    "id": {
```
matus-tomlein, line 17: Also please add max length here (255?)

schemas/com.apple.swiftui/window_group/jsonschema/1-0-0
```diff
@@ -0,0 +1,45 @@
...
+      "type": "string",
+      "description": "A string that uniquely identifies the window group. Identifiers must be unique among the window groups in your app."
+    },
+    "title_key": {
```
matus-tomlein, line 21: And here – I think 4096 would be a safe choice.

matus-tomlein (approved):
LGTM!

matus-tomlein:
Cherry picked to R152

PR #1363, opened 2023-12-19
Add new loader alert schema

pondzix:
Similar to the alert schema used by RDB Loader -> https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.monitoring.batch/alert/jsonschema/1-0-0, but only with 3 fields: 

* `application`
* `message`
* `tags`

### DIFF
--- a/schemas/com.snowplowanalytics.monitoring.loader/alert/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.monitoring.loader/alert/jsonschema/1-0-0
@@ -1,0 +1,35 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Webhook payload describing an alert coming from loader",
+  "self": {
+    "vendor": "com.snowplowanalytics.monitoring.loader",
+    "name": "alert",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "appName": {
+      "type": "string",
+      "maxLength": 64,
+      "description": "Application's name"
+    },
+    "appVersion": {
+      "type": "string",
+      "maxLength": 64,
+      "description": "Application's version"
+    },
+    "message": {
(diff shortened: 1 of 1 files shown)


schemas/com.snowplowanalytics.monitoring.loader/alert/jsonschema/1-0-0
```diff
@@ -0,0 +1,29 @@
...
+      "description": "Application's name and version"
+    },
+    "message": {
+      "description": "Free-form message describing the problem",
+      "maxLength": 4096,
+      "type": "string"
+    },
+    "tags": {
+      "description": "Set of key value pairs providing additional information",
+      "type": "object"
```
istreeter, line 24: For tags, how about

```
"additionalProperties": { "type": "string" }
```

schemas/com.snowplowanalytics.monitoring.loader/alert/jsonschema/1-0-0
```diff
@@ -0,0 +1,30 @@
...
+  "description": "Webhook payload describing an alert coming from loader",
+  "self": {
+    "vendor": "com.snowplowanalytics.monitoring.loader",
+    "name": "alert",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "application": {
```
jbeemster, line 12: It might be easier to query this if the application name and version were in different fields?

pondzix: Why not. Then `application` field would be just name and `version` would just version?

jbeemster: Yeah essentially you would have the app_name and app_version as different fields - but if the current convention is to have them concatenated we can leave it as is - not a strong feeling!

pondzix:
Released in https://github.com/snowplow/iglu-central/pull/1365, closing

PR #1387, opened 2024-02-13
Add/ga4 schemas

adatzer:
This PR cherry-picks #1245 and splits adding the schemas into commits.

### DIFF
--- a/schemas/com.google.tag-manager.server-side/post_score/jsonschema/1-0-0
+++ b/schemas/com.google.tag-manager.server-side/post_score/jsonschema/1-0-0
@@ -1,0 +1,28 @@
+{
+        "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+        "self": {
+                "vendor": "com.google.tag-manager.server-side",
+                "name": "post_score",
+                "version": "1-0-0",
+                "format": "jsonschema"
+        },
+        "description": "Send this event when the user posts a score. Use this event to understand how users are performing in your game and correlate high scores with audiences or behaviors.",
+        "type": "object",
+        "properties": {
+                "score": {
+                        "description": "The score to post.",
+                        "type": ["number", "null"]
+                },
+                "level": {
+                        "description": "The level for the score.",
+                        "type": ["number", "null"]
+                },
+                "character": {
+                        "description": "The character that achieved the score.",
+                        "type": ["string", "null"],
+                        "maxLength": 4096
+                }
+        },
+        "required": ["score"],
+        "additionalProperties": false
+}
--- a/schemas/com.google.tag-manager.server-side/spend_virtual_currency/jsonschema/1-0-0
+++ b/schemas/com.google.tag-manager.server-side/spend_virtual_currency/jsonschema/1-0-0
@@ -1,0 +1,30 @@
+{
+        "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+        "self": {
+                "vendor": "com.google.tag-manager.server-side",
+                "name": "spend_virtual_currency",
+                "version": "1-0-0",
+                "format": "jsonschema"
+        },
+        "description": "This event measures the sale of virtual goods in your app and helps you identify which virtual goods are the most popular.",
+        "type": "object",
+        "properties": {
+                "virtual_currency_name": {
+                        "description": "The name of the virtual currency.",
+                        "type": ["string", "null"],
+                        "maxLength": 4096
+                },
+                "value": {
+                        "description": "The value of the virtual currency.",
+                        "type": ["number", "null"],
+                        "minimum": 0
+                },
+                "item_name": {
+                        "description": "The name of the item the virtual currency is being used for.",
+                        "type": ["string", "null"],
+                        "maxLength": 500
+                }
+        },
+        "required": ["virtual_currency_name", "value"],
+        "additionalProperties": false
+}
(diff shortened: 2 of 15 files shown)


istreeter (commented):
The schemas look good.  My only comment is about monetary values.

A field with `"type": "number"` will get loaded into the warehouse as a double.  If you add `"multipleOf": 0.01` then it gets loaded as a decimal, which might be more convenient for the end user working with the data.  (Although I am not an expert in how analysts work with this data).

See for example the following schemas which specify a `multipleOf` for monetary values:

- schemas/com.google.analytics.ecommerce/item/jsonschema/1-0-0
- schemas/com.google.analytics.ecommerce/transaction/jsonschema/1-0-0
- schemas/com.google.analytics.enhanced-ecommerce/actionFieldObject/jsonschema/1-0-0
- schemas/com.google.analytics.enhanced-ecommerce/impressionFieldObject/jsonschema/1-0-0
- schemas/com.google.analytics.enhanced-ecommerce/productFieldObject/jsonschema/1-0-0
- schemas/com.google.analytics.measurement-protocol/item/jsonschema/1-0-0
- schemas/com.google.analytics.measurement-protocol/product/jsonschema/1-0-0
- schemas/com.google.analytics.measurement-protocol/product_impression/jsonschema/1-0-0
- schemas/com.google.analytics.measurement-protocol/transaction/jsonschema/1-0-0
- schemas/com.snowplowanalytics.snowplow.ecommerce/cart/jsonschema/1-0-0
- schemas/com.snowplowanalytics.snowplow.ecommerce/product/jsonschema/1-0-0
- schemas/com.snowplowanalytics.snowplow.ecommerce/refund/jsonschema/1-0-0
- schemas/com.snowplowanalytics.snowplow.ecommerce/transaction/jsonschema/1-0-0

miike:
>If you add "multipleOf": 0.01 then it gets loaded as a decimal, which might be more convenient for the end user working with the data. (Although I am not an expert in how analysts work with this data).

@istreeter this makes sense, the other schemas are a little bit incorrect here (we need multipleOf 0.001) for consistency with three decimal place currencies. This will still create a decimal with the correct scale right?

If so I'm happy to update.

istreeter:
I confirm multipleOf 0.001 will still create a decimal field.  So that sounds like a good solution.

miike:
Thanks for confirming @istreeter.

I've added `multipleOf` for fields where it feels required, primarily for ecommerce. I left virtual currencies as is because for invented currencies (or things like crypto) we don't know what scale is necessarily being used.

@greg-el Do you mind giving this another once over and then we can hopefully get it merged? Thanks.

schemas/com.google.tag-manager.server-side/post_score/jsonschema/1-0-0
```diff
@@ -0,0 +1,28 @@
...
+                "name": "post_score",
+                "version": "1-0-0",
+                "format": "jsonschema"
+        },
+        "description": "Send this event when the user posts a score. Use this event to understand how users are performing in your game and correlate high scores with audiences or behaviors.",
+        "type": "object",
+        "properties": {
+                "score": {
+                        "description": "The score to post.",
+                        "type": ["number", "null"]
```
AlexBenny, line 14: Are we okay to keep it nullable even if it's a required field?
(for ref: https://developers.google.com/analytics/devguides/collection/protocol/ga4/reference/events#post_score)

schemas/com.google.tag-manager.server-side/spend_virtual_currency/jsonschema/1-0-0
```diff
@@ -0,0 +1,30 @@
...
+                        "type": ["number", "null"],
+                        "minimum": 0
+                },
+                "item_name": {
+                        "description": "The name of the item the virtual currency is being used for.",
+                        "type": ["string", "null"],
+                        "maxLength": 500
+                }
+        },
+        "required": ["virtual_currency_name", "value"],
```
AlexBenny, line 28: Same here. Is it ok to keep them nullable even if required?
Ref: https://developers.google.com/analytics/devguides/collection/protocol/ga4/reference/events#spend_virtual_currency

schemas/com.google.tag-manager.server-side/unlock_achievement/jsonschema/1-0-0
```diff
@@ -0,0 +1,20 @@
...
+        "description": "Log this event when the user has unlocked an achievement. This event can help you understand how users are experiencing your game.",
+        "type": "object",
+        "properties": {
+                "achievement_id": {
+                        "description": "The id of the achievement that was unlocked.",
+                        "type": ["string", "null"],
+                        "maxLength": 500
+                }
+        },
+        "required": ["achievement_id"],
```
AlexBenny, line 18: Required but nullable. Is it ok?

AlexBenny (commented):
@miike @greg-el  can you quickly check the comments I raised? 
In a few places the field is nullable but also required.
They are technically valid but ambiguous. I'd like to check with you if this is by design or a mistake.

schemas/com.google.tag-manager.server-side/post_score/jsonschema/1-0-0
miike: Yeah, null is fine here.

schemas/com.google.tag-manager.server-side/spend_virtual_currency/jsonschema/1-0-0
miike: Yeah no issue, this just means you can send null in here if you want to.

schemas/com.google.tag-manager.server-side/unlock_achievement/jsonschema/1-0-0
miike: Yep.

istreeter (approved):

PR #1388, opened 2024-02-14
Incomplete events schema

colmsnowplow:


schemas/com.snowplowanalytics.snowplow/failure/jsonschema/1-0-0
```diff
@@ -0,0 +1,56 @@
...
+    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "self": {
+      "vendor": "com.snowplowanalytics.snowplow",
+      "name": "failure",
+      "format": "jsonschema",
+      "version": "1-0-0"
+    },
+    "type": "object",
+    "properties": {
+        "source": {
```
colmsnowplow, line 11: https://www.notion.so/Failure-Entity-Schema-Design-6361207b9801468ca8ac2bbf448c81f0?d=ad33201aefb0406b90dd6a916757ae21&pvs=4#29ce10ca04ad4648912dd079d7020768

cc @stanch

schemas/com.snowplowanalytics.snowplow/failure/jsonschema/1-0-0
```diff
@@ -0,0 +1,49 @@
...
+            "type": ["string", "null"],
+            "description": "Source of the error. If the failure happened during enrichment, it'll be the enriched field - for example `user_ipaddress`, `unstruct` or `contexts`"
+        },
+        "failureType": {
+            "type": "string"
+        },
+        "errors": {
+            "type": "array",
+            "items": {
+                "type": "object"
```
colmsnowplow, line 21: https://www.notion.so/Failure-Entity-Schema-Design-6361207b9801468ca8ac2bbf448c81f0?d=9baa41a80d8e47ed8237c98ac9346cf7&pvs=4#29ce10ca04ad4648912dd079d7020768

schemas/com.snowplowanalytics.snowplow/failure/jsonschema/1-0-0
```diff
@@ -0,0 +1,49 @@
...
+        "schema": {
+					"type": ["string", "null"],
+					"description": "The schema for the object which caused the failure, if it was self-descrbing."
+				},
+        "data": {
+            "type": ["object", "null"],
+            "additionalProperties": true,
+            "description": "The original data object which caused the failure."
+        },
+        "timestamp": {
```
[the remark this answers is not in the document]
colmsnowplow: Yes, but importantly we create one failure entity per failure. So if you have a schema violation, then a sql enrichment failure, then an api failure - then you will get 3 failure entities.

Each of those _might_ have more than one error - which is why that field is an array. For example if more than one field is invalid, there will be more than one error.

schemas/com.snowplowanalytics.snowplow/failure/jsonschema/1-0-0
```diff
@@ -0,0 +1,49 @@
...
+      "version": "1-0-0"
+    },
+    "type": "object",
+    "properties": {
+        "source": {
+            "type": ["string", "null"],
+            "description": "Source of the error. If the failure happened during enrichment, it'll be the enriched field - for example `user_ipaddress`, `unstruct` or `contexts`"
+        },
+        "failureType": {
+            "type": "string"
```
oguzhanunlu, line 16: `enum` for safer usage, wdyt @colmsnowplow ?

colmsnowplow: No objection in principal, but need to validate that changing the enum in future won't result in any problems.

miike: Enum can still be a little awkward with Redshift as I don't believe we bump the maxLength as of yet.

schemas/com.snowplowanalytics.snowplow/failure/jsonschema/1-0-0
```diff
@@ -0,0 +1,49 @@
+{
+    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "self": {
+      "vendor": "com.snowplowanalytics.snowplow",
+      "name": "failure",
+      "format": "jsonschema",
+      "version": "1-0-0"
+    },
+    "type": "object",
+    "properties": {
```
miike, line 10: I would add something like an `error_code`. At the moment we put a lot of description elements into the failed event schemas but these aren't apparent as to what the failure relates to unless you actually add it into the schema. The way I've been thinking about this is to provide a high level classification e.g.,

D - the failure is related to data that has been sent with the event
S - the failure is related to an issue with a schema sent with the event
I - the failure is related to an internal processing error (e.g., pipeline issue)
E - the failure is related to an external processing issue (e.g., failed to get a response from an API enrichment).

schemas/com.snowplowanalytics.snowplow/failure/jsonschema/1-0-0
```diff
@@ -0,0 +1,49 @@
...
+            "description": "Timestamp at which the failure occurred",
+            "format": "date-time"
+        },
+        "producingAppName": {
+            "type": "string",
+            "description": "Name of the app which produced the failure"
+        },
+        "producingAppVersion": {
+            "type": "string",
+            "description": "Version of the app which produced the failure"
```
rlh1994, line 45: I'm confused, are these trackers? App_ids? Or the part of the pipeline?

colmsnowplow: Component of the pipeline

colmsnowplow: I see the confusion though. Worth considering alternative names I guess. Any suggestions?

schemas/com.snowplowanalytics.snowplow/failure/jsonschema/1-0-0
colmsnowplow: Thanks Mike. Interesting suggestion. @stanch would love to hear what you think.

For incomplete events I'd be inclined to treat it as post-MVP, but depending on how difficult to implement it might be something we could do before rolling out the end product

schemas/com.snowplowanalytics.snowplow/failure/jsonschema/1-0-0
miike: `component` / `component_version`? I think we call this artifact at the moment but component is likely more clear.

schemas/com.snowplowanalytics.snowplow/failure/jsonschema/1-0-0
```diff
@@ -0,0 +1,49 @@
...
+        }
+    },
+    "required": ["errors", "timestamp", "producingAppName", "producingAppVersion"]
```
colmsnowplow, line 48: Add failureType here

schemas/com.snowplowanalytics.snowplow/failure/jsonschema/1-0-0
colmsnowplow: @miike now that we have some firmer examples of what the actual data looks like, we tweaked some things. I think something like the requirement you describe will be satisfied by the failureType field.

This will essentially indicate the classification of the error: eg. schema violation or enrichment error, for example.

(additionally we're releasing 4.2.0 before this one, which fixes things so that schema violations at enrichment level don't get reported as enrichment errors.)

I'm not sure if this would give you everything you're asking for here, so we can still consider it post-MVP

colmsnowplow:
The most recent commits, following from a meeting to discuss design after seeing example failure data:

- Define failureType as the classification of the failure, rather than anything to do with where the failure came up (eg. schema violation)
- removes source - source is to indicate the field which caused the issue, but this is actually specific to each error rather than general for the whole failure. For example one can have a failure where there is an error for page_url as well as page_referrer.
- Renames unclear fields
- Makes failureType required.

colmsnowplow:
The [schema is merged](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/failure/jsonschema/1-0-0), closing

PR #1395, opened 2024-03-27
Release R156

AlexBenny:


igneel64:
Can we maybe wait for https://github.com/snowplow/iglu-central/pull/1397 ? 😄

istreeter (commented):
I think this is safe.  Before this PR, `sender` was an optional field, i.e. it is not listed in the schema's `required` fields.  This means in all warehouses the column is already nullable.

But before this PR, if `sender` is present in the object then it MUST be a string, and not null.  Whereas after this PR, we allow `sender` to be an explicit null.

So... it seems we have discovered that Mandrill is liable to write explicit `null` into the payload -- is that correct?  In that case, I would be tempted to make that same change for **all** non-required fields in this schema.  It doesn't make much difference to how the data gets loaded to the warehouse.  But it might prevent future events from getting rejected for having nulls in other places.

igneel64:
@AlexBenny Can you cherry pick this one since it has been approved https://github.com/snowplow/iglu-central/pull/1397  ?
Thanks folks

jbeemster (approved):

istreeter (approved):

igneel64 (approved):

PR #1399, opened 2024-03-28
Add com.snowplowanalytics.snowplow/event_specification/jsonschema/1-0-2

### LINKED ISSUE #1398: Add com.snowplowanalytics.snowplow/event_specification/jsonschema/1-0-2
Add `data_product_id` and `data_product_name` to the event specification context.

igneel64:
(close #1398)


schemas/com.snowplowanalytics.snowplow/event_specification/jsonschema/1-0-2
```diff
@@ -0,0 +1,41 @@
...
+      "description": "Name for the event specification that the event adheres to",
+      "type": "string",
+      "maxLength": 256,
+      "minLength": 1
+    },
+    "data_product_id": {
+      "description": "Identifier for the data product that the event specification belongs to",
+      "type": "string",
+      "maxLength": 256,
+      "minLength": 36
```
cksnp, line 27: I know these are UUIDs right now, but let's keep it to '1' like we've done for event spec IDs?
BTW, I'm not sure if it would break the loader, but if not I'm thinking perhaps we should make that one's max length 256 as well.

cksnp (changes requested):

schemas/com.snowplowanalytics.snowplow/event_specification/jsonschema/1-0-2
igneel64: 👍 
Will change `data_product_id` minLength to 1.
By "that one's max length 256 as well." you mean the `id` ?

cksnp: Yup, the event spec ID max length which I've originally set to 254 but have no recollection why :D

igneel64: Reflexively I would consider this a breaking change since the in some DBs `string` with maxLength is translated as CHAR(N). 

Do you think it is worth moving to 2-0-0 now ?

cksnp: Nahhh not worth it. So let's leave 254 for now and we revise when the time for 2-0-0 comes. We'll have to make those other fields required at some point I guess.

cksnp (approved):

PR #1400, opened 2024-04-02
Release/r157

jbeemster:


igneel64 (approved):

schemas/com.snowplowanalytics.snowplow/event_specification/jsonschema/1-0-2
```diff
@@ -0,0 +1,41 @@
...
+    },
+    "data_product_id": {
+      "description": "Identifier for the data product that the event specification belongs to",
+      "type": "string",
+      "maxLength": 256,
+      "minLength": 1
+    },
+    "data_product_name": {
+      "description": "Name for the data product that the event specification belongs to",
+      "type": "string",
```
matus-tomlein, line 31: Shouldn't we make this and the other properties (except for `id`) accept null since they are not required?

matus-tomlein (approved):
LGTM!

Had a nitpick about accepting null in the properties, but that was the case also in the previous versions of the schema, so not a blocker.

cksnp (approved):

schemas/com.snowplowanalytics.snowplow/event_specification/jsonschema/1-0-2
igneel64: 👍 
Done _(In general we do not expect this to happen since these contexts would be autogenerated with all the information but let's opt for correctness)_

PR #1401, opened 2024-05-15
Adding initial hyper t schemas.

misterpig:
The table [here](https://github.com/snowplow-incubator/snowplow-hyper-t?tab=readme-ov-file#events-support) lists which schema (event and entity) is currently implemented in which environment (Shopify app, DBT models and data gen app).

### DIFF
--- a/schemas/io.snowplow.ecomm/add_payment_method/jsonschema/1-0-0
+++ b/schemas/io.snowplow.ecomm/add_payment_method/jsonschema/1-0-0
@@ -1,0 +1,28 @@
+{
+    "description": "Event occurs when a payment method is added during checkout.",
+    "properties": {
+        "payment_id": {
+            "type": "string",
+            "maxLength": 255
+        },
+        "new_payment": {
+            "type": [
+                "boolean",
+                "null"
+            ],
+            "description": "If this payment ID is new or from a saved payment on the customer's account."
+        }
+    },
+    "additionalProperties": false,
+    "type": "object",
+    "self": {
+        "vendor": "io.snowplow.ecomm",
+        "name": "add_payment_method",
+        "format": "jsonschema",
+        "version": "1-0-0"
+    },
+    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "required": [
+        "payment_id"
+    ]
+}
--- a/schemas/io.snowplow.ecomm/add_payment_method/jsonschema/1-0-0
+++ b/schemas/io.snowplow.ecomm/add_payment_method/jsonschema/1-0-0
@@ -3,6 +3,7 @@
     "properties": {
         "payment_id": {
             "type": "string",
+            "description": "The `payment` entity ID for which this event is attached with.",
             "maxLength": 255
         },
         "new_payment": {
--- a/schemas/io.snowplow.ecomm/add_to_cart/jsonschema/1-0-0
+++ b/schemas/io.snowplow.ecomm/add_to_cart/jsonschema/1-0-0
@@ -13,11 +13,13 @@
         },
         "quantity": {
             "type": "integer",
+            "description": "The number of the item that is being added to the cart.",
             "minimum": 1,
             "maximum": 2147483647
         },
         "cart_id": {
             "type": "string",
+            "description": "The ID of the cart entity that the item it being added to.",
             "maxLength": 255
         },
         "source": {
--- a/schemas/io.snowplow.ecomm/add_to_cart/jsonschema/1-0-0
+++ b/schemas/io.snowplow.ecomm/add_to_cart/jsonschema/1-0-0
@@ -6,11 +6,9 @@
             "maxLength": 255
         },
         "item_type": {
-            "type": [
-                "string",
-                "null"
-            ],
-            "description": "Examples: \"product\", \"bundle\", \"gift_card\"",
+            "type": "string",
+            "enum": ["product", "bundle"],
+            "description": "The general type of the item, currently supports the two types of a singular product and a bundle of products.",
             "maxLength": 255
         },
         "quantity": {
@@ -42,6 +40,7 @@
     "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
     "required": [
         "item_id",
+        "item_type",
         "quantity",
         "cart_id"
     ]
--- a/schemas/io.snowplow.ecomm/add_to_cart/jsonschema/1-0-0
+++ b/schemas/io.snowplow.ecomm/add_to_cart/jsonschema/1-0-0
@@ -1,0 +1,48 @@
+{
+    "description": "Event occurs when a product or bundle is added to a cart.",
+    "properties": {
+        "item_id": {
+            "type": "string",
+            "maxLength": 255
+        },
+        "item_type": {
+            "type": [
+                "string",
+                "null"
+            ],
+            "description": "Examples: \"product\", \"bundle\", \"gift_card\"",
+            "maxLength": 255
+        },
+        "quantity": {
+            "type": "integer",
+            "minimum": 1,
+            "maximum": 2147483647
+        },
+        "cart_id": {
+            "type": "string",
+            "maxLength": 255
+        },
+        "source": {
+            "type": [
+                "string",
+                "null"
+            ],
+            "description": "Examples: \"catalog\", \"wishlist\"",
+            "maxLength": 255
+        }
+    },
+    "additionalProperties": false,
+    "type": "object",
+    "self": {
+        "vendor": "io.snowplow.ecomm",
+        "name": "add_to_cart",
+        "format": "jsonschema",
+        "version": "1-0-0"
+    },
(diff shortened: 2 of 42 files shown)


schemas/io.snowplow.ecomm/click_item/jsonschema/1-0-0
```diff
@@ -0,0 +1,36 @@
+{
+    "description": "Event occurrs when an item (a product or bundle) is clicked on.",
+    "properties": {
+        "item_id": {
+            "type": "string",
+            "description": "ID reference to the item (as listed in item) that was clicked on.",
+            "maxLength": 255
+        },
+        "item_type": {
+            "type": "string",
```
matus-tomlein, line 10: Should this. be a required property or should we allow it to be null?

misterpig: So after discussing with the other guys we agree that the item type field should be required.
It is also a join key used in our data models so to ensure that users will use the correct values lest we break any downstream models we've also opted for making it an enum field.

schemas/io.snowplow.ecomm/cart/jsonschema/1-0-0
```diff
@@ -0,0 +1,166 @@
...
+        "total_items": {
+            "type": [
+                "integer",
+                "null"
+            ],
+            "description": "Total number of items in the cart.",
+            "minimum": 0,
+            "maximum": 2147483647
+        },
+        "subtotal": {
```
igneel64, line 17: I feel the `subtotal`, `shipping_cost` and `tax` belong in the `/order/` schema. Am I right ?

schemas/io.snowplow.ecomm/return_search/jsonschema/1-0-0
```diff
@@ -0,0 +1,28 @@
+{
+    "description": "When the submit_search is completed, this event will be triggered along with the contexts view_collection (the viewing of the search results) and search (the search event parameters).",
```
igneel64, line 2: Tempted to name it `search_result`.

schemas/io.snowplow.ecomm/collection/jsonschema/1-0-0
```diff
@@ -0,0 +1,154 @@
...
+            ],
+            "description": "The field that is being sorted on.",
+            "maxLength": 255
+        },
+        "items_per_page": {
+            "type": [
+                "integer",
+                "null"
+            ],
+            "description": "The total number items to be displayed on each page.",
```
igneel64, line 73: > The total number items to be displayed on each page.

Maybe

> The total number of items to be displayed on each page.

schemas/io.snowplow.ecomm/click_item/jsonschema/1-0-0
```diff
@@ -0,0 +1,38 @@
...
+            "enum": ["product", "bundle"],
+            "description": "The general type of the item, currently supports the two types of a singular product and a bundle of products.",
+            "maxLength": 255
+        },
+        "click_position": {
+            "type": [
+                "integer",
+                "null"
+            ],
+            "description": "The numerical position of the item on the page, if the item appears first, click_position would be 1",
```
igneel64, line 20: I suppose that the position is based on the concept of a 'collection'. This means you can have multiple collections of products in the same page and all of them have a product with a position of 1.
E.g.
image


In the case we are talking about a product list type of page this can be based on 'the items on the page' because you have a single collection.

Is my thinking correct here ?

schemas/io.snowplow.ecomm/click_item/jsonschema/1-0-0
```diff
@@ -0,0 +1,38 @@
...
+            "description": "ID reference to the item (as listed in item) that was clicked on.",
+            "maxLength": 255
+        },
+        "item_type": {
+            "type": "string",
+            "enum": ["product", "bundle"],
+            "description": "The general type of the item, currently supports the two types of a singular product and a bundle of products.",
+            "maxLength": 255
+        },
+        "click_position": {
```
igneel64, line 15: I would be tempted to name it `position` to make it more simlar to other ecommerce implementations. That way feels more familiar.

schemas/io.snowplow.ecomm/cart/jsonschema/1-0-0
```diff
@@ -0,0 +1,166 @@
...
+                "array",
+                "null"
+            ],
+            "description": "IDs of coupon entities that have been attached to the cart.",
+            "items": {
+                "type": "string",
+                "maxLength": 255
+            }
+        },
+        "items_in_cart": {
```
igneel64, line 79: You can use a local $ref for both items_in_cart and discovery. Might make this a bit more readable.

schemas/io.snowplow.ecomm/cancel_subscription/jsonschema/1-0-0
```diff
@@ -0,0 +1,29 @@
...
+        "subscription_id": {
+            "type": "string",
+            "maxLength": 255
+        },
+        "reason": {
+            "type": [
+                "string",
+                "null"
+            ],
+            "description": "The reason if given in notes for why the subscription was cancelled.",
```
igneel64, line 13: > The reason if given in notes for why the subscription was cancelled.

Maybe

> Reason why the subscription was cancelled.

schemas/io.snowplow.ecomm/add_payment_method/jsonschema/1-0-0
```diff
@@ -0,0 +1,28 @@
+{
+    "description": "Event occurs when a payment method is added during checkout.",
+    "properties": {
+        "payment_id": {
```
igneel64, line 4: Here I would be tempted to change this to `payment_type_id` . Since `payment` might be confused with the whole process of paying so someone might wrongly add something like the transaction Id or checkout Id.

Otherwise we can also add a description here and explain that if an 'id' is not available, they can use a friendly name for the payment method.

schemas/io.snowplow.ecomm/cart/jsonschema/1-0-0
misterpig: We think for purpose of analysing behaviour with events before an order is created it'll be helpful to have these fields with the cart entity, so worth to leave them in here as well as on the order/payment entities.

schemas/io.snowplow.ecomm/return_search/jsonschema/1-0-0
misterpig: Since we have submit_search probably best to leave that naming convention as is. Also the name `search_result` denotes that the search returned with some result but it is possible to have searches that have no results.

schemas/io.snowplow.ecomm/add_payment_method/jsonschema/1-0-0
misterpig: I'll add an description to state that the ID here is for the payment entity that is attached to the same event. Since this isn't a payment type per se we should leave this field's naming with the convention we have everywhere else, to denote that ID points to the entities (the entities' name will form the field ID's name).

schemas/io.snowplow.ecomm/click_item/jsonschema/1-0-0
misterpig: You're right, I'll add a `collection_id` field to this schema and have it to be required, so users will know where to find the specific collection from which the item that is clicked is in.

schemas/io.snowplow.ecomm/click_item/jsonschema/1-0-0
matus-tomlein: That makes sense! I just meant that if it is required then we should have it in the list of `required` properties (currently there is only `item_id`)? Otherwise it's in a weird state where we don't accept null for it but it can be missing from payload.

schemas/io.snowplow.ecomm/update_customer/jsonschema/1-0-0
```diff
@@ -0,0 +1,66 @@
...
+    "self": {
+        "vendor": "io.snowplow.ecomm",
+        "name": "update_customer",
+        "format": "jsonschema",
+        "version": "1-0-0"
+    },
+    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "required": [
+        "customer_id",
+        "change_trigger",
```
matus-tomlein, line 63: This property is required but accepts null, maybe it should not be required then?

matus-tomlein (approved):
A couple more comments, but otherwise LGTM! This is a huge piece of work! 👏

schemas/io.snowplow.ecomm/add_to_cart/jsonschema/1-0-0
```diff
@@ -0,0 +1,48 @@
...
+        "item_type": {
+            "type": [
+                "string",
+                "null"
+            ],
+            "description": "Examples: \"product\", \"bundle\", \"gift_card\"",
+            "maxLength": 255
+        },
+        "quantity": {
+            "type": "integer",
```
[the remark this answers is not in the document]
misterpig: Will update with more descriptions 🫡

igneel64 (approved):
LGTM, excellent effort 🚀

PR #1408, opened 2024-07-12
Add com.snowplowanalytics.snowplow/custom_event/jsonschema/1-0-0 

### LINKED ISSUE #1407: Add com.snowplowanalytics.snowplow/custom_event/jsonschema/1-0-0

igneel64:
Add com.snowplowanalytics.snowplow/custom_event/jsonschema/1-0-0 

close #1407


schemas/com.snowplowanalytics.snowplow/custom_event/jsonschema/1-0-0
```diff
@@ -0,0 +1,35 @@
...
+        "vendor": "com.snowplowanalytics.snowplow",
+        "name": "custom_event",
+        "format": "jsonschema",
+        "version": "1-0-0"
+    },
+    "type": "object",
+    "properties": {
+        "category": {
+            "type": ["string", "null"],
+            "maxLength": 1024,
```
gleb-lobov, line 14: 1024 feels a bit arbitrary and small, but I'm not sure if it's some sort of standard size. I would probably go with 4096, but that might be because of my lack of context

gleb-lobov: Seems we have both 1024 and 4096 limits, so it's up to us to decide. I don't have any hard opinions

gleb-lobov (approved):

cksnp (approved):

PR #1411, opened 2024-07-25
Add com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-3 (close #1410)

### LINKED ISSUE #1410: Add com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-2
Patch the 1-0-2 version of the schema to make the `previousSessionId` non-required (it already is nullable).

matus-tomlein:
Issue #1410 

This PR adds the `1-0-3` version of the `client_session` entity that has only one change compared to the `1-0-2` version:  it makes the `previousSessionId` property non-required. The property is nullable, so making it non-required should not be a breaking change.

The reason for this update is a customer issue where their mobile apps in production seem to be tracking events with the entity but for some reason are omitting the null value. We are unable to reproduce and thus fix in the tracker, but we have seen this in the past a few times, so making the schema resilient to this is the only option for us.

Also making the schema supersede the `1-0-2` schema so that it is automatically applied to events tracked with the `1-0-2` entity.

It is somewhat urgent as the customer is pushing for us since they are seeing failed events in production.

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-3
+++ b/schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-3
@@ -1,0 +1,82 @@
+{
+	"$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+	"$supersedes": ["1-0-2"],
+	"description": "Schema for a client-generated user session",
+	"self": {
+		"vendor": "com.snowplowanalytics.snowplow",
+		"name": "client_session",
+		"format": "jsonschema",
+		"version": "1-0-3"
+	},
+	"type": "object",
+	"properties": {
+		"userId": {
+			"type": "string",
+			"pattern": "^[0-9a-f]{8}-([0-9a-f]{4}-){3}[0-9a-f]{12}$|^[0-9a-f]{16}$",
+			"maxLength": 36,
+			"description": "An identifier for the user of the session"
+		},
+		"sessionId": {
+			"type": "string",
+			"format": "uuid",
+			"description": "An identifier for the session"
+		},
+		"sessionIndex": {
+			"type": "integer",
+			"minimum": 0,
+			"maximum": 2147483647,
+			"description": "The index of the current session for this user"
+		},
+		"eventIndex": {
+			"type": [
+				"null",
+				"integer"
+			],
+			"minimum": 0,
+			"maximum": 2147483647,
+			"description": "Optional index of the current event in the session"
+		},
+		"previousSessionId": {
+			"type": [
+				"null",
+				"string"
+			],
+			"format": "uuid",
+			"description": "The previous session identifier for this user"
+		},
+		"storageMechanism": {
+			"type": "string",
+			"enum": [
+				"SQLITE",
+				"COOKIE_1",
+				"COOKIE_3",
+				"LOCAL_STORAGE",
+				"FLASH_LSO"
+			],
+			"description": "The mechanism that the session information has been stored on the device"
+		},
+		"firstEventId": {
+			"type": [
+				"null",
+				"string"
+			],
+			"format": "uuid",
+			"description": "The optional identifier of the first event for this session"
+		},
+		"firstEventTimestamp": {
+			"description": "Optional date-time timestamp of when the first event in the session was tracked",
+			"type": [
+				"null",
+				"string"
+			],
+			"format": "date-time"
+		}
+	},
+	"required": [
+		"userId",
+		"sessionId",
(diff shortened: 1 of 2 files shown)


jbeemster:
Waiting for green-ticks @matus-tomlein and I can then merge it in.

mscwilson (approved):
LGTM

schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-3
```diff
@@ -0,0 +1,82 @@
+{
+	"$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+	"$supersedes": ["1-0-2"],
```
istreeter, line 3: First time we've done this in Iglu Central!  Which is exciting and terrifying.

Have you checked the dev registry created by this PR?

```
http://iglucentral-dev.com.s3-website-us-east-1.amazonaws.com/BRANCH_NAME
```

It would be great if you could confirm:

- If you fetch the 1-0-3 schema from the dev registry then it includes the `$supersedes` field
- If you fetch the 1-0-2 schema from the dev registry then it includes the `$supersededBy` field.

The dev registry is built in the same way as the main iglu-central registry.  If it works on the dev registry then it will also work in Iglu Central when this PR gets merged.

matus-tomlein: Good point! Altough the 1-0-3 schema contains the `$supersedes` field, it seems that the `$supersededBy` field is not added to the 1-0-2 schema.

Here are links to the schemas:

* http://iglucentral-dev.com.s3-website-us-east-1.amazonaws.com/issue/1410-update_client_sesison/schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-2
* http://iglucentral-dev.com.s3-website-us-east-1.amazonaws.com/issue/1410-update_client_sesison/schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-3

I will check with Nick if he has any idea why.

matus-tomlein: So as discussed with Nick on Slack, I have removed the `$supersedes` relation from the `1-0-3` schema and added a `$supersededBy` relation to the `1-0-2` schema. Have tested this by pointing a Micro to the dev Iglu Server and it has worked – validated using the new schema.

istreeter: Hi @matus-tomlein sorry this is not quite right still.  You need the `$supersedes` relation in `1-0-3` AND the `$supersededBy` relation in `1-0-2`.  Sorry to disagree with what you got told in Slack!!

The reason is... for Iglu Central we push schemas to two different types of registry.  Two of them are static registries in S3:

- `http://iglucentral.com`
- `http://mirror01.iglucentral.com`

And two of them are Iglu Servers:

- `https://com-iglucentral-eu1-prod.iglu.snplow.net`
- `https://com-iglucentral-eu2-prod.iglu.snplow.net`

For the Iglu Servers it is important that the newer schema contains the `$supersedes` field or else it drops the information.

I just tested these schemas on a local Iglu Server, and I'm fairly sure everything will be ok with pushing these schemas.  However, we should definitely flag to Josh that there is a small element of risk because it's the first time we've done this with the Iglu Central Iglu Servers.

matus-tomlein: Thanks a lot for validating that, @istreeter! I have added the `$supersedes` relation to the `1-0-3` schema back in.

istreeter (approved):

matus-tomlein:
@jbeemster can you please merge it in when you have the chance (also happy to put together a release PR, just let me know)?

As Ian noted, this is the first time that we add schemas with a supersedes relationship to Iglu Central, so we should be a bit more careful here. Also the `client_session` schema is a very popular one, almost all mobile apps track it.

schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-3
istreeter: Well done for battling through this :)  If this works out well, then it sets a nice pattern for future PRs.  Hopefully it won't be so difficult next time!

jbeemster:
@matus-tomlein can you please prepare the release PR and ill take a look and merge.  Going to need to enable patches on our Iglu Central server mirrors as well to allow for this change.

matus-tomlein:
Cherry picked into release #1412

PR #1412, opened 2024-07-29
Release/r161

matus-tomlein:
This release contains the schema for the `client_session` entity version 1-0-3 that supersedes the 1-0-2 version. It adds the `$supersededBy` relation to the 1-0-2 version and the `supersedes` relation to the 1-0-3 version. The change in the 1-0-3 schema is that the `previousSessionId` property is no longer required (not a breaking change since it was nullable).

**Changes**
* Add com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-3 (#1410)
* Patch com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-2 (#1410)

Needs extra attention as this is the first time that we use the supersedes relation in Iglu Central and the `client_session` schema is a popular one.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+Release 161 (2024-07-29)
+------------------------
+Patch com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-2 (#1410)
+
 Release 160 (2024-07-17)
 ------------------------
 Add com.snowplowanalytics.snowplow/custom_event/jsonschema/1-0-0 (close #1407)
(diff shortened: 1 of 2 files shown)


jbeemster:
@matus-tomlein @istreeter silly question here but if we are going to need to patch `1-0-2` regardless with the `$supersededBy` relation is it not simpler to just remove the required condition on `1-0-2` as part of the patch and not introduce any superseded logic anywhere (nor `1-0-3`).

Why do we prefer this approach to just patching `1-0-2`?

matus-tomlein:
@jbeemster Happy to go with just patching `1-0-2`, I don't have a strong preference. Originally I thought we could just get by with the `supersedes` relation in the `1-0-3` schema, but it's true that since we are patching, we can just patch the `1-0-2`.

jbeemster:
@istreeter @matus-tomlein assuming the patch is safe lets do that rather than make a much bigger amount of work out of all of this.

matus-tomlein:
Sounds good! Have just updated the PR to only have the 1-0-2 patch.

jbeemster (approved):

istreeter (approved):

PR #1419, opened 2024-11-08
Add com.snowplowanalytics.snowplow/application/jsonschema/1-0-0 (close #1418)

### LINKED ISSUE #1418: Add com.snowplowanalytics.snowplow/application/jsonschema/1-0-0
Add a schema for specifying the app version that will be tracked from Web trackers.

matus-tomlein:
Add a schema for the application context entity tracked in Web and other apps that specifies the version number of the application. This can be in the form of semver or a Git commit SHA, so the schema doesn't put any further limits on it other than the max length.

Note that [we already have a `application` context entity](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.mobile/application/jsonschema/1-0-0) that is tracked on mobile – it has the vendor `com.snowplowanalytics.mobile`. In addition to the version, it also has a required `build` property that is available in mobile app builds. We can't reuse the same as the `build` number is not so useful on Web apps.


johnmicahreid:
LGTM

johnmicahreid (approved):
LGTM

PR #1421, opened 2024-12-11
Element Tracking Plugin Schemas

jethron:
This PR adds schemas for a new [Element Tracking plugin](https://github.com/snowplow/snowplow-javascript-tracker/pull/1400) for the JavaScript Tracker.

It contains 4 Self Describing Events:
- `create_element`
- `destroy_element`
- `expose_element`
- `obscure_element`

And 4 entities:
- `element`
- `element_content`
- `element_statistics`
- `component_parents`

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/element_statistics/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow/element_statistics/jsonschema/1-0-0
@@ -1,0 +1,86 @@
+{
+  "description": "Entity detailing aggregate statistics for a named matched using the Element Tracking plugin.",
+  "properties": {
+    "element_name": {
+      "description": "The name of the element configuration that this entity is detailing.",
+      "type": "string",
+      "maxLength": 255
+    },
+    "element_index": {
+      "description": "The (1-based) index of the element amongst the `element_matches` elements that match its same configuration.",
+      "type": "number",
+      "minimum": 1
+    },
+    "element_matches": {
+      "description": "The count of the elements that matched the configuration named `element_name`.",
+      "type": "number",
+      "minimum": 0
+    },
+    "current_state": {
+      "description": "The visibility state of the element at the time of this event.",
+      "type": "string",
+      "maxLength": 40
+    },
+    "min_size": {
+      "description": "The smallest dimensions observed of the named element, in virtual pixels. (width x height)",
+      "type": "string",
+      "pattern": "^[0-9.]+x[0-9.]+$",
+      "maxLength": 40
+    },
+    "current_size": {
+      "description": "The current dimensions observed of the named element, in virtual pixels. (width x height)",
+      "type": "string",
+      "pattern": "^[0-9.]+x[0-9.]+$",
+      "maxLength": 40
+    },
+    "max_size": {
+      "description": "The largest dimensions observed of the named element, in virtual pixels. (width x height)",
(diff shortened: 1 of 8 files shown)


schemas/com.snowplowanalytics.snowplow/element_statistics/jsonschema/1-0-0
```diff
@@ -0,0 +1,86 @@
...
+      "pattern": "^[0-9.]+x[0-9.]+$",
+      "maxLength": 40
+    },
+    "y_depth_percentage": {
+      "description": "The maximum vertical position seen of the element at the time of the event, expressed as a percentage of its height; from 0.0 - 1.0 (0% - 100%).",
+      "type": ["number", "null"],
+      "minimum": 0.0,
+      "maximum": 1.0
+    },
+    "max_y_depth_percentage": {
```
matus-tomlein, line 48: Nit: the `_percentage` suffix is a bit confusing given that the range is from 0 to 1 (would expect 0 to 100). Maybe call it ratio or change the scale?

schemas/com.snowplowanalytics.snowplow/element_statistics/jsonschema/1-0-0
```diff
@@ -0,0 +1,86 @@
...
+      "minimum": 0
+    },
+    "times_in_view": {
+      "description": "The amount of times this element has appeared in view.",
+      "type": "integer",
+      "minimum": 0
+    },
+    "total_time_visible_ms": {
+      "description": "The cumulative amount of time (in milliseconds) this element has been in view so far.",
+      "type": "number",
```
matus-tomlein, line 72: Probably we can go for `integer` given this is in ms already?

matus-tomlein (approved):
LGTM!

schemas/com.snowplowanalytics.snowplow/element_statistics/jsonschema/1-0-0
```diff
@@ -0,0 +1,86 @@
...
+    },
+    "max_y_depth": {
+      "description": "The maximum vertical position seen of the element so far, as a fraction.",
+      "type": "string",
+      "pattern": "^[0-9.]+/[0-9.]+$",
+      "maxLength": 40
+    },
+    "element_age_ms": {
+      "description": "The amount of time (in milliseconds) this element has existed on page since it was first observed.",
+      "type": "integer",
```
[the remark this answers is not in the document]
istreeter: For Redshift, looks like [this is the relevant line](https://github.com/snowplow/schema-ddl/blob/0.27.0/modules/core/src/main/scala/com.snowplowanalytics/iglu.schemaddl/redshift/internal/ColumnTypeSuggestions.scala#L111).  It's a BigInt (64-bit integer).

For Databricks/BigQuery/Lakes, [this is the relevant line](https://github.com/snowplow/schema-ddl/blob/0.27.0/modules/core/src/main/scala/com.snowplowanalytics/iglu.schemaddl/parquet/Suggestion.scala#L159), which similarly means 64-bit integer.

If you want it to be stored as a 32-bit integer, then you could set maximum size to `2147483647`.  It's a trick used in many other iglu-central schemas, e.g. [this one](https://github.com/snowplow/iglu-central/blob/master/schemas/io.snowplow.ecomm/add_to_cart/jsonschema/1-0-0#L17-L18)

PR #1426, opened 2025-02-25
Add nl.basjes/yauaa_context/jsonschema/1-0-5

oguzhanunlu:
This PR adds `yauaa_context/1-0-5` by converting enums to strings with a future-proof maxLength(128). We don’t have control on those enums and any new version of yauaa lib can add and remove items.

**Why not bump to `2-0-0`?**
- Please read the comments starting [here](https://github.com/snowplow/iglu-central/pull/1426#issuecomment-2697459988) first
- As a team we considered that it is not worth to require updates in downstream data models or other apps just to load `Desktop App` in _2 table instead of `Desktop Ap` in _1 table.

```bash
$ diff schemas/nl.basjes/yauaa_context/jsonschema/1-0-4 schemas/nl.basjes/yauaa_context/jsonschema/1-0-5
8c8
<         "version": "1-0-4"
---
>         "version": "1-0-5"
14c14,15
<             "enum": ["Desktop", "Anonymized", "Unknown", "UNKNOWN", "Mobile", "Tablet", "Phone", "Watch", "Virtual Reality", "eReader", "Set-top box", "TV", "Game Console", "Home Appliance", "Handheld Game Console", "Voice", "Car", "Robot", "Robot Mobile", "Spy", "Hacker", "Augmented Reality", "Robot Imitator"]
---
>             "type": "string",
>             "maxLength": 128
45c46
<             "enum": ["Desktop", "Mobile", "Cloud", "Embedded", "Game Console", "Hacker", "Anonymized", "Unknown"]
---
>             "maxLength": 128
67c68
<             "enum": ["Browser", "Mobile App", "Hacker", "Robot", "Unknown", "Special", "Cloud", "eReader"]
---
>             "maxLength": 128
96c97
<             "enum": ["Browser", "Browser Webview", "Mobile App", "Robot", "Robot Mobile", "Cloud Application", "Email Client", "Voice", "Special", "Testclient", "Hacker", "Unknown", "Desktop App", "eReader"]
---
>             "maxLength": 128
139c140
<             "enum": ["Weak security", "Strong security", "Unknown", "Hacker", "No security"]
---
>             "maxLength": 128
153a155,158
>         "webviewAppNameVersion": {
>             "type": "string",
>             "maxLength": 1000
>         },
```

ref: PDP-1090


istreeter (commented):
My only doubt is whether `32` is long enough to future-proof the new fields.

In #1216 we got bored of forever bumping max lengths, whenever the yauaa library changed, or whenever a new user agent got invented.  So we deliberately bumped all max lengths to 128 or more, to better future-proof the schema.

istreeter (approved):
Looks good from my point of view.

@oguzhanunlu please can you remind me why this is 2-0-0, not 1-0-5?  I'm sure you're probably right, but I've forgotten the reason.

Even better... you could add the explanation to the PR description, so we have a long term reference of why it is 2-0-0.

How urgently does this need to be merged?  Presumably you can you use the dev registry while working on the feature in Enrich?

oguzhanunlu:
Thanks @istreeter ! I just updated PR description with a short note on why it is `2-0-0` from my point of view. There is no urgency as I can reference my branch when needed. Let me add that this is a blocker for the next enrich release which is coming in the new few weeks.

oguzhanunlu:
thanks @istreeter ! my bad, I was asking the wrong question. The blog post asks `Will this new schema validate all historical data?`.  Here all historical data will be valid against new schema as the new field isn't required and additionalProperties is false, so it indicates an addition on its own.

>  It's because you changed several fields from enum to regular string.  If there is any chance that might break any loader, then this should be a 2-0-0 schema.

Asking the same question. Since previous enum possibilites are always string with length <128, it looks like new schema can validate all historical data. So it indicates an addition as well. However I'm not sure if type conversion could break some of the loaders, I need to check them.

oguzhanunlu:
For redshift,
- For an enum field with text255 encoding, enum -> string max(128) results in values loaded partially, as wide as current varchar size, e.g. for deviceClass field, it is varchar(21) and even if new schema allows up to 128 chars, only first 21 chars will be loaded

- For an enum field with ztsd encoding, enum -> string max(128) results in values loaded completely, varchar size is increased

Both bullets make me think that we don't need a model bump and we can go for `1-0-5`.

istreeter:
Thanks @oguzhanunlu.  Is there any _known_ values which would get partially loaded not fully loaded?  With the newest version of the yauaa library.

For example... you mentioned `deviceClass`.  I guess we have lots of legacy Redshift users whose column is `varchar(21)`.  Is there any _known_ value of `deviceClass` which is longer than 21 characters?

Same question for the other enum fields.

I think it changes the decision.  If there are values that would get partially loaded, then we might choose to make it a `2-0-0` schema.  A `2-0-0` upgrade is inconvenient for users, but at least it means all users get fully loaded data.

On the other hand, if there are no known values that get partially loaded, then we might as well make it a `1-0-5` schema.

oguzhanunlu:
Thanks @istreeter! Among 5 enum fields, only `layoutEngineClass` has a minor issue. Other 4 enum fields don't have a new value longer then the current field limit.

Based on [7.11.0](https://github.com/nielsbasjes/yauaa/blob/v7.11.0/documentation/content/expect/FieldValues.md) and [7.30.0](https://github.com/nielsbasjes/yauaa/blob/v7.30.0/documentation/content/expect/FieldValues.md) pages,  `layoutEngineClass` should be `varchar(10)` due to length of `Mobile App` for our customers, however yauaa 7.30.0 has `Desktop App` which has 11 chars. If we pick 1-0-5, then `Desktop Ap` will be loaded. Is it worth breaking downstream modeling at the expense of loading the last `p`? Personally I don't think so and would go for `1-0-5`, what do you think @istreeter ?

istreeter (approved):

jbeemster:
@istreeter @oguzhanunlu this has been merged right? Should this PR be closed?

oguzhanunlu:
that's correct @jbeemster , closing

PR #1428, opened 2025-02-28
Add com.snowplowanalytics.snowplow/button_click/jsonschema/2-0-0

Jack-Keene:
This PR makes the `label` property of the `button_click` schema an optional property. This is to handle buttons without any inner text - for example icon buttons.

I've also added the $supersedes property for the previous v1-0-0 schema

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/button_click/jsonschema/1-0-1
+++ b/schemas/com.snowplowanalytics.snowplow/button_click/jsonschema/1-0-1
@@ -1,0 +1,53 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for a button click event",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow",
+    "name": "button_click",
+    "format": "jsonschema",
+    "version": "1-0-1"
+  },
+  "type": "object",
+  "properties": {
+    "label": {
+      "type": [
+        "string",
+        "null"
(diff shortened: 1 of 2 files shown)


istreeter:
Hi @Jack-Keene regarding `$supersedes`, please refer to [this conversation in a previous PR](https://github.com/snowplow/iglu-central/pull/1411#discussion_r1692884594).  We concluded that in Iglu Central you need to add `$supersedes` to the new schema AND `$supersededBy` to the old schema.

After you've read that, then please refer to [this follow-up conversation](https://github.com/snowplow/iglu-central/pull/1412#issuecomment-2256027459) where it got decided simply to remove the `required` from the previous schema, instead of releasing a new schema.

I think the situation for `button_click` is the same as the situation for `client_session` which got discussed in the earlier conversations.

matus-tomlein:
Sorry missed this PR. Agree with @istreeter, but just want to add that when in the `client_session` case we replaced the schema, Lego complained that it "broke a bunch of events for them". I don't really understand how, @johnmicahreid followed up with them to get more info. But I think it might be better to avoid the replace and just release a new patch version that we update in our trackers. It's not as neat as customers will need to update their tracking, but probably safer.

johnmicahreid:
Hmmm, it's annoying that it would break Redshift, but I guess we'd better be safe than sorry. Let's release a new major version and update the trackers to use that going forward - this is resulting in failed events for many of our customers so I think it's worth the pain of a major release.

Jack-Keene:
I'm closing this as we'll make change in the tracker instead

PR #1423, opened 2024-12-15
Add io.growthbook/experiment_viewed/jsonschema/1-0-0

Auz:
Added GrowthBook experiment_viewed event tracking schema
Fixes #1422

PR #1433, opened 2025-04-28
Release R166

matus-tomlein:
This release adds schemas for the [Element Tracking plugin](https://github.com/snowplow/snowplow-javascript-tracker/pull/1400) in the JS tracker by @jethron.

The schemas has already been reviewed and approved in this PR: https://github.com/snowplow/iglu-central/pull/1421

It also adds a io.growthbook schema reviewed in https://github.com/snowplow/iglu-central/pull/1423

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/element/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow/element/jsonschema/1-0-0
@@ -1,0 +1,85 @@
+{
+  "description": "Entity detailing a specific element matched using the Element Tracking plugin.",
+  "properties": {
+    "element_name": {
+      "description": "The name of the element configuration that this entity is detailing.",
+      "type": "string",
+      "maxLength": 255
+    },
+    "originating_page_view": {
+      "description": "The active page view ID when this element was first observed. For elements that persist across multiple pageviews, this is the pageview the element originated on and may be different to the current pageview ID of this event. May be empty if the element has not actually been observed yet.",
+      "type": "string",
+      "maxLength": 36
+    },
+    "height": {
+      "description": "The height of the named element at time of event, in virtual pixels.",
+      "type": "number"
+    },
+    "width": {
+      "description": "The width of the named element at time of event, in virtual pixels.",
+      "type": "number"
+    },
+    "position_x": {
+      "description": "The offset position of the named element at time of event, in virtual pixels from the left side of the viewport.",
+      "type": "number"
+    },
+    "position_y": {
+      "description": "The offset position of the named element at time of event, in virtual pixels from the top side of the viewport.",
+      "type": "number"
+    },
+    "doc_position_x": {
+      "description": "The offset position of the named element at time of event, in virtual pixels from the left side of the document.",
+      "type": "number"
+    },
+    "doc_position_y": {
+      "description": "The offset position of the named element at time of event, in virtual pixels from the top side of the document.",
+      "type": "number"
+    },
+    "element_index": {
+      "description": "The index of the named element among other elements that match it's selector within the document. (one-based, out of `element_matches` in document)",
+      "type": "integer",
+      "minimum": 1
+    },
+    "element_matches": {
+      "description": "The number of elements that match the named element's configured selector within the document. (one-based)",
+      "type": "integer",
+      "minimum": 1
+    },
+    "attributes": {
+      "description": "List of attributes captured about the named element from its configuration.",
+      "type": ["array", "null"],
+      "items": {
+        "description": "Individual attribute information about the named element, derived from its configuration.",
+        "type": "object",
+        "required": ["source", "attribute", "value"],
+        "properties": {
+          "source": {
+            "description": "The type of configuration, and source of its value.",
+            "enum": ["callback", "content", "selector", "dataset", "attributes", "properties", "child_text", "error"],
+            "maxLength": 40
+          },
+          "attribute": {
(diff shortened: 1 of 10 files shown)


schemas/com.snowplowanalytics.snowplow/element_statistics/jsonschema/1-0-0
```diff
@@ -0,0 +1,86 @@
...
+    },
+    "max_y_depth": {
+      "description": "The maximum vertical position seen of the element so far, as a fraction.",
+      "type": "string",
+      "pattern": "^[0-9.]+/[0-9.]+$",
+      "maxLength": 40
+    },
+    "element_age_ms": {
+      "description": "The amount of time (in milliseconds) this element has existed on page since it was first observed.",
+      "type": "integer",
```
matus-tomlein, line 62: @istreeter Can I please get your take about a question that @miike raised here: https://github.com/snowplow/iglu-central/pull/1421#discussion_r2060884422

The question is what schema ddl does for Redshift in terms of int size where there is no max? Does it do something like INT2?

miike: I think int64 is probably fine, I can't really see that causing any issues in modern warehouses.

jethron (approved):
LGTM!

miike:
Happy for this to be merged.

miike (approved):

matus-tomlein:
Just to explain the last force push – I have updated the growthbook schema making the optional properties (`hashAttribute`, `hashValue` and `custom`) nullable. This was spotted by @miike.

matus-tomlein:
Sorry for pushing, but this iglu release is something that we need for a JS tracker element tracking feature that is planned for this week.

Can I please ask @istreeter if you could take a look through the schemas if there is anything blocking and @jbeemster if you could then make the release? Thank you!

schemas/com.snowplowanalytics.snowplow/element/jsonschema/1-0-0
```diff
@@ -0,0 +1,85 @@
...
+  "description": "Entity detailing a specific element matched using the Element Tracking plugin.",
+  "properties": {
+    "element_name": {
+      "description": "The name of the element configuration that this entity is detailing.",
+      "type": "string",
+      "maxLength": 255
+    },
+    "originating_page_view": {
+      "description": "The active page view ID when this element was first observed. For elements that persist across multiple pageviews, this is the pageview the element originated on and may be different to the current pageview ID of this event. May be empty if the element has not actually been observed yet.",
+      "type": "string",
```
istreeter, line 11: Other schemas in this PR allow for a null when a field is not required, e.g. `"type": ["string", "null"]`.  But this schema doesn't.

That's ok!  (And actually personally I dislike the trend of adding `null` everywhere to the type).  So I'm just pointing out the inconsistency.

schemas/com.snowplowanalytics.snowplow/element/jsonschema/1-0-0
```diff
@@ -0,0 +1,85 @@
...
+      "maxLength": 36
+    },
+    "height": {
+      "description": "The height of the named element at time of event, in virtual pixels.",
+      "type": "number"
```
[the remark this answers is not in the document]
jethron: The data reported by the browser would require `number` as virtual pixels in browsers can quite easily be fractional. We could probably round to an int if you think it's worth it, but for current implementation we need floating point. See first example event in the [Scroll sections example](https://docs.snowplow.io/docs/sources/trackers/javascript-trackers/web-tracker/tracking-events/element-tracking/#examples)

schemas/com.snowplowanalytics.snowplow/element/jsonschema/1-0-0
```diff
@@ -0,0 +1,85 @@
...
+      "minimum": 1
+    },
+    "element_matches": {
+      "description": "The number of elements that match the named element's configured selector within the document. (one-based)",
+      "type": "integer",
+      "minimum": 1
+    },
+    "attributes": {
+      "description": "List of attributes captured about the named element from its configuration.",
+      "type": ["array", "null"],
```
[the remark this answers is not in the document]
jethron: Not a mistake; the one in `component_parents` is the only property of that schema; if it's empty the entity shouldn't be attached at all; but for this one there is no requirement to have any `attributes` defined so it can be empty.

schemas/com.snowplowanalytics.snowplow/element/jsonschema/1-0-0
jethron: I think historically the "null for optional fields" pattern is mostly relevant for cross-platform schemas; because they don't have an easy way to mark a map value as "empty; don't serialize in JSON", whereas JS has `undefined` and `JSON.stringify({a: undefined}) === JSON.stringify({})`. So in other languages if an optional property has no value you have to explicitly have conditional logic to add/remove the property from a hashmap, and so as a convenience we encourage allowing `null` as a value to avoid that boilerplate.

These schemas are specifically web-centric, so will only be produced in a JS environment, so I have skipped that pattern as it is unnecessary; the fields that still accept `null` explicitly are usually because that is the result of a computation in the plugin e.g. quotients where an input can be 0, the list of attributes can be truncated.

The other schema is cross-platform I think, so it continues the pattern.

istreeter:
Thanks for answering all my questions @jethron.  This all looks good to me!

istreeter (approved):

PR #1437, opened 2025-06-19
Add schema for event forwarding error

colmsnowplow:


schemas/com.snowplowanalytics.snowplow.badrows/event_forwarding_error/jsonschema/1-0-0
```diff
@@ -0,0 +1,61 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Generic, very unstructured bad row that can be emitted by any Snowplow component",
```
istreeter: Looks like you copied the description from the `generic_error` schema.

colmsnowplow:
I guess this is why I asked for your feedback - here, I've elected for a different format, aiming at the most simple structure to query for purposes of this product's debugging flow.

But I'm not entirely confident in this - so I guess I'd like to hear opinions on this deviation.

schemas/com.snowplowanalytics.snowplow.badrows/event_forwarding_error/jsonschema/1-0-0
colmsnowplow: I did, thanks for catching. Fixed.

schemas/com.snowplowanalytics.snowplow.badrows/event_forwarding_error/jsonschema/1-0-0
```diff
@@ -0,0 +1,70 @@
...
+        "errorMessage": {
+          "type": "string",
+          "description": "The error message, if we have one"
+        },
+        "errorCode": {
+          "type": "string",
+          "description": "The error code, if we have one"
+        }
+      },
+      "required": ["timestamp", "errorType", "errorMessage"],
```
istreeter, line 64: Just pointing out `latestState` and `errorCode` are not listed in `required`.  Deliberate or accident?

istreeter (approved):
Aside from my one comment below, I think this looks good.  Thank you for your patience in changing this compared to the original submission.  I this is the right decision to stick to the design of the other bad rows, i.e

- Same three top level fields
- Consistent `processor` field across all bad rows
- Consistent `payload` field across all bad rows
- Messy chaos for the `failure` field.

schemas/com.snowplowanalytics.snowplow.badrows/event_forwarding_error/jsonschema/1-0-0
colmsnowplow: Deliberate. When we handle oversized data, `latestState` is first to be trimmed - and in the unlikely event that `payload` alone exceeds the target byte limit, `latestState` won't be provided at all.

`errorCode` - we deal with both API errors and transformation errors here. So we might not have an `errorCode`. We probably will create a code classification for those that don't have one but I'd like dealing with that to be more about the details of implementation rather than a constraint the schema forces on us.

Would be happy for scrutiny on those details though!

PR #1439, opened 2025-07-01
Add schema for event forwarding metric

pondzix:


jbeemster (commented):
Schema looks good - @pondzix can you check out other release branches and prepare one with the CHANGELOG and so on so that this can be merged please?

schemas/com.snowplowanalytics.snowplow/event_forwarding_metrics/jsonschema/1-0-0
```diff
@@ -0,0 +1,120 @@
...
+      }
+    },
+    "failedErrors": {
+      "type": "array",
+      "description": "List of aggregated transient errors with their counts",
+      "items": {
+        "type": "object",
+        "properties": {
+          "code": {
+            "type": "string",
```
istreeter, line 80: I noticed some of your string fields have a max length, whereas others (e.g. this one) do not. Just checking that's not an accident?

pondzix: I think I was inspired by 2 schemas:

* the other one for event forwarding, for the new badrow type where we don't set max length for any error fields (like code) https://github.com/snowplow/iglu-central/blob/0a08c5799144b59bec071870ba1188012e5c3da4/schemas/com.snowplowanalytics.snowplow.badrows/event_forwarding_error/jsonschema/1-0-0#L53 
* loader alert/heartbeat where we have `appVersion` and `appName` fields with limits. https://github.com/snowplow/iglu-central/blob/0a08c5799144b59bec071870ba1188012e5c3da4/schemas/com.snowplowanalytics.monitoring.loader/alert/jsonschema/1-0-0#L20

So... it's kind of an accident.

PR #1444, opened 2025-09-09
Signals tracking schemas

jethron:
Add event/entity schemas used for tracking the receipt and handling of Signals Interventions, used with the browser tracker plugin.


schemas/com.snowplowanalytics.signals/intervention_handle_error/jsonschema/1-0-0
```diff
@@ -0,0 +1,25 @@
...
+	"properties": {
+		"handler": {
+			"description": "The name of the handler that processed the intervention unsuccessfully.",
+			"type": "string",
+			"maxLength": 255
+		},
+		"error": {
+			"description": "Error text describing why the handler was unable to handle the received intervention.",
+			"type": ["string", "null"],
+			"maxLength": 255
```
matus-tomlein: Feels like this is too short for an error message? At least comparably to the handler name would make sense for this to allow longer errors.

schemas/com.snowplowanalytics.signals/intervention_instance/jsonschema/1-0-0
```diff
@@ -0,0 +1,69 @@
...
+			"properties": {
+				"name": {
+					"description": "The name that defines which attribute key was targeted.",
+					"type": "string",
+					"minLength": 1,
+					"maxLength": 128
+				},
+				"id": {
+					"description": "The actual identifier value of the attribute key that was targeted.",
+					"type": "string"
```
matus-tomlein, line 40: Not sure why, but in the SDK, the `id` has optional type.

schemas/com.snowplowanalytics.signals/intervention_instance/jsonschema/1-0-0
```diff
@@ -0,0 +1,69 @@
...
+				"type": "object",
+				"properties": {
+					"attribute": {
+						"description": "The name of the attribute group and attribute that had a value.",
+						"type": "string",
+						"minLength": 3,
+						"maxLength": 257
+					},
+					"value": {
+						"description": "The JSON-stringified version of the attribute's value.",
```
matus-tomlein, line 59: Are we actually JSON stringifying the values on the SDK? Don't see that happening.

jethron: It's enforced in the API, where the type is `dict[str, str]`, so streaming engine etc does the stringification but when pulling an intervention you should only get string values, yeah.

schemas/com.snowplowanalytics.signals/intervention_instance/jsonschema/1-0-0
jethron: Legacy, it's no longer optional it just wasn't updated.

schemas/com.snowplowanalytics.signals/intervention_instance/jsonschema/1-0-0
matus-tomlein: Ah okay, was just confused by the types in the SDK

PR #1445, opened 2025-09-09
Release/r169

jethron:
This release adds:
* schemas for tracking the receipt of Signals Interventions as reviewed in #1444.
* event forwarding metrics schema as reviewed in #1439

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,11 @@
+Release 169 (2025-09-11)
+------------------------
+Add com.snowplowanalytics.snowplow/event_forwarding_metrics/jsonschema/1-0-0
+Add com.snowplowanalytics.signals/intervention_instance/jsonschema/1-0-0
+Add com.snowplowanalytics.signals/intervention_receive/jsonschema/1-0-0
+Add com.snowplowanalytics.signals/intervention_handle/jsonschema/1-0-0
+Add com.snowplowanalytics.signals/intervention_handle_error/jsonschema/1-0-0
+
 Release 168 (2025-07-25)
 ------------------------
 Add com.snowplowanalytics.snowplow.badrows/event_forwarding_error/jsonschema/1-0-0


schemas/com.snowplowanalytics.signals/intervention_instance/jsonschema/1-0-0
```diff
@@ -0,0 +1,69 @@
...
+		"name": {
+			"description": "The name of the intervention.",
+			"type": "string",
+			"minLength": 1,
+			"maxLength": 128
+		},
+		"version": {
+			"description": "The version number of the intervention.",
+			"type": "integer",
+			"minimum": 0
```
istreeter, line 26: If there is no maximum, then some of our loaders store it as a 64-bit integer.  If you set a smaller maximum, then you can make it stored as a 32-bit integer.  Just checking you're ok with 64-bit integer?

schemas/com.snowplowanalytics.signals/intervention_instance/jsonschema/1-0-0
```diff
@@ -0,0 +1,69 @@
...
+			"properties": {
+				"name": {
+					"description": "The name that defines which attribute key was targeted.",
+					"type": "string",
+					"minLength": 1,
+					"maxLength": 128
+				},
+				"id": {
+					"description": "The actual identifier value of the attribute key that was targeted.",
+					"type": "string"
```
istreeter, line 40: In most places you set a `maxLength` for your strings, but here you did not.  Just checking this is not a mistake?

schemas/com.snowplowanalytics.signals/intervention_instance/jsonschema/1-0-0
```diff
@@ -0,0 +1,69 @@
...
+				"properties": {
+					"attribute": {
+						"description": "The name of the attribute group and attribute that had a value.",
+						"type": "string",
+						"minLength": 3,
+						"maxLength": 257
+					},
+					"value": {
+						"description": "The JSON-stringified version of the attribute's value.",
+						"type": "string"
```
istreeter, line 60: Here also you don't set `maxLength`.

jethron: One of the value types is "unique list" of strings, and we have no limit on the length of those strings (but will soon have a limit on the number of items), so this JSON is basically unbounded for the moment, so this is intentional.

schemas/com.snowplowanalytics.signals/intervention_instance/jsonschema/1-0-0
jethron: Not a mistake; the value can be any atomic-column value and in future any value from within entities/self-desc-events, so this is deliberately unknown unfortunately. We have a cap in place so there's mostly a 128-char limit, but it's not concrete. :/

schemas/com.snowplowanalytics.signals/intervention_instance/jsonschema/1-0-0
jethron: At the time there was no limit here, but it looks like we just capped it at 1_000_000, @matus-tomlein should I update to reflect that? In earlier versions the theory was that some people might want to use timestamps or something which would be >1000000 but I think we manage the numbering now?

@istreeter if needed, is it ok to rebase this branch or should it be another PR?

matus-tomlein: I wouldn't rule out the timestamp versioning case. The API validations are a bit more flexible and we can adjust those easier than changing schemas, so I would prefer to be a bit more permissive about the max here.

istreeter: Fine by me if you want to rebase, or anything else.  The schemas all look fine to me.

matus-tomlein (approved):

istreeter (approved):

PR #1476, opened 2026-03-24
Add 2-0-0 identity schemas with snake_case field names

peel:
Previously identity and identity_merge schemas used camelCase field names (snowplowId, createdAt, mergedAt, triggeringEventId) in their 1-0-0 versions.

Now 2-0-0 versions use snake_case field names (snowplow_id, created_at, merged_at, triggering_event_id) to align with Snowplow conventions. The 1-0-0 schemas remain unchanged for backward compatibility.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,8 @@
+Release 180 (2026-03-30)
+------------------------
+Add com.snowplowanalytics.snowplow/identity_merge/jsonschema/2-0-0 (#1476)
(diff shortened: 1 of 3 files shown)


istreeter (approved):
Looks good to me.  Simple renaming of fields.

I'm curious... what's the background for this change?  Iglu Central schemas use a mixture of snake case and camel case - we have always been inconsistent.  Our customers also use a mixture of cases for their own schemas.  I've never had an opinion on it.

jamessnowplow (approved):
I also have no strong opinion. Looks good to me.

PR #1477, opened 2026-03-31
Add com.snowplow.agent.tracking schemas (12 schemas)

jrpeck1989:
## Summary

- Adds 12 new schemas under `com.snowplow.agent.tracking` for agentic application tracking
- **3 entities**: `agent_context`, `tool_context`, `message_context`
- **9 events**: `agent_invocation`, `agent_step`, `agent_completion`, `tool_execution`, `message_sent`, `message_received`, `user_intent_detected`, `agent_decision_logged`, `constraint_violation`
- Domain-specific fields (`intent_category`, `constraint_type`, `tool_category`, `decision_type`) use `examples` rather than `enum` so any application can define its own values via app-level validation while the Iglu schema remains universal

## Schema inventory

| Schema | Type | Purpose |
|--------|------|---------|
| `message_context` | Entity | Chat message metadata (role, length, position) |
| `agent_context` | Entity | Agent config and state (model, provider, session) |
| `tool_context` | Entity | Tool being invoked (name, category, parameters) |
| `message_sent` | Event | User sends a chat message |
| `message_received` | Event | User receives agent response |
| `agent_invocation` | Event | Start of an agent invocation |
| `agent_step` | Event | Single iteration in agent reasoning loop |
| `agent_completion` | Event | Agent invocation completes with summary metrics |
| `tool_execution` | Event | Agent executes a tool |
| `user_intent_detected` | Event | Agent detects user intent with confidence score |
| `agent_decision_logged` | Event | Agent logs a decision with reasoning |
| `constraint_violation` | Event | User requirement cannot be met due to a constraint |

## Design decisions

- **Vendor**: `com.snowplow.agent.tracking` — follows the convention of descriptive sub-namespaces (cf. `com.snowplowanalytics.snowplow.ecommerce`, `com.snowplowanalytics.snowplow.media`)
- **Generalized enums → examples**: Four fields use `examples` + `maxLength` instead of strict `enum` to keep schemas domain-agnostic. App-level validation (e.g. Zod) enforces specific values per domain
- **All string fields** have `maxLength` constraints for warehouse compatibility
- **All integer fields** have `minimum`/`maximum` bounds

## Test plan

- [ ] Verify JSON schema validity for all 12 files
- [ ] Confirm directory structure follows `{vendor}/{name}/jsonschema/{version}` convention
- [ ] Check all `self.vendor` fields match directory path
- [ ] Review field constraints for warehouse compatibility

### DIFF
--- a/schemas/com.snowplow.agent.tracking/agent_completion/jsonschema/1-0-0
+++ b/schemas/com.snowplow.agent.tracking/agent_completion/jsonschema/1-0-0
@@ -1,0 +1,76 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Records completion of an agent invocation with summary metrics. Agent context provides invocation details.",
+  "self": {
+    "vendor": "com.snowplow.agent.tracking",
+    "name": "agent_completion",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "invocation_id": {
+      "type": "string",
+      "description": "Invocation identifier",
+      "maxLength": 36
+    },
+    "total_steps": {
+      "type": "integer",
+      "description": "Total number of reasoning steps",
+      "minimum": 1,
+      "maximum": 1000
+    },
+    "total_duration_ms": {
+      "type": "integer",
+      "description": "Total time from invocation to completion",
+      "minimum": 0,
+      "maximum": 600000
+    },
+    "total_tokens": {
+      "type": "integer",
+      "description": "Total tokens used (prompt + completion)",
+      "minimum": 0,
+      "maximum": 2000000
+    },
+    "tools_called": {
+      "type": "integer",
+      "description": "Total number of tool calls made",
+      "minimum": 0,
+      "maximum": 1000
+    },
+    "business_tools_called": {
+      "type": ["integer", "null"],
+      "description": "Number of business tools called",
+      "minimum": 0,
+      "maximum": 1000
+    },
+    "self_tracking_tools_called": {
+      "type": ["integer", "null"],
+      "description": "Number of self-tracking tools called",
+      "minimum": 0,
+      "maximum": 1000
+    },
+    "finish_reason": {
+      "type": "string",
+      "enum": ["stop", "length", "error", "max_steps"],
+      "description": "Why the agent stopped"
+    },
+    "success": {
+      "type": "boolean",
+      "description": "Whether invocation completed successfully"
+    },
+    "final_response_length": {
+      "type": ["integer", "null"],
+      "description": "Length of final response to user",
+      "minimum": 0,
+      "maximum": 100000
+    },
+    "completed_at": {
+      "type": "string",
+      "format": "date-time",
+      "description": "Timestamp when invocation completed"
+    }
+  },
+  "required": ["invocation_id", "total_steps", "total_duration_ms", "total_tokens", "tools_called", "finish_reason", "success", "completed_at"],
+  "additionalProperties": false
+}
--- a/schemas/com.snowplow.agent.tracking/agent_completion/jsonschema/1-0-0
+++ b/schemas/com.snowplow.agent.tracking/agent_completion/jsonschema/1-0-0
@@ -12,7 +12,7 @@
     "invocation_id": {
       "type": "string",
       "description": "Invocation identifier",
-      "maxLength": 36
+      "format": "uuid"
     },
     "total_steps": {
       "type": "integer",
@@ -30,7 +30,7 @@
       "type": "integer",
       "description": "Total tokens used (prompt + completion)",
       "minimum": 0,
-      "maximum": 2000000
+      "maximum": 2147483647
     },
     "tools_called": {
       "type": "integer",
--- a/schemas/com.snowplow.agent.tracking/agent_completion/jsonschema/1-0-0
+++ b/schemas/com.snowplow.agent.tracking/agent_completion/jsonschema/1-0-0
@@ -18,7 +18,7 @@
       "type": "integer",
       "description": "Total number of reasoning steps",
       "minimum": 1,
-      "maximum": 1000
+      "maximum": 10000
     },
     "total_duration_ms": {
       "type": "integer",
--- a/schemas/com.snowplow.agent.tracking/agent_context/jsonschema/1-0-0
+++ b/schemas/com.snowplow.agent.tracking/agent_context/jsonschema/1-0-0
@@ -1,0 +1,62 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Context entity describing the agent, its configuration, and current state when performing actions.",
+  "self": {
+    "vendor": "com.snowplow.agent.tracking",
+    "name": "agent_context",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "invocation_id": {
+      "type": "string",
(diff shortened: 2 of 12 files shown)


schemas/com.snowplow.agent.tracking/agent_decision_logged/jsonschema/1-0-0
```diff
@@ -0,0 +1,45 @@
...
+    "name": "agent_decision_logged",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "type": "object",
+  "properties": {
+    "invocation_id": {
+      "type": "string",
+      "description": "Unique identifier for the agent invocation",
+      "maxLength": 36
```
istreeter: Is this a uuid?  If yes then you could add `format: uuid`.  Asking since you have max length 36, so I'm guessing uuid.  Same for `decision_id`, and several other IDs in this PR.

schemas/com.snowplow.agent.tracking/agent_decision_logged/jsonschema/1-0-0
```diff
@@ -0,0 +1,45 @@
...
+    },
+    "reasoning": {
+      "type": "string",
+      "description": "Natural language explanation of the decision",
+      "maxLength": 2000
+    },
+    "context": {
+      "type": ["object", "null"],
+      "description": "Additional context about the decision"
+    },
```
istreeter: Should there be any restriction at all on what is allowed in this `context` object?  E.g. force it to be string values?  Or do you want to truly allow anything?

schemas/com.snowplow.agent.tracking/agent_completion/jsonschema/1-0-0
```diff
@@ -0,0 +1,76 @@
...
+      "type": "integer",
+      "description": "Total time from invocation to completion",
+      "minimum": 0,
+      "maximum": 600000
+    },
+    "total_tokens": {
+      "type": "integer",
+      "description": "Total tokens used (prompt + completion)",
+      "minimum": 0,
+      "maximum": 2000000
```
istreeter: I wonder if there will be a time 2 or 3 years from now, when 2 million seems like a very small maximum.

schemas/com.snowplow.agent.tracking/user_intent_detected/jsonschema/1-0-0
```diff
@@ -0,0 +1,51 @@
...
+    "confidence": {
+      "type": "number",
+      "minimum": 0,
+      "maximum": 1,
+      "description": "Confidence level (0-1)"
+    },
+    "extracted_entities": {
+      "type": "object",
+      "description": "Extracted entities from user message"
+    },
```
istreeter: Same comment again for restrictions on an object.

istreeter (commented):
Technically these schemas are all fine.

We should get the opinions of a couple of people who know this domain better than I do.

schemas/com.snowplow.agent.tracking/agent_decision_logged/jsonschema/1-0-0
jrpeck1989: We definitely support `"format": "uuid"` as a format, correct? This [page](https://json-schema.org/understanding-json-schema/reference/type#resource-identifiers) suggests it was made available in draft 2019-09

schemas/com.snowplow.agent.tracking/agent_completion/jsonschema/1-0-0
jrpeck1989: Very good point. Is there an accepted "high max" value we typically provide to numeric fields like this? 999999999?

schemas/com.snowplow.agent.tracking/user_intent_detected/jsonschema/1-0-0
jrpeck1989: These objects act a little bit like an escape valve to allow application developers/analysts some flexibility for their business model. Essentially, in the application code, we use a front end schema validation library (Zod) to enforce a certain shape to this object, but the schema is freeform, to allow the end user to send which ever extracted entities there are. For instance, in the demo travel assistant app I've built alongside these schemas, when a user is asking about booking a flight, the agent is instructed to extract things like `origin`, `destination`, `date`, passengers`,`budget` etc. if the agent believes they're present in the user message. This gives the user application/business specific flexibility and application level validation. However, I can see why this is potentially problematic to have this in a schema on Iglu Central, as the user can essentially send anything, causing challenges in the data warehouse. 

What are your thoughts? Does allowing a passthrough object in the schema, but instructing users to enforce structure and validate in application code is acceptable? Should we create another schema, which will shore up this schema, but I still don't know how we provide the necessary flexibility in business logic in that schema either.

schemas/com.snowplow.agent.tracking/agent_completion/jsonschema/1-0-0
istreeter: A nice value to use is `2147483647` because it's the maximum size that a warehouse will store in a 32-bit integer column.

schemas/com.snowplow.agent.tracking/agent_decision_logged/jsonschema/1-0-0
istreeter: I am 99% certain we support it.  You could do a quick test with micro if there is any doubt.  Should be quite easy to add these schemas to a local micro.

jrpeck1989: Tested with Micro. `"format": "uuid"` works with both uppercase and lowercase UUIDs

schemas/com.snowplow.agent.tracking/agent_context/jsonschema/1-0-0
```diff
@@ -0,0 +1,62 @@
...
+      "type": ["integer", "null"],
+      "description": "Number of messages in conversation history at this point",
+      "minimum": 0,
+      "maximum": 10000
+    },
+    "current_step_number": {
+      "type": ["integer", "null"],
+      "description": "Current step number within the invocation",
+      "minimum": 1,
+      "maximum": 1000
```
Jack-Keene: Is there a reason to limit the maximum to 1000, is there a possibility that in the not too distant future more complex agents will exceed this?

schemas/com.snowplow.agent.tracking/agent_step/jsonschema/1-0-0
```diff
@@ -0,0 +1,71 @@
...
+      "description": "Sequential step number in this invocation",
+      "minimum": 1,
+      "maximum": 1000
+    },
+    "step_type": {
+      "type": "string",
+      "enum": ["initial", "continue", "tool-result"],
+      "description": "Type of step in the agent loop"
+    },
+    "prompt_tokens": {
```
Jack-Keene: Would input/output_tokens be a more standard name instead of prompt/completion tokens?

matus-tomlein (approved):
LGTM! Sorry for the copilot review noise, wanted to see if it finds something substantial but it's mostly just missing maxLength, so I don't see any blockers.

schemas/com.snowplow.agent.tracking/agent_step/jsonschema/1-0-0
jrpeck1989: Good point. I've updated the schemas accordingly

schemas/com.snowplow.agent.tracking/agent_context/jsonschema/1-0-0
jrpeck1989: Increased these to 10,000 (so they stay in the same size DB column

johnmicahreid:
@istreeter could we get a green tick from you now that @matus-tomlein and @Jack-Keene have given it a once-over?

istreeter (approved):

PR #1481, opened 2026-04-10
Added event_specification_validation schema for enrich

patofp:
Summary                                                                                                                                                                                        
                                                                                                                                                                                                 
  - Adds com.snowplowanalytics.snowplow/event_specification_validation/jsonschema/1-0-0
                                                                                                                                                                                                 
  Context                                                                                                                

  This schema defines a context entity that enrich attaches to events during event specification validation. Initially it will only be attached when validation fails, but we want to keep the door open to also attach it on passed validations without requiring a major version bump.
                                                                                                                                                                                                 
  This is reflected in the design:                                                                                       
  - isValid is a required boolean flag, so the entity can represent both outcomes
  - errors is optional, so it can be omitted entirely in the pass case                                                                                                                           
                                                                      
  The errors array items are intentionally unconstrained ("type": "object" with no required properties). This follows the precedent set by the existing failure schema, which documents expected fields in the description but does not enforce them in the schema itself.

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/event_specification_validation/jsonschema/1-0-0
+++ b/schemas/com.snowplowanalytics.snowplow/event_specification_validation/jsonschema/1-0-0
@@ -1,0 +1,40 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow",
+    "name": "event_specification_validation",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "description": "Validation result for an event against its declared event specification version.",
+  "type": "object",
+  "properties": {
+    "isValid": {
+      "type": "boolean",
+      "description": "Whether the event passed validation against its event specification."
+    },
+    "errors": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "message": {
+            "type": "string",
+            "description": "Human-readable error description."
+          },
+          "schema": {
+            "type": ["string", "null"],
+            "description": "Iglu schema URI of the entity or event that caused the error."
+          },
+          "path": {
+            "type": ["string", "null"],
+            "description": "JSON path to the property that failed validation."
+          }
+        },
+        "required": ["message"]
+      },
+      "description": "A list of validation errors."
+    }
+  },
+  "required": ["isValid"]
+}
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+Release 183 (2026-04-23)
+------------------------
+Add com.snowplowanalytics.snowplow/event_specification_validation/jsonschema/1-0-0 (#1481)
+
 Release 182 (2026-04-20)
 ------------------------
 Add com.optimizely.optimizelyx/summary/jsonschema/1-1-0 (#1483)


schemas/com.snowplowanalytics.snowplow/event_specification_validation/jsonschema/1-0-0
```diff
@@ -0,0 +1,26 @@
...
+  "properties": {
+    "isValid": {
+      "type": "boolean",
+      "description": "Whether the event passed validation against its event specification.",
+      "default": false
+    },
+    "errors": {
+      "type": "array",
+      "items": {
+        "type": "object"
```
istreeter: I'm interested in the design decision to leave this object unspecified, i.e. no `properties`.  Presumably you chose this to be equivalent to the `failure` schema [here](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/failure/jsonschema/1-0-0#L15-L21)?

It will be interesting to see the corresponding Enrich PR to see how these errors are set.

schemas/com.snowplowanalytics.snowplow/event_specification_validation/jsonschema/1-0-0
```diff
@@ -0,0 +1,26 @@
...
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
+  "description": "Validation result for an event against its declared event specification version.",
+  "type": "object",
+  "properties": {
+    "isValid": {
+      "type": "boolean",
+      "description": "Whether the event passed validation against its event specification.",
+      "default": false
```
istreeter: I don't think this default achieves anything, because it's a required field anyway.

schemas/com.snowplowanalytics.snowplow/event_specification_validation/jsonschema/1-0-0
patofp: Hey @ianarsenault, exactly, we want to follow failure schema error shape. 

Here how it would look a sample error:
```
{
    "message": "$.totalValue: must have a minimum value of 0",
    "schema": "iglu:com.snowmart/cart_details/jsonschema/1-0-0",
    "path": "$.totalValue"
 }
```

Still I would keep the error open for now.

schemas/com.snowplowanalytics.snowplow/event_specification_validation/jsonschema/1-0-0
patofp: Removed

schemas/com.snowplowanalytics.snowplow/event_specification_validation/jsonschema/1-0-0
istreeter: Will it always have just those three keys, `message`, `schema`, `path`?

The failure schema is a bit of a special case... when designing it, we needed to overcome the fact that Enrich errors are messy and inconsistent.  So as a compromise we made it `type: object` with no properties.  But that design is not great for users, because the loaders load the object as a JSON string (in all warehouses apart from Snowflake).  So the user needs to parse the object in the warehouse.

From the perspective of a user working with the entity in the warehouse, it is better to specify the fields explicitly in the schema.

patofp: I see, this was my initial proposal with three levels of errors:

Level 1: just a message
```
Event constraint:
  {
    "message": "Event data does not match constraint for spec purchase"
  }  Entity constraint:
  {
    "message": "Entity iglu:com.snowmart/cart_details/jsonschema/1-0-0 data does not match constraint"
  }  Entity cardinality:
  {
    "message": "Entity iglu:com.snplow.demo/product/jsonschema/1-0-0: expected at least 1, found 0"
  }
```

  Level 2: add source and schema

```
 Event constraint:
  {
    "message": "Event data does not match constraint for spec purchase",
    "source": "event_constraint",
    "schema": "iglu:com.snplow.demo/purchase/jsonschema/1-0-0"
  }  Entity constraint:
  {
    "message": "Entity data does not match constraint",
    "source": "entity_constraint",
    "schema": "iglu:com.snowmart/cart_details/jsonschema/1-0-0"
  }  Entity cardinality:
  {
    "message": "Entity expected at least 1, found 0",
    "source": "entity_cardinality",
    "schema": "iglu:com.snplow.demo/product/jsonschema/1-0-0",
    "expected": 1,
    "actual": 0,
    "bound": "min"
  }
```

  Level 3: matching the failure/1-0-0 shape with path, keyword, targets

```
  Event constraint:
  {
    "message": "$.transaction_id: does not have a value in the enumeration [1234]",
    "source": "event_constraint",
    "schema": "iglu:com.snplow.demo/purchase/jsonschema/1-0-0",
    "path": "$.transaction_id",
    "keyword": "enum",
    "targets": ["1234"]
  }  Entity constraint:
  {
    "message": "$.totalValue: must have a minimum value of 0",
    "source": "entity_constraint",
    "schema": "iglu:com.snowmart/cart_details/jsonschema/1-0-0",
    "path": "$.totalValue",
    "keyword": "minimum",
    "targets": ["0"]
  }  Entity cardinality:
  {
    "message": "Entity expected at least 1, found 0",
    "source": "entity_cardinality",
    "schema": "iglu:com.snplow.demo/product/jsonschema/1-0-0",
    "expected": 1,
    "actual": 0,
    "bound": "min"
  }
```

After discussing with @jborlase-snowplow and @gleb-lobov we ended up with the one that I showed you previously with `message`, `schema` and `path` but only `message` is present in all cases. You can have a look at the current errors that I'm generating in my local enrich:

```
  T6 (event constraint fail):
  {
    "message": "$.status: does not have a value in the enumeration [\"ok\"]",
    "schema": "iglu:com.test/custom_event/jsonschema/1-0-0",
    "path": "$.status"
  }

  T7 (cardinality min):
  {
    "message": "Entity iglu:com.test/custom_entity/jsonschema/1-0-0: expected at least 1, found 0",
    "schema": "iglu:com.test/custom_entity/jsonschema/1-0-0"
  }

  T8 (cardinality max):
  {
    "message": "Entity iglu:com.test/custom_entity/jsonschema/1-0-0: expected at most 2, found 3",
    "schema": "iglu:com.test/custom_entity/jsonschema/1-0-0"
  }

  T9 (entity constraint fail):
  {
    "message": "$.region: does not have a value in the enumeration [\"eu\"]",
    "schema": "iglu:com.test/custom_entity/jsonschema/1-0-0",
    "path": "$.region"
  }

  T10 (multiple: event constraint + cardinality):
  {
    "message": "$.status: does not have a value in the enumeration [\"ok\"]",
    "schema": "iglu:com.test/custom_event/jsonschema/1-0-0",
    "path": "$.status"
  }
  {
    "message": "Entity iglu:com.test/custom_entity/jsonschema/1-0-0: expected at least 1, found 0",
    "schema": "iglu:com.test/custom_entity/jsonschema/1-0-0"
  }

  T11 (SpecNotFound):
  {
    "message": "Event spec not found: id=00000000-0000-0000-0000-000000000000, version=999"
  }

  T12 (schema fail + validation):
  {
    "message": "Entity iglu:com.test/custom_entity/jsonschema/1-0-0: expected at least 1, found 0",
    "schema": "iglu:com.test/custom_entity/jsonschema/1-0-0"
  }

  T13 (event + entity constraint both fail):
  {
    "message": "$.status: does not have a value in the enumeration [\"ok\"]",
    "schema": "iglu:com.test/custom_event/jsonschema/1-0-0",
    "path": "$.status"
  }
  {
    "message": "$.region: does not have a value in the enumeration [\"eu\"]",
    "schema": "iglu:com.test/custom_entity/jsonschema/1-0-0",
    "path": "$.region"
  }

  T14 (event constraint fail, entity passes):
  {
    "message": "$.status: does not have a value in the enumeration [\"ok\"]",
    "schema": "iglu:com.test/custom_event/jsonschema/1-0-0",
    "path": "$.status"
  }

  T15 (event passes, entity constraint fail):
  {
    "message": "$.region: does not have a value in the enumeration [\"eu\"]",
    "schema": "iglu:com.test/custom_entity/jsonschema/1-0-0",
    "path": "$.region"
  }
```

There might be the case were we need to add more fields on the errors for data quality dashboard.

patofp: I committed the three fields, `schema` and `path` as optionals 05d6c9531087c795c64815bdb334d23f0620b10e
Let me know if this would improve user readability on the warehouse

istreeter (approved):
Having reviewed the corresponding Enrich PR snowplow/enrich-private#222 I think this schema is fine. It correctly matches the feature in Enrich.

cksnp (approved):

istreeter:
Force pushed to squash the commits and added a "Prepare for release" commit.  No change made to the schema.

PR #1483, opened 2026-04-20
Add com.optimizely.optimizelyx/summary/jsonschema/1-1-0

matus-tomlein:
## Summary

Adds schema version `1-1-0` of `com.optimizely.optimizelyx/summary`, introducing an optional `campaignId` field.

## Why

The `browser-plugin-optimizely-x` plugin in [snowplow/snowplow-javascript-tracker#1463](https://github.com/snowplow/snowplow-javascript-tracker/pull/1463) switches from `getActiveExperimentIds()` to `getCampaignStates({ isActive: true })` to correctly capture the campaign layer for Optimizely Personalization campaigns, which was previously dropped entirely. The tracker PR emits events against this new schema version.

## Changes vs 1-0-0

- Adds optional `campaignId` (`integer | null`, same bounds as other IDs).
- All existing fields unchanged.

## SchemaVer bump

REVISION (`1-0-0` → `1-1-0`). The base schema has `additionalProperties: true`, so per [SchemaVer rules](https://docs.snowplow.io/docs/api-reference/iglu/common-architecture/schemaver/) adding an optional property is a REVISION rather than an ADDITION — historical events could have carried a conflicting `campaignId` via the open map.

## Compatibility

`campaignId` is optional, so existing `1-0-0` events remain valid under `1-1-0`.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+Release 182 (2026-04-20)
+------------------------
+Add com.optimizely.optimizelyx/summary/jsonschema/1-1-0 (#1483)
+
 Release 181 (2026-04-15)
 ------------------------
 Add com.iterable/system_webhook/jsonschema/2-0-1 (#1480)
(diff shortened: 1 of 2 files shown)


istreeter (approved):
Looks fine.  Simple addition of a new optional property.

igneel64 (approved):

PR #1485, opened 2026-06-03
New agent classification enrichment

benjben:
JIRA ref: [PDP-2680](https://snplow.atlassian.net/browse/PDP-2680)

Related Enrich PR: https://github.com/snowplow/enrich-private/pull/241

[PDP-2680]: https://snplow.atlassian.net/browse/PDP-2680?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


schemas/com.snowplowanalytics.snowplow.enrichments/agent_classification_enrichment_config/jsonschema/1-0-0
```diff
@@ -0,0 +1,72 @@
...
+                        "type": "object",
+                        "properties": {
+                            "operator": {
+                                "type": "string",
+                                "description": "The vendor operating the agent"
+                            },
+                            "purpose": {
+                                "type": "string",
+                                "description": "The purpose of the agent, if known"
+                            }
```
istreeter, line 55: Neither of these fields are required?  Just checking.  Should we allow nulls?

schemas/com.snowplowanalytics.snowplow.enrichments/agent_classification_enrichment_config/jsonschema/1-0-0
```diff
@@ -0,0 +1,72 @@
...
+                        },
+                        "minProperties": 1,
```
istreeter: Why not allow zero properties?

istreeter (commented):
I'm sure this is all fine, I just left some comments where they occurred to me.

I suggest get @stanch to be the second reviewer.

schemas/com.snowplowanalytics.snowplow.enrichments/agent_classification_enrichment_config/jsonschema/1-0-0
stanch: Operator should be required. My reasoning is:
* at least one field should be present, otherwise the entry is not useful
* if purpose is known, then surely the operator is known (otherwise where do we get the authoritative info on the purpose)
* therefore, we should allow `operator` alone but not `purpose` alone

schemas/com.snowplowanalytics.snowplow.enrichments/agent_classification_enrichment_config/jsonschema/1-0-0
stanch: See above, I would rather set `operator` as required and remove the `minProperties` restriction

benjben:
I addressed the feedback, please have a look!

schemas/com.snowplowanalytics.snowplow.enrichments/agent_classification_enrichment_config/jsonschema/1-0-0
istreeter: The other part of my earlier question was: should we allow `null` for the field which is not required?

benjben: If we are consistent with ASN we should: https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/asn/jsonschema/1-0-1#L18-L21

Even though Enrich omits these fields in the entitties when they are not known. Let me update.

benjben: Oops sorry I mixed enrichment config and entity. For ASN lookups config we don't allow `null`: https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow.enrichments/asn_lookups/jsonschema/1-0-0#L53-L56

But in this case the enrichment allows it already, so I'll update.

istreeter (approved):
All good from my point of view

stanch (approved):

istreeter:
Added the release commit to this PR so we can merge it all together.

PR #1486, opened 2026-07-15
Release 185

benjben:
[PDP-2743](https://snplow.atlassian.net/browse/PDP-2743)

- Add com.snowplowanalytics.snowplow/utm_referrer/jsonschema/1-0-0
- Add com.snowplowanalytics.snowplow/referer_parser/jsonschema/2-0-2

[PDP-2743]: https://snplow.atlassian.net/browse/PDP-2743?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/referer_parser/jsonschema/2-0-2
+++ b/schemas/com.snowplowanalytics.snowplow/referer_parser/jsonschema/2-0-2
@@ -1,0 +1,76 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "description": "Schema for referer-parser enrichment",
+    "self": {
+      "vendor": "com.snowplowanalytics.snowplow",
+      "name": "referer_parser",
+      "format": "jsonschema",
+      "version": "2-0-2"
+    },
+    "type": "object",
+    "properties": {
+      "vendor": {
+        "type": "string"
+      },
+      "name": {
+        "type": "string"
+      },
+      "enabled": {
+        "type": "boolean"
+      },
+      "parameters": {
+        "type": "object",
+        "properties": {
+          "internalDomains": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            }
+          },
+          "database": {
+            "type": "string"
+          },
+          "uri": {
+            "type": "string",
+            "format": "uri"
+          },
+          "referrers": {
+            "type": "object",
+            "additionalProperties": {
+              "type": "object",
+              "additionalProperties": {
+                "type": "object",
+                "properties": {
+                  "domains": {
+                    "type": "array",
+                    "minItems": 1,
+                    "items": {
+                      "type": "string"
+                    }
+                  },
+                  "parameters": {
+                    "type": "array",
+                    "items": {
+                      "type": "string"
+                    }
+                  },
+                  "utm_sources": {
+                    "type": "array",
+                    "items": {
+                      "type": "string"
+                    }
+                  }
+                },
+                "required": ["domains"],
+                "additionalProperties": false
+              }
+            }
+          }
+        },
+        "required": ["internalDomains", "database", "uri"],
+        "additionalProperties": false
+      }
+    },
+    "required": ["name", "vendor", "enabled", "parameters"],
+    "additionalProperties": false
+}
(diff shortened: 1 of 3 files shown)


schemas/com.snowplowanalytics.snowplow/referer_parser/jsonschema/2-0-2
```diff
@@ -0,0 +1,76 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "description": "Schema for referer-parser customization enrichment",
```
stanch: Why customization?

schemas/com.snowplowanalytics.snowplow/referer_parser/jsonschema/2-0-2
```diff
@@ -0,0 +1,76 @@
...
+                    "items": {
+                      "type": "string"
+                    }
+                  },
+                  "utm_sources": {
+                    "type": "array",
+                    "items": {
+                      "type": "string"
+                    }
+                  }
```
stanch, line 62: This is the new part right?

stanch (approved):

schemas/com.snowplowanalytics.snowplow/referer_parser/jsonschema/2-0-2
benjben: No idea, I (Claude) just copy pasted `2-0-1` and added the new field. I removed it.

schemas/com.snowplowanalytics.snowplow/referer_parser/jsonschema/2-0-2
benjben: Yes!

```
$ diff com.snowplowanalytics.snowplow/referer_parser/jsonschema/2-0-2 com.snowplowanalytics.snowplow/referer_parser/jsonschema/2-0-1
3c3
<     "description": "Schema for referer-parser enrichment",
---
>     "description": "Schema for referer-parser customization enrichment",
8c8
<       "version": "2-0-2"
---
>       "version": "2-0-1"
52,57d51
<                     "type": "array",
<                     "items": {
<                       "type": "string"
<                     }
<                   },
<                   "utm_sources": {

```

spenes (approved):

schemas/com.snowplowanalytics.snowplow/utm_referrer/jsonschema/1-0-0
```diff
@@ -0,0 +1,30 @@
...
+    },
+    "medium": {
+      "type": "string",
+      "description": "The medium of the referer (e.g. chatbot), equivalent to refr_medium",
+      "maxLength": 25
+    },
+    "term": {
+      "type": ["string", "null"],
+      "description": "The search term extracted from the referer, equivalent to refr_term",
+      "maxLength": 255
```
benjben: I added the lengths limits. They seem quite aggressive but they are the same values as defined [here](https://github.com/snowplow/enrich-private/blob/develop/modules/common/src/main/resources/reference.conf#L190-L192) and in `atomic` schema.

istreeter (approved):

schemas/com.snowplowanalytics.snowplow/utm_referrer/jsonschema/1-0-0
benjben: @stanch there has been some discussions internally whether we should add `maxLength`s or not. For validation it is not useful as we already have validation in-place for the same fields of the atomic event. Data modeling with BigQuery, Databricks and Snowflake is fine without them. The only sensible reason to add them is data modeling with Redshift, in case these fields are used in a join/group/sort rather than just being projected. Do you think that we need these `maxLength`s?

stanch: Let’s add maxLength but double the limits compared to the atomic fields, just to be on the safer side. This is because the atomic limits can be easily extended per customer, and these ones can’t.

benjben: Sounds good!

@istreeter I updated.

PR #1489, opened 2026-09-11
Add com.snowplowanalytics.snowplow/identity/jsonschema/2-0-1

peel:
[ISP-279](https://snplow.atlassian.net/browse/ISP-279) (parent: [ISP-278](https://snplow.atlassian.net/browse/ISP-278))

## Summary

Adds schema version `2-0-1` of `com.snowplowanalytics.snowplow/identity`, introducing an optional `decision_reasons` field that reports why Snowplow Identities made a decision for the event (for example, refusing a merge because the identity is at its merge limit). The warehouse models use it to label a split identifier and prefer the correct identity.

## Changes vs 2-0-0

- Adds optional `decision_reasons`: array of strings. The set of reason codes is owned by the service and is deliberately not enumerated in the schema, so new reasons can be added without a schema change. Omitted entirely when there is nothing to report; never `null`.
- `required` and `additionalProperties: false` unchanged. All existing fields unchanged.

## SchemaVer bump

ADDITION (`2-0-0` → `2-0-1`): adding a new optional field is compatible with all historical data, per [SchemaVer](https://docs.snowplow.io/docs/fundamentals/schemas/versioning/). Stays in MODEL 2, so it lands in the existing `contexts_com_snowplowanalytics_snowplow_identity_2` column.

## Compatibility

`decision_reasons` is optional, so existing `2-0-0` events remain valid under `2-0-1`. This schema must be published before Enrich emits against it.

## Related

- Snowplow Identities and Enrich changes to follow under ISP-278 (this schema ships first).

Supersedes #1487, which GitHub closed when the head branch was renamed from `isp-279-identity-2-1-0`.

[ISP-279]: https://snplow.atlassian.net/browse/ISP-279?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ISP-278]: https://snplow.atlassian.net/browse/ISP-278?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/schemas/com.snowplowanalytics.snowplow/identity/jsonschema/2-0-1
+++ b/schemas/com.snowplowanalytics.snowplow/identity/jsonschema/2-0-1
@@ -1,0 +1,31 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "type": "object",
+  "self": {
+    "vendor": "com.snowplowanalytics.snowplow",
+    "name": "identity",
+    "format": "jsonschema",
+    "version": "2-0-1"
+  },
+  "description": "Identity context that is enriched onto events by Snowplow Identities",
+  "properties": {
+    "created_at": {
+      "type": "string",
+      "format": "date-time"
+    },
+    "snowplow_id": {
+      "type": "string",
+      "maxLength": 29,
+      "pattern": "^sp_[A-Za-z2-7]{26}$"
+    },
+    "decision_reasons": {
+      "description": "Why the service made this decision. Absent when there is nothing to report.",
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    }
+  },
+  "required": ["snowplow_id", "created_at"],
+  "additionalProperties": false
+}


schemas/com.snowplowanalytics.snowplow/identity/jsonschema/2-0-1
```diff
@@ -0,0 +1,31 @@
...
+    "snowplow_id": {
+      "type": "string",
+      "maxLength": 29,
+      "pattern": "^sp_[A-Za-z2-7]{26}$"
+    },
+    "decision_reasons": {
+      "description": "Why the service made this decision. Absent when there is nothing to report.",
+      "type": "array",
+      "items": {
+        "type": "string"
```
istreeter: Would you consider a maxLength on these strings?

We tend to put `maxLength` on every string field, even ones the pipeline generates itself.  The reasoning is: nothing stops a tracker sending this entity directly. It's cheap protection against a weird client putting something huge in the warehouse, and it's a useful contract for downstream consumers.

That said, it's a nice-to-have.  So it's fine to ignore this idea if you don't want to be bound by a max length.

peel: Good spot, thanks Ian. I made it big enough so it fits our wildest future decision reasons.

istreeter (approved):

jamessnowplow (approved):
Looks good to me.

istreeter:
@peel @jamessnowplow I just had another thought about this PR.

Have you considered how the new column looks for Redshift customers?  Our Redshift integration is much less elegant than our other loaders.  The customer will see a string column containing a JSON array.  They will need to json-parse the value to extract the strings.

```
|-------------|------------------|
| snowplow_id | decision_reasons |
|-------------|------------------|
| sp_1234     | ["foo"]          |
| sp_2345     | ["bar"]          |
| sp_3456     | ["baz"]          |
```

I'm not saying it's a problem, but maybe something you want to consider?  In particular any impact on DBT models.

peel:
I am not worried about this. I'm very much ok with this being a string array. This is somewhat narrow case and when we hit it, the models should be able to cope.

istreeter:
Merging into the release branch

Issue #66, opened 2014-11-10
Update payload schema to make p (platform) an enum rather than a string


yalisassoon:
Note that it can only take a handful of parameters, as specified [here](https://github.com/snowplow/snowplow/wiki/snowplow-tracker-protocol#11-application-parameters)

PR #1490, opened 2026-09-18
Add parquet column names

istreeter:
See snowplow/snowplow-rdb-loader-private#66.

We are changing RDB loader so that parquet column names are carried in the shredding_complete message between the transformer and loader.  This eliminates a class of potential error, in case different versions of transformer and loader disagree on recovery column names.

This PR adds a single new field `columnName` to the existing schemas.


AlexBenny (approved):
Hey @istreeter , looks good for me. Few thoughts:

  - The identity 2-0-1 commit from #1489 is here, not sure if it was expected.
  - Claude mentioned you added regex in the loader (^[A-Za-z_][A-Za-z0-9_]*$) as a pattern on columnName. Is it worth to add here or do we avoid regex in the loader's schemas? I guess fine to skip since none of these schemas use patterns.

Side note: the loader sends shredding_complete as ...storage.rdbloader, but here it's ...storage. Btw it's older than this PR.

Links:
Loader, where the vendor with .rdbloader is set (the #66 merge commit):
https://github.com/snowplow/snowplow-rdb-loader-private/blob/972a39c8bd0a369267e5cf2696d8d4eceb3d9733/modules/common/src/main/scala/com/snowplowanalytics/snowplow/rdbloader/common/LoaderMessage.scala#L43-L48

The same lines on the loader's master, showing it was like this before #66 (2-0-1):
https://github.com/snowplow/snowplow-rdb-loader-private/blob/8cd124b0ee780458e8ea91648aa351eee4ecbcb5/modules/common/src/main/scala/com/snowplowanalytics/snowplow/rdbloader/common/LoaderMessage.scala#L43-L48

Iglu Central, the schema in this PR, which has no .rdbloader:
https://github.com/snowplow/iglu-central/blob/9a371d9264e7672cf40c0cddb99cfe660feee71c/schemas/com.snowplowanalytics.snowplow.storage/shredding_complete/jsonschema/2-0-2#L4-L9

For contrast, load_succeeded, where the vendor matches:
https://github.com/snowplow/snowplow-rdb-loader-private/blob/972a39c8bd0a369267e5cf2696d8d4eceb3d9733/modules/loader/src/main/scala/com/snowplowanalytics/snowplow/rdbloader/dsl/Monitoring.scala#L66

istreeter:
> The identity 2-0-1 commit from https://github.com/snowplow/iglu-central/pull/1489 is here, not sure if it was expected.

Whoops, I will change the merge branch to `develop`. That will remove that other schema from this PR.

> you added regex in the loader `(^[A-Za-z_][A-Za-z0-9_]*$)` as a pattern on columnName

Interesting idea and I hadn't considered it. The loader would need to do that validation in order to protect itself, whether the regex is in the schema or not.  I don't strongly feel the regex is needed in the schema.

> Side note: the loader sends shredding_complete as ...storage.rdbloader, but here it's ...storage. Btw it's older than this PR.

Yeah I noticed that.  Well, Claude pointed it out to me recently.  It's definitely a mistake.  I decided not to fix it because it's been that way for a long time and I don't want to make this rollout any more complicated.

schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/3-0-2
```diff
@@ -0,0 +1,197 @@
...
+                        "type": "string",
+                        "maxLength": 256
+                      },
+                      "snowplowEntity": {
+                        "description": "Type of the self-describing JSON in the event",
+                        "enum": ["SELF_DESCRIBING_EVENT", "CONTEXT"]
+                      },
+                      "columnName": {
+                        "description": "Parquet column this entity was written to. Absent and null are equivalent: no column holds its data, because the output is not parquet or the schema yields no column",
+                        "type": ["string", "null"],
```
benjben, line 90: As I see it we should allow `null` values only if they have a special meaning compared with the field being absent, which is not the case here. Do you have a different opinion?

benjben (approved):

schemas/com.snowplowanalytics.monitoring.batch/load_succeeded/jsonschema/3-0-2
istreeter: I did spend some time thinking about null vs missing values.  And eventually I decided on this:

A missing value and a null value have the same semantic meaning.  If a value is missing or if a value is null then it means the same thing.  And therefore the schema should treat the two the same.  If the schema allows missing then it should also allow null.

